PR #15, opened 2021-06-16
feat(reader): add initial settings

dskuza:


Shared/Item/ItemDestinationView.swift
```diff
@@ -35,7 +41,9 @@ struct ItemDestinationView: View {
         destinationView
             .toolbar(item: item,
```
jacobsimeon, line 42: Another bit of style commentary.

I feel like there are two basic styles to writing method calls: I'll call them 1) single-line and 2) chopped-down.

The formatting for the single-line style is pretty self-evident so I won't go into that. As for the chopped-down version, here's what I like:

1. First parameter goes on a new line
2. Each parameter goes on its own line
2. Closing paren is on its own line

For example:

```swift
destinationView.toolbar( // no param here
    item: item, // first param goes on a new line
    presentationMode: ...,
    shouldPresentWebView: ...
).anotherMethod() 
// closing paren on its own line, easier to see what the chained method is attached to
```

I particularly like this over the format of keeping the first param on the same line as the opening paren because it avoid large swaths of whitespace before your eyes get to the next param

```swift
destinationView.toolbar(item: item, // first param on same line as opening paren
                        presentationMode: ..., // now there is a huge crater of whitespace before the second param
                        shouldPresentWebView: ...) // closing paren here is hard for my eyes to see the end of the statement
    .anotherMethod() // now this chained method is kinda out on its own
```

Shared/Item/ItemDestinationView.swift
```diff
@@ -76,16 +90,31 @@ private struct ItemToolbar: ViewModifier {
...
 }
 
 private extension View {
-    func toolbar(item: Item, presentationMode: Binding<PresentationMode>, shouldPresentWebView: Binding<Bool>) -> some View {
-        self.modifier(ItemToolbar(item: item, presentationMode: presentationMode, shouldPresentWebView: shouldPresentWebView))
+    func toolbar(item: Item,
+                 presentationMode: Binding<PresentationMode>,
+                 shouldPresentWebView: Binding<Bool>,
+                 shouldPresentOverflow: Binding<Bool>) -> some View {
+        self.modifier(ItemToolbar(item: item,
```
jacobsimeon, line 115: See above comment re: chopped down method calls

Shared/Reader/ReaderSettingsView.swift
```diff
@@ -0,0 +1,62 @@
...
+                    }
+                    
+                    Stepper("Font Size",
+                            value: $settings.fontSizeAdjustment,
+                            in: Constants.allowedAdjustments,
+                            step: Constants.adjustmentStep)
+                }
+                
+                Section(header: Text("Preview")) {
+                    Text(
```
jacobsimeon, line 39: I like to make a special exception for chopped down method calls when the first param is an array or dictionary literal. Keep the opening square bracket on the same line as the opening paren.

Example:

```swift
myFunctionThatTakesAnArray([
  "item 1",
  "item 2",
  "item 3"
]) // closing bracket and paren on the same line
```

Textile/Style/Typography/FontDescriptor.swift
```diff
@@ -55,6 +59,16 @@ public struct FontDescriptor {
     func with(slant: Slant) -> FontDescriptor {
         FontDescriptor(family: family, size: size, weight: weight, slant: slant)
     }
+    
+    func adjustingSize(by adjustment: Int) -> FontDescriptor {
```
jacobsimeon, line 63: What do you think of extending this one step further and adding `adjusting(by: Int)` to `FontSize`? That might make this a bit nicer, like so:

```swift
return FontDescriptor(family: family, size: size?.adjusting(by: adjustment), weight: weight, slant: slant)
```

jacobsimeon (commented):
This looks great! Super love the use of `@AppStorage`!

Just a couple of questions/comments regarding style and a suggestion to read `ReaderSettings` from the environment so we can avoid some repetition.

Great work!

Shared/Article/ArticleComponentView.swift
```diff
@@ -17,23 +20,23 @@ struct ArticleComponentView: View {
     var body: some View {
         switch component {
         case .bodyText(let bodyText):
-            TextContentView(bodyText.text, style: .body.serif)
+            TextContentView(bodyText.text, style: .body.serif.with(settings: settings))
```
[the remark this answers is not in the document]
dskuza: 👍 Definitely a thing, and is much cleaner!

Shared/Item/ItemDestinationView.swift
```diff
@@ -11,10 +11,16 @@ struct ItemDestinationView: View {
     private let article: Article?
     
     @Environment(\.presentationMode) @Binding
-    var presentationMode: PresentationMode
+    private var presentationMode: PresentationMode
+    
+    @State
+    private var shouldPresentWebView = false
```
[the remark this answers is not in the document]
dskuza: Love it. I overlooked the typical `isSomething` convention for `bool` here, so thanks for calling it out!

Shared/Item/ItemDestinationView.swift
dskuza: I like this! Does this also come into play when defining a function, or primarily at the call site?

Textile/Style/Typography/FontDescriptor.swift
dskuza: Yep, that ends up cleaner! Something else that's popping up in my head: naming this function `adjusted(by:)` vs `adjusting(by:)` (-ed vs -ing). What are your thoughts?

Shared/Reader/ReaderSettingsView.swift
dskuza: 👍 I think the other thing to do here is extract the preview text generation into its own thing as well. I feel like having a length single argument isn't the _nicest_ thing. I'll give that a shot and see how things get cleaned up!

Shared/Item/ItemDestinationView.swift
jacobsimeon: Yes, same rules apply for method definitions as well 👍

Textile/Style/Typography/FontDescriptor.swift
jacobsimeon: yeah, I like `adjusted` too

PR #27, opened 2021-06-25
fix(sync): safely save on main thread

dskuza:
Calling `sync` on the main thread results in a crash. This can be resolved by directly saving the context while on the main thread, otherwise dispatching `sync` to the main queue.

### DIFF
--- a/Sync/Space.swift
+++ b/Sync/Space.swift
@@ -39,9 +39,11 @@ class Space {
     }
     
     func save() throws {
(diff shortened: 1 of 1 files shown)


nzeltzer (approved):
🍭 lgtm

nzeltzer:
@dskuza @jacobsimeon  On the legacy client, we had a macro `PKTDispatchMain(...)` for this sort of thing, if that's useful.

PR #28, opened 2021-07-06
Target iOS 15 and macOS 12

jacobsimeon:


nzeltzer (approved):
lgtm

(I can't speak to the test change, but I trust the test will speak for itself)

Shared/Item/ItemListView.swift
```diff
@@ -23,9 +23,11 @@ struct ItemListView: View {
     var body: some View {
         List(items) { item in
             ZStack(alignment: .leading) {
+                NavigationLink(
+                    destination: ItemDestinationView(item: item)
+                ) { EmptyView() }
+
                 ItemRowView(model: ItemPresenter(item: item))
-                NavigationLink(destination: ItemDestinationView(item: item)) { }
-                    .hidden()
```
dskuza, line 28: Aha! Nice discovery.

dskuza (approved):
Looks good to me. I noticed the commit messages aren't the previously used "conventional commit" styling. Should we retro that styling?

jacobsimeon:
> (I can't speak to the test change, but I trust the test will speak for itself)

@nzeltzer looks like iOS 14 could only find that button based on the name of the image icon. iOS 15 does too, unless you've given the button an accessibility identifier, in which case you need to search for it based on the identifier.

jacobsimeon:
@dskuza, I squashed the commits and changed the message to use conventional commits

PR #30, opened 2021-07-12
feat(list): add support for right-to-left languages

dskuza:


PocketKit/Sources/PocketKit/Item/ItemPresenter.swift
```diff
@@ -19,6 +19,12 @@ class ItemPresenter: ItemRow {
     public var title: String {
         item.title ?? item.thumbnailURL?.absoluteString ?? "Missing Title"
     }
+    
+    public var languageDirection: LanguageDirection {
+        let language = item.language ?? "en"
```
dskuza, line 24: The majority of articles returned are "en", so I think it's safe to default to "en" if no language is returned from the server.

PocketKit/Sources/Textile/Views/ItemRow.swift
```diff
@@ -71,6 +81,17 @@ public struct ItemRowView<Model: ItemRow>: View {
...
+extension ItemRowView {
+    private var textAlignment: TextAlignment {
+        switch model.languageDirection {
+        case .leftToRight:
+            return layoutDirection == .leftToRight ? .leading : .trailing
+        case .rightToLeft:
+            return layoutDirection == .rightToLeft ? .leading : .trailing
+        }
+    }
+}
```
nzeltzer, line 93: Interesting. As I understand it, leading and trailing will be be left/right or right/left, depending on the phone's locale. So, if I'm in an Arabic locale, leading will be right edge, and trailing the left. Will this properly align English text in that locale, or have the opposite intended effect?

nzeltzer (commented):
@dskuza One question.

PocketKit/Sources/Textile/Views/ItemRow.swift
dskuza: @nzeltzer I think screenshots will showcase this best:

System language: left-to-right, with an example of both left-to-right and right-to-left article titles
image

System language: right-to-left with an example of both left-to-right and right-to-left article titles
image

nzeltzer (approved):
lgtm

PocketKit/Sources/PocketKit/Item/ItemPresenter.swift
jacobsimeon: The majority of articles are "en", sure. But that is probably because the majority of our users are "en" speakers.

For speakers of RTL languages, what language are the majority of articles in? Put more broadly, for any user are the majority of articles in that user's system language? Or are the majority of articles still "en", even if their system language is non-en?

PR #34, opened 2021-07-20
feat: exclude archived items from list

jacobsimeon:
- Update graphql schema
- Query for `isArchived` property of SavedItem
- Include `isArchived: false` filter when fetching data from API
- Persist `isArchived` property in core data
- Add `isArchived = false` predicate to Item fetch request

### DIFF
--- a/PocketKit/Sources/Sync/Requests.swift
+++ b/PocketKit/Sources/Sync/Requests.swift
@@ -8,6 +8,7 @@ import CoreData
 public enum Requests {
     public static func fetchItems() -> NSFetchRequest<Item> {
         let request: NSFetchRequest<Item> = Item.fetchRequest()
+        request.predicate = NSPredicate(format: "isArchived = false")
         request.sortDescriptors = [
             NSSortDescriptor(keyPath: \Item.timestamp, ascending: false),
             NSSortDescriptor(keyPath: \Item.title, ascending: true)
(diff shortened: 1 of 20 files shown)


dskuza (approved):

PocketKit/Sources/Sync/Requests.swift
```diff
@@ -7,12 +7,20 @@ import CoreData
 
 public enum Requests {
     public static func fetchItems() -> NSFetchRequest<Item> {
+        let request = fetchAllItems()
```
dskuza, line 10: I think the word `all` may be overloaded. `All` items may be interpreted as all items a user has, including unarchived, archived, and deleted, but here, `all` means "unarchived and not deleted". What do you think about renaming this? `fetchActiveItems`, `fetchCurrentItems`, `fetchListItems`, `fetchSomeOtherWordItems`?

dskuza (commented):
Simple changes, I like it. I just have one question regarding the naming of a function!

PocketKit/Sources/Sync/Requests.swift
jacobsimeon: well... `All` here actually means "anything in the table". the archived/deleted flags only come into account in `fetchItems` request (which adds predicates to `fetchAllItems`)

dskuza: Ah! You're right. My interpretation of this was incorrect :)

dskuza (approved):

PR #35, opened 2021-07-21
fix: show URL as item title when title is missing or empty

jacobsimeon:
https://getpocket.atlassian.net/browse/IN-168


dskuza (approved):
Ha, whoops! Nice catch.

PR #37, opened 2021-07-22
feat(analytics): initial snowplow support

dskuza:
This pull request introduces a subset of the initial requirements for this app, where the subset is those available based on the current development status of the app.

## Dependencies

[+] [Snowplow](https://github.com/snowplow/snowplow-objc-tracker) / [Docs](https://docs.snowplowanalytics.com/)

## Config

To set up which endpoint to send Snowplow events to, and with what identifier, add the following build settings to the `secrets.xcconfig` file:

- SNOWPLOW_ENDPOINT
  - Format: `http%3a%2f%2fYOUR_ENDPOINT_HERE`
    - Due to how `.xcconfig` files are used, a `//` is considered a comment. App setup assumed the endpoint to be percent-escaped.
- SNOWPLOW_IDENTIFIER
  - The identifier dictating which app (scheme) is sending the events.

## Analytics

- `app_open`
- `app_background`
- `content_open`
  - When an item is selected from `My List`
  - When a link is opened from Artivle View
- `impression`
  - When an item is scrolled to in `My List`
- `engagement`
  - When the user switches to web view in Article View

### Not Yet Implemented

- `impression`
  - When a user opens the app or switches tabs
    - The user is currently always on `My List`, and the navigation is yet to be defined / implemented.
- `engagement`
  - Item actions (not yet implemented within this Pull Request)
    - Save / Archive
    - Share
    - Favorite / Unfavorite
    - Add Tags
    - Delete

## Tracking

Tracking is performed using an object that conforms to the `Tracker` protocol. A Pocket-specific tracker can be found in `Tracker.swift` (`PocketTracker`). This Pocket-specific tracker requires one dependency - an object capable of sending events / contexts to Snowplow. The pre-defined Pocket-specific dependency can be found in `SnowplowTracking.swift` (`PocketSnowplowTracker`). This tracker uses the newly-added `Snowplow` dependency to build the correct event, attaching the appropriate contexts as well.

The app-wide dependency can be found in `Services.swift` as the `tracker` property.

Once implemented, tracking is simple - call `track(event:_)` with the appropriate `SnowplowEvent` and `SnowplowContext` objects. For defining events and contexts, see below.
  
### Event

To define a Snowplow event, simply conform to the `SnowplowEvent` protocol. This will require you to define a schema for the event, as well as ensure that the object is encodable.

```swift
struct MyEvent: SnowplowEvent {
    static let schema = "iglu:..."

    let value: Int
}
```

### Context

Defining a Snowplow context is very similar to defining an event, but conforming to `SnowplowContext` instead.

```swift
struct MyContext: SnowplowContext {
    static let schema = "iglu:..."

    let value: Int
}
```

## SwiftUI

This pull request introduces the concept of a `TrackableView`, which, when used in conjunction with nested SwiftUI views, creates a stack containing the contexts of each view on the stack, so long as the view is tracked.

To create a trackable view, you can either initialize a `TrackableView` (which wraps a given SwiftUI view, capturing its context), or use the `.trackable` convenience function on any view, supplying the correct context for the view.

```swift
// Define an event that can be sent to Snowplow.
struct ContentEvent: SnowplowEvent {
    let schema = ...

    let value = "Hello, world"
}

// Dummy row that contains a button that will send an event to Snowplow.
struct ContentRow: View {
    @Environment(\.uiContexts)
    var uiContexts: [UIContext]

    @Environment(\.tracker)
    var tracker: Tracker

    var body: some View {
        Button("Hello, world") {
            // When this event is tracked, the uiContexts will be the list contained in ContentView, 
            // as well as this button's parent view (ContentRow) since it was marked as trackable within its parent.
            tracker.track(ContentEvent(), uiContexts)
        }
    }    
}

// Dummy (trackable) view that lists n number of (trackable) rows.
struct ContentView: View {
    var body: some View {
        List {
            ForEach(...) { content in
                ContentRow().trackable(
                    UIContext(type: .card, identifier: .home, hierarchy: index(of: content), componentDetail: .itemRow)
                )
            }
        }.trackable(UIContext(type: .list, identifier: .home, hierarchy: 0, componentDetail: nil))
    }
}
```

### Under the Hood

When calling `.trackable` on a view, that view is added to a "stack" of views that are being tracked, so that we can later appropriately send an event with its "stack" of views sent as the context of the event, in this case that it was a row within a list. `TrackableView` is a wrapper view of sorts that simply presents the content of its initial closure, but internally adds the new trackable view to its stack of already-tracked views. The new context(s) are then forwarded to the view that is being presented _by_ the `TrackableView`, such that when a given view tracks an event, it contains the stack of views up to and including itself. Note that not all nested views will be tracked; only those modified by `trackable` will be tracked.

TL;DR - Upon tracking an event, the hierarchy of these views is updated to reflect the Pocket requirement from Snowplow. The view hierarchy is inside-out, where the deepest nested view is hierarchy 0, and the top level n.

## GUID

The persistent `SnowplowUser` context requires the addition of a server-generated `guid`. This `guid` is generated at sign-in as a pre-requesite to authenticating with Pocket, at which point it is always used when signing in. It is then added to all subsequent GraphQL queries as well.

### DIFF
--- a/Pocket.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/Pocket.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -20,6 +20,15 @@
         }
       },
       {
+        "package": "FMDB",
+        "repositoryURL": "https://github.com/ccgus/fmdb",
+        "state": {
+          "branch": null,
+          "revision": "61e51fde7f7aab6554f30ab061cc588b28a97d04",
+          "version": "2.7.7"
+        }
+      },
+      {
         "package": "InflectorKit",
         "repositoryURL": "https://github.com/apollographql/InflectorKit",
         "state": {
@@ -62,6 +71,15 @@
           "branch": null,
           "revision": "f7c3de2d9fc09d7a61f7259521d97091fb2f8bda",
           "version": "7.1.3"
+        }
+      },
+      {
+        "package": "SnowplowTracker",
+        "repositoryURL": "https://github.com/snowplow/snowplow-objc-tracker",
+        "state": {
+          "branch": null,
+          "revision": "2cf806e834a75ce801c46f0182f23b481655762b",
+          "version": "2.2.1"
         }
       },
       {
@@ -123,8 +141,8 @@
         "repositoryURL": "https://github.com/apple/swift-nio.git",
         "state": {
           "branch": null,
-          "revision": "d79e33308b0ac83326b0ead0ea6446e604b8162d",
-          "version": "2.30.0"
+          "revision": "9a992ee3de1f8da9f2968fc96b26714834f3105f",
+          "version": "2.31.1"
         }
       },
       {
--- a/Pocket.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/Pocket.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -6,8 +6,8 @@
         "repositoryURL": "https://github.com/apollographql/apollo-ios.git",
         "state": {
           "branch": null,
-          "revision": "84e28febb0a6cc3b4b08968c0d7043afd288004e",
-          "version": "0.45.0"
+          "revision": "f3d9f55144b636a2569f938f5969e50174d39009",
+          "version": "0.46.0"
         }
       },
       {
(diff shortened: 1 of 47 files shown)


jacobsimeon (approved):
(discussion happened in person)

Great work @dskuza!

Couple of parting thoughts for the future:

- Lots of potential to reduce duplication in the auth client
- We're starting to see a lot of code surrounding authentication, access token storage, attaching GUID & userID to API requests, etc. I'd like to take a step back and see if we can situate these bits under a single-unified...something 🤔

PR #40, opened 2021-07-29
Replace SwiftUI article view with UIKit-based version

jacobsimeon:
- Textile: add support paragraph and text alignment
- Resurrect code for generating NSAttributedString from TextContent
- Extend Header (particle component) to provide a Style for each header
  level
- Extend item to provide text alignment based on article language
- Extract Style+ReaderSettings
  This is a little helper extension that provides an adjusted style based
  on the given reader settings
- Add UICollectionView+registration
  A small extension that makes it a tad bit easier to register/dequeue
  cell classes in a collection view.
- Add TextContentCell
  This is a collection view cell that contains a single UITextView. The
  text view will be populated with an attributed string and displayed in
  an article view.
- Add EmptyCell
  This is a placeholder cell that is used for unsupported/unrecognized
  particle components.
- Remove SwiftUI implementation of Article view
- Rename PocketKit/Reader -> PocketKit/ReaderSettings
  This folder only contains the ReaderSettingsView and corresponding
  ReaderSettings model
- Remove unnecessary usages of `@available` and `canImport`
- Update UI tests to scroll down after verifying each particle component
- Allow Xcode to update Package.resolved


PocketKit/Sources/Textile/Style/Paragraph/ParagraphStyle.swift
```diff
@@ -0,0 +1,11 @@
+public struct ParagraphStyle {
```
dskuza, line 1: This is a great addition early-on. It'll come in super-handy once we include support for line height adjustment n' such.

PocketKit/Sources/Textile/Style/Style.swift
```diff
@@ -45,31 +49,35 @@ public struct Style {
...
     }
 
     public func with(slant: FontDescriptor.Slant) -> Style {
-        return Style(fontDescriptor: fontDescriptor.with(slant: slant), color: colorAsset, underlineStyle: underlineStyle, strike: strike)
+        return Style(fontDescriptor: fontDescriptor.with(slant: slant), color: colorAsset, underlineStyle: underlineStyle, strike: strike, paragraph: paragraph)
     }
 
     public func with(color: ColorAsset) -> Style {
-        return Style(fontDescriptor: fontDescriptor, color: color, underlineStyle: underlineStyle, strike: strike)
+        return Style(fontDescriptor: fontDescriptor, color: colorAsset, underlineStyle: underlineStyle, strike: strike, paragraph: paragraph)
```
dskuza, line 68: I _think_ we want to be passing `color` as a parameter to the initializer here, as the argument label is `color` in the function declaration. Pretty sure `colorAsset` is the current color asset, so it wouldn't actually get updated in this case.

dskuza (changes requested):
Cool; simple enough! One little thing for you to look at.

PocketKit/Sources/Textile/Style/Style.swift
jacobsimeon: Nice catch!

dskuza (approved):

PR #41, opened 2021-07-29
Update button style to avoid deprecation warning

jacobsimeon:
This just gets rid of a deprecation warning. 

Stashing this change here since Bitrise builds won't pass until they update to Xcode 13 beta 4.


dskuza (approved):

dskuza:
Closing this pull request as the same fix is made in #37, so we don't need to add any unnecessary duplicate work!

PR #42, opened 2021-07-29
Don't capture messages when encountering unsupported Particle elements

jacobsimeon:
These captures are causing a huge performance issue so disabling for
now. We can come up with a way to capture them in a more efficient
manner in the future.


dskuza (approved):
:+1: Let's add a ticket to the backlog to investigate this further.

PR #43, opened 2021-07-30
Include item URL and title when sharing selected text

jacobsimeon:
share-demo

https://getpocket.atlassian.net/browse/IN-93


dskuza (approved):
Bummer that overriding a private function is the way to go, but… eh, what can you do.

PR #44, opened 2021-07-30
Remove macOS Target

jacobsimeon:
This will be replaced by a catalyst-powered version of the iPadOS app.

Created [this ticket][0] to track the work to add catalyst support.

[0]: https://getpocket.atlassian.net/browse/IN-202


dskuza (commented):
Can / should we remove the macOS platform from `Package.swift`?

jacobsimeon:
We actually need the macOS platform so that we can build the ApolloCodegen stuff. Might be better to split that into its own separate package

dskuza (approved):

PR #47, opened 2021-08-16
Support deleting an item from My List

jacobsimeon:
https://getpocket.atlassian.net/browse/IN-174


PocketKit/Sources/PocketKit/Item/ItemListView.swift
```diff
@@ -42,6 +42,11 @@ struct ItemListView: View {
                     .swipeActions {
                         favoriteButton(for: item)
                             .tint(Color(.branding.amber3))
+                        Button {
+                            source.delete(item: item)
+                        } label: {
+                            Label("Delete", systemImage: "trash")
+                        }.tint(.red)
```
dskuza, line 49: I kind of like how the favorite button is extracted into its own convenience function, and feel like maybe it would "clean things up" a little, here, even though it's quite obvious what's being done 🙃

PocketKit/Sources/Sync/Operations/DeleteItem.swift
```diff
@@ -0,0 +1,36 @@
...
+
+class DeleteItem: AsyncOperation {
+    private let apollo: ApolloClientProtocol
+    private let events: PassthroughSubject<SyncEvent, Never>
+    private let itemID: String
+
+    init(
+        apollo: ApolloClientProtocol,
+        events: PassthroughSubject<SyncEvent, Never>,
+        itemID: String
```
dskuza, line 14: I feel like this is an initializer we'll see for… most? all? item mutations. Ideas:

1. Define a protocol
2. Subclass `AsyncOperation` to `AsyncItemMutation` or something, and then just have a subclass fill in `main`
3. None of the above!

I'm alright with any of the three.

dskuza (approved):
All the changes from #45 are making this look real clean!

Just a few comments, but nothing blocking whatsoever.

PocketKit/Sources/Sync/Operations/DeleteItem.swift
jacobsimeon: Agree. We are starting to see a lot of duplication here. I am working on the Archive action now and will be refactoring to eliminate it as part of that PR. Sound good?

dskuza: Absolutely!

jacobsimeon:
@dskuza I went ahead and added the archive action to this PR, along with a refactor to eliminate all the duplication for item actions.

Now you can just create a generic `ItemMutataionOperation`, passing in the graphQL mutation you want to perform instead of having a separate operation for each mutation.

Mind taking another look?

dskuza (approved):

PR #50, opened 2021-08-18
Add item actions to reader view

jacobsimeon:
- Share
- Archive
- Delete
- Favorite/Unfavorite

https://getpocket.atlassian.net/browse/IN-180
https://getpocket.atlassian.net/browse/IN-177
https://getpocket.atlassian.net/browse/IN-179
https://getpocket.atlassian.net/browse/IN-176

This one is dependent on #49. So let's hold off on review until that one gets approved/merged :)


PocketKit/Sources/PocketKit/Item/ItemViewController.swift
```diff
@@ -68,7 +144,53 @@ class ItemViewController: UIViewController {
...
+
+// MARK: - Item Actions
+
+extension ItemViewController {
+    private func favorite() {
+        guard let item = itemHost.item else {
+            return
+        }
+
+        source.favorite(item: item)
```
dskuza, line 160: Curious on your thoughts here: for something like `showReaderSettings`, we tell the _delegate_ to perform the action, but here, the view controller itself is unfavoriting the item. Do we want the current instance of `ItemViewController` to (un)favorite the item (as well as the other actions), or would we rather push that off to the delegate?

dskuza (approved):
Awesome. Just one non-blocking question for ya!

PocketKit/Sources/PocketKit/Item/ItemViewController.swift
jacobsimeon: I struggled with this issue too. I settled on calling the `source` method here in the view controller. The reason being that the VC's responsible for performing actions but it delegates the work of presenting or showing other views.

PR #51, opened 2021-08-18
Track analytics events for item actions

jacobsimeon:
This builds on #49 and #50 so let's get those reviewed and merged before diving to deep into discussing this one

https://getpocket.atlassian.net/browse/IN-149
https://getpocket.atlassian.net/browse/IN-150

### DIFF
--- a/PocketKit/Sources/PocketKit/Item/ItemViewController.swift
+++ b/PocketKit/Sources/PocketKit/Item/ItemViewController.swift
@@ -15,10 +15,10 @@ protocol ItemViewControllerDelegate: AnyObject {
 class ItemViewController: UIViewController {
     private let itemHost: ArticleViewController
     private let source: Source
+    private let tracker: Tracker
     private var moreButtonItem: UIBarButtonItem?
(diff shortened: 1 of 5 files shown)


PocketKit/Sources/PocketKit/Item/ItemViewController.swift
```diff
@@ -68,7 +145,72 @@ class ItemViewController: UIViewController {
...
+    }
+
+    private func track(identifier: UIIdentifier, item: Item) {
+        var eventContexts = [
+            uiContext,
+            UIContext.button(identifier: identifier)
+        ]
+
+        if let url = item.url {
+            eventContexts.append(Content(url: url))
```
dskuza, line 210: I think we should `guard` against `url` here before tracking; tracking engagement (favorite, archive, etc) without the context of its content (in this scenario specifically) doesn't seem valuable.

dskuza (commented):
I believe I picked out all the analytics-specific changes with my eyes! One comment I'd like you to address before approval.

dskuza (approved):

PR #53, opened 2021-08-18
Fix sending impression events with no content

dskuza:
# Fixes

1. When scrolling a user's list, impression events are fired whenever a row comes on-screen. These impression events should _only_ be sent if there is a `content` context to send along with the event. Adding a guard prevents sending impression events incorrectly.
2. To determine the context of (failing) events, the `client_version` property has been added to `APIUser`, where the property refers to the build number of the app. This context is sent with each event tracked.

### DIFF
--- a/PocketKit/Sources/PocketKit/Item/ItemListView.swift
+++ b/PocketKit/Sources/PocketKit/Item/ItemListView.swift
@@ -137,6 +137,11 @@ private struct ItemListViewRow: View {
             )
                 .trackable(.home.item(index: UInt(index)))
                 .onAppear {
+                    let contexts = contexts
+                    guard contexts.count > 1 else {
+                        return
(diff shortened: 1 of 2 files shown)


PocketKit/Sources/PocketKit/Item/ItemListView.swift
```diff
@@ -137,6 +137,11 @@ private struct ItemListViewRow: View {
             )
                 .trackable(.home.item(index: UInt(index)))
                 .onAppear {
+                    let contexts = contexts
```
dskuza, line 140: `contexts` is a (simple) computed property. Despite it being simple, this assignment is to ensure we're only running the computation once.

jacobsimeon (approved):
😎

PR #62, opened 2021-09-16
Foxfooding Bug Fixes

dskuza:
1. Fixes a crash when applying (some) modifiers. Range inconsistencies post-unescaping would in some cases cause a crash. This is a temporary resolution that will be more properly rectified during our upcoming work in article view.
2. Adds support for disabling Sentry via launch arg. This code went poof! So, let's bring it back.
3. Adds support for copying an item link, as well as the link and any selected text when viewing an article.
  - Currently, the new share support exists within the article view, since the work is "skeleton" work that will have to be refactored once merged with other work in active development.
  - The list (to be fixed post-refactor) will currently only copy the item URL when sharing.


jacobsimeon (approved):
Nice. Well done 👍

PR #64, opened 2021-09-23
Add missing license headers

dskuza:
This pull request adds missing license headers. I believe this is a side-effect of our change to a dedicated Swift package, but that's okay!


jacobsimeon:
So this if one and all but a couple of things come to mind:

1. Is it really necessary (from a legal perspective) to have these three lines of text at the top of every file. The license file at the root of the repository should suffice, no? I find the comments to be just plain noisy.
2. If we really do need these headers, is there a way we can automate adding them? I'm thinking a simple script that scans the repo for any `.swift` file and adds the header if not present.

dskuza:
I think we should resolve the question of "do we have to?" before moving on with this pull request. Closing for now.

PR #69, opened 2021-09-29
Apply various UI tweaks to mimic legacy styling

dskuza:
- [x] Apply the legacy background color to all views
- [x] Apply the correct navigation and tab bar tint colors, as well as their respective items
- [x] Apply common styling for article view, with reference from the item list
- [x] Hide the tab bar when the reader is presented in compact mode (as there are no actions to take from there)
- [x] Remove extra spacing caused by nil / empty titles
- [x] Improve the sidebar experience on iPad, including proper cell selection


PocketKit/Sources/PocketKit/Main/RegularMainCoordinator.swift
```diff
@@ -40,23 +40,46 @@ class RegularMainCoordinator: NSObject {
...
         splitController.setViewController(myList, for: .supplementary)
         splitController.setViewController(item, for: .secondary)
+        
+        splitController.view.tintColor = UIColor(.ui.grey1)
+
+        navigationSidebar.navigationBar.prefersLargeTitles = true
+        navigationSidebar.navigationBar.barTintColor = UIColor(.ui.white1)
+        navigationSidebar.navigationBar.tintColor = UIColor(.ui.grey1)
+
+        myList.navigationController?.navigationBar.prefersLargeTitles = true
```
jacobsimeon, line 72: kinda sad that we have to wait until these have been added to the split before we can access the nav controller. As an alternative, we could preemptively wrap `myList` and `home` in navigation controllers that we own. Not a big deal but would be a bit prettier than having to dig so far down to get to the navigation bar (and wouldn't need to unwrap the optional).

That said, it's not a huge deal so I'll leave it to your discretion.

PocketKit/Sources/PocketKit/Navigation/NavigationSidebarViewController.swift
```diff
@@ -0,0 +1,98 @@
...
+    private let model: MainViewModel
+    private var subscriptions: Set<AnyCancellable> = []
+
+    init(model: MainViewModel) {
+        self.model = model
+        super.init(nibName: nil, bundle: nil)
+        
+        self.title = "Pocket"
+        
+        collectionView.alwaysBounceVertical = true
```
jacobsimeon, line 70: is this not the default?

PocketKit/Sources/PocketKit/Navigation/NavigationSidebarViewController.swift
```diff
@@ -0,0 +1,98 @@
...
+        return UICollectionViewCompositionalLayout.list(using: configuration)
+    }()
+    
+    private static let snapshot: NSDiffableDataSourceSnapshot<MainViewModel.AppSection, String> = {
+        var snapshot = NSDiffableDataSourceSnapshot<MainViewModel.AppSection, String>()
+        snapshot.appendSections(MainViewModel.AppSection.allCases)
+        MainViewModel.AppSection.allCases.forEach {
+            snapshot.appendItems([$0.navigationTitle], toSection: $0)
+        }
+        return snapshot
```
jacobsimeon, line 24: Let's not conflate the term "app section" with "collection view section" here. This collection only needs a single section with an item for each app section.

Something like this should be all you need:
```swift
var snapshot = NSDiffableDataSourceSnapshot<String, MainViewModel.AppSection>()
let section = "app-sections"
snapshot.appendSections([section])
snapshot.appendItems(MainViewModel.AppSection.allCases, toSection: section)
```

PocketKit/Sources/PocketKit/Navigation/NavigationSidebarViewController.swift
```diff
@@ -0,0 +1,98 @@
...
+        collectionView.alwaysBounceVertical = true
+        collectionView.delegate = self
+        collectionView.dataSource = dataSource
+        dataSource.apply(Self.snapshot)
+        
+        model.$selectedSection
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] section in
+                let indexPath = self?.dataSource.indexPath(for: section.navigationTitle)
+                self?.collectionView.selectItem(at: indexPath, animated: false, scrollPosition: .top)
```
jacobsimeon, line 79: 1. I'd suggest pulling this bit of code into a private function so we don't have to unwrap self a bunch of times. I get that it's only two lines, but still. It also helps avoid passing optionals into the `collectionView.selectItem(at:)` method.
2. See not above regarding building the snapshot. Getting the indexPath could be something like:
```
let indexPath: IndexPath = [0, MainViewModel.AppSection.allCases.index(of: section)]
```

PocketKit/Sources/Textile/Textiles.swift
```diff
@@ -18,14 +18,6 @@ private extension Style {
...
     public static func initialize() {
         loadFonts()
-        
-        if #available(iOS 1.0, *) {
-            UINavigationBar.appearance().titleTextAttributes = Style.navigationTitle.textAttributes
-            UINavigationBar.appearance().largeTitleTextAttributes = Style.largeNavigationTitle.textAttributes
-            
-            UIToolbar.appearance().barTintColor = UIColor(ColorAsset.ui.white1)
-            UIToolbar.appearance().tintColor = UIColor(ColorAsset.ui.grey1)
-        }
```
jacobsimeon, line 28: Good riddance 👋

jacobsimeon (changes requested):
Really nice visual improvements throughout the app! It all makes the whole experience feel much better.

Regarding the collection view in the navigation sidebar, I think we need to make another pass at it and try to simplify. Mainly, I don't think we need to create a section in the list for each app section. And second, digging into the current snapshot of the collection view's datasource feels backwards to me.

We should have our own data model that we can dig into (as with any other collection view). That data model happens to be `MainViewModel.AppSection.allCases`. So any time we get a callback from UIKit, we should be able to access the relevant data simply by using the provided indexPaths–without having to dig into the snapshot. (see comments throughout)

Let me know if you want to pair on this!

PocketKit/Sources/PocketKit/Navigation/NavigationSidebarViewController.swift
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) is not implemented")
+    }
+}
+
+extension NavigationSidebarViewController: UICollectionViewDelegate {
+    func collectionView(_ collectionView: UICollectionView, didSelectItemAt indexPath: IndexPath) {
+        let section = dataSource.snapshot().sectionIdentifiers[indexPath.section]
```
jacobsimeon, line 95: See above note regarding building the snapshot, but we should be able to set the app section like so:

```swift
model.selectedSection = MainViewModel.AppSection.allCases[indexPath.item]
```

PocketKit/Sources/PocketKit/Navigation/NavigationSidebarViewController.swift
dskuza: Oddly enough, nope - docs state that the default value is `false` (I did have to double check, as I thought the same thing).

PocketKit/Sources/PocketKit/Navigation/NavigationSidebarViewController.swift
dskuza: @jacobsimeon Fair statement. My initial thought behind multiple collection view sections for each app section was the possibility of this type of sidebar design:

image

Which, in the current state of this project, results in:

image

This could probably be modified to use one section with item spacing, however at the time, this did seem more like an inset grouped layout, and letting the system handle the rest.

dskuza:
@jacobsimeon Thanks for the thorough review! I think there are some valuable discussions we can have, so let's take a short time to pair on this, as you mentioned. Shouldn't be too long, but let's discuss our thoughts!

jacobsimeon (approved):

PR #71, opened 2021-10-04
Separate saved item

jacobsimeon:


PocketKit/Sources/PocketKit/Article/Item+Readable.swift
```diff
@@ -1,6 +1,16 @@
 import Sync
+import Foundation
+
+
+extension SavedItem: Readable {
+    var readerURL: URL? {
+        item?.resolvedURL ?? item?.givenURL ?? url
```
dskuza, line 7: Looks like this mimics `SavedItem.bestURL`. Potentially a place for piggybacking off of it.

dskuza (approved):
Awesome. This definitely follows closer to the GraphQL schema, which does indeed make it more readable and understandable. Great idea!

PR #73, opened 2021-10-05
Track card impressions on main home screen

dskuza:
A few things that this pull request tackles: 

- Moving from a "single" tracker to child trackers that can track within their own context, and forward up to a parent (thus building a context "stack", if you will, before the parent performs the final tracking)
- Moving from `fetchSlates()` to `fetchSlateLineup(_:)`
  - One of the requirements when tracking an impression is information about the slate lineup itself, as well as the slate, and recommendation viewed. We were currently only fetching the slates from a specific lineup, so the change here makes it easier to pull "any" lineup, and receive its details.
- Project (file) cleanup


PocketKit/Sources/PocketKit/Home/HomeViewController.swift
```diff
@@ -63,14 +69,14 @@ class HomeViewController: UIViewController {
         super.viewDidLoad()
 
         Task {
-            let slates = try? await source.fetchSlates()
-            setSlates(slates: slates)
+            let lineup = try? await source.fetchSlateLineup(Self.lineupID)
+            self.slateLineup = lineup
```
jacobsimeon, line 73: this should either call `setSlateLineup` or we should get rid of that function. I'm really not sure if the setSlateLineup function is necessary. I had only put it so that we could execute that code on the main thread. However, I'm pretty sure that `Task {}` will resume on the thread it was originally started on so it may not be necessary.

PocketKit/Sources/Sync/Slates/Slate+remoteMapping.swift
```diff
@@ -3,10 +3,24 @@ import Foundation
 
 extension Slate {
     typealias Remote = SlateParts
+    typealias GetSlateLineupRemote = GetSlateLineupQuery.Data.GetSlateLineup.Slate
```
jacobsimeon, line 6: This bit shouldn't be necessary, you can access the `SlateParts` fragment from the `GetSlateLineupQuery.Data.GetSlateLineup.Slate` object and initialize using the fragment instead. (that way it doesn't matter what query we're using)

PocketKit/Tests/SyncTests/SourceTests.swift
```diff
@@ -165,18 +165,40 @@ class SourceTests: XCTestCase {
...
-        let actualSlates = try await subject().fetchSlates()
+        let actualLineup = try await subject().fetchSlateLineup("")
 
-        XCTAssertEqual(actualSlates, expectedSlates)
+        XCTAssertEqual(actualLineup, expectedLineup)
     }
 
     func test_fetchSlate_returnsResultFromSlateService() async throws {
-        let expectedSlate = Slate(id: "my-slate", name: "My Slate", description: "My very awesome slate", recommendations: [])
+        let expectedSlate = Slate(id: "my-slate",
```
jacobsimeon, line 194: newline before after opening paren here.

```swift
Slate(
    id:
```

jacobsimeon (changes requested):
Looks good. Just a couple of comments.

Only thing that needs changed is the code that maps the graphQL responses into `SlateLineup`, `Slate`, `Slate.Recommendation`, etc. We should be using the fragments there, otherwise we have to a bunch of repetitive work to map responses from each query.

PR #74, opened 2021-10-07
Use diffable data source in HomeViewController

jacobsimeon:


PocketKit/Sources/PocketKit/Home/HomeViewController.swift
```diff
@@ -16,7 +26,24 @@ class HomeViewController: UIViewController {
 
     private var slates: [Slate]? {
         didSet {
-            collectionView.reloadData()
+            var snapshot = NSDiffableDataSourceSnapshot<HomeSection, HomeItem>()
+
+            guard let slates = slates else {
+                dataSource.apply(snapshot)
```
dskuza, line 32: This is equivalent to saying that the new data source is completely empty, which would then render "nothing", correct? If that's true, is there an edge case where there _was_ some data present, "something happened" to return no slates, and the previous content disappears?

dskuza (approved):
:+1: Simple enough change. Just one comment, but it doesn't necessarily prevent this from being merged.

PR #81, opened 2021-10-14
Add ability to report recommendations 

dskuza:
This pull request adds support for reporting recommendations from both the main Home view, as well as a slate detail view.

- Adds `ReportRecommendationView`
  - SwiftUI view that supports selecting from various reasons, optionally supplying custom text as part of the report.
  - Reporting is performed via Snowplow
- Adds `ReportRecommendationHostingController`
  - Tweaks various UIKit element appearances when contained within an instance of this hosting controller
- Synchronization between compact and regular modes via view model
- Various updates to Analytics-related types
- Updated trackers for various views to be the correct linked trackers
- Addition of new image and color assets
- UI test for verifying visibility and reporting from the new view


PocketKit/Sources/Analytics/UIContext.swift
```diff
@@ -84,6 +87,8 @@ public extension UIContext {
     struct MyList {
         public let screen = UIContext(type: .screen, hierarchy: 0, identifier: .myList)
         
+        public var report = UIContext(type: .dialog, hierarchy: 0, identifier: .reportItem)
+
```
jacobsimeon, line 91: Eliminate this in favor of the generic version

PocketKit/Sources/PocketKit/Home/RecommendationOverflowButton.swift
```diff
@@ -8,7 +8,7 @@ class RecommendationOverflowButton: UIButton {
 
         configuration = .plain()
         configuration?.contentInsets = .zero
-        configuration?.image = UIImage(asset: .verticalOverflow)?
+        configuration?.image = UIImage(asset: .alert)
```
jacobsimeon, line 11: Once we want to re-use this button w/ different icons we can make the image a param to the initializer

Tests iOS/ReportARecommendationTests.swift
```diff
@@ -0,0 +1,93 @@
...
+        report.submitButton.wait().tap()
+        
+        wait(for: [reportExpectation], timeout: 1)
+        guard let requestBody = requestBody else {
+            XCTFail("Expected request body to not be nil")
+            return
+        }
+        
+        XCTAssertTrue(requestBody.contains("reason"))
+    }
```
jacobsimeon, line 92: 🥇

jacobsimeon (approved):

Tests iOS/ReportARecommendationTests.swift
```diff
@@ -0,0 +1,100 @@
...
+            [
+                "accessToken": "test-access-token",
+                "sessionGUID": "session-guid",
+                "sessionUserID": "session-user-id",
+            ]
+        ) { _, new in new }
+        app.strictLaunch(
+            arguments: arguments,
+            environment: environment
+        )
```
jacobsimeon, line 54: I feel like we can come up with a better abstraction here. What if we had like a `LaunchArguments` struct. It could have a property for each argument with defaults for each property. Then you could just initialize the struct and specify the properties you want to change.

Something like this:
```swift
struct PocketAppElement {
  init(args: LaunchArguments) {
    // map each property of launchArguments to an array
  )
}

struct LaunchArgument {
  let clearKeychain: Bool
  let clearCoreData: Bool
  let clearImageCache: Bool
  let disableSentry: Bool
  let disableSnowplow: Bool

  init(
    clearKeychain: Bool = true, 
    clearCoreData: Bool = true, 
    clearImageCache: Bool = true, 
    disableSentry: Bool = true,
    disableSnowplow: Bool = true
}

// then at the call site you can specify only the arg you care about
// and allow the rest to use defaults
let args = LaunchArguments(disableSnowplow: false)
app.launch(args: args)
```

Thinking we could do something similar for the launch env vars as well

PR #88, opened 2021-11-05
Initial marticle support for divider components

dskuza:
- [x] Create cell for divider components
- [x] Return the correct height for a divider component item

This pull request sets the height of a divider component cell to `16`, which is some number that I thought looked nice. This height does _not_ account for the inter-item spacing; that will be visible _in addition to_ the height of this cell. The divider itself within the view is 1pt.


jacobsimeon (approved):
nice!

PR #89, opened 2021-11-09
Initial MArticle support for code blocks

dskuza:
This pull request adds initial MArticle support for code blocks.

- [x] Create `CodeBlockComponentCell`
- [x] Register and utilize `CodeBlockComponentCell` for layout
- [x] Add support for presenting code blocks to `ArticleComponentPresenter`

_How_ the code block is presented is … interesting. `UITextView` does _not_ like being told what its content size is for long horizontal text. Even if you do manage to set its width appropriately, it refuses to draw past the bounds of the text view. One might think, "oh, I'll just update the text container size". Well, that doesn't work either.

SO.

There is a non-scrollable `UITextView` that dynamically sizes itself within a scroll view, which is itself sized to the height of the cell. Therefore, the text view's height will always be correct, its width will expand to fit the content, thus causing the parent scroll view to update its content size and allow for horizontal scrolling of long text. Fun times.

### DIFF
--- a/PocketKit/Sources/PocketKit/Article/ArticleComponentPresenter.swift
+++ b/PocketKit/Sources/PocketKit/Article/ArticleComponentPresenter.swift
@@ -11,6 +11,8 @@ private extension Style {
         .with(slant: .italic)
 
     static let imageCaption: Self = .body.sansSerif.with(size: .p3)
+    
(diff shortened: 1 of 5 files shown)


jacobsimeon (approved):
```swift
codeblocks.are(🆒)
```

PR #91, opened 2021-11-15
Add support for MArticle bulleted lists

dskuza:
This pull requests adds (initial) support for MArticle bulleted lists by piggybacking off of previously implemented functionality. Since bulleted lists are comprised of rows that are Markdown components, we can create a styled attributed string for each row's content, indent it appropriately, and then join all rows into one attributed string to then present in a text view. 

With this pull request, it appears that we've solidified a pattern for adding components:
1. Ensure that `isEmpty` returns the correct value for your `ArticleComponent` case
2. Return the component's appropriate size in the `size(fittingWitdh:)` function in `ArticleComponentPresenter`
3. Add or modify an overloaded `present(component:in:...)` function in `ArticleComponentPresenter`
4. Create and register any necessary collection view cell classes
5. Return the correct cell for the component, calling the correct presenter function for rendering.

This pull request refactors a teeny bit of code to follow more closely to the above.

### DIFF
--- a/PocketKit/Sources/PocketKit/Article/ArticleComponentPresenter.swift
+++ b/PocketKit/Sources/PocketKit/Article/ArticleComponentPresenter.swift
@@ -2,6 +2,7 @@ import Sync
 import UIKit
 import Kingfisher
 import Textile
+import Down
 
 
 private extension Style {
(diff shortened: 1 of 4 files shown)


dskuza:
This pull request's work will be combined with numbered list work, the sum of which will have its own PR created.

PR #92, opened 2021-11-16
Initial support for MArticle bulleted and numbered lists

dskuza:
This pull requests adds (initial) support for MArticle bulleted and numbered lists by piggybacking off of previously implemented functionality. Since these lists are comprised of rows that are Markdown components, we can create a styled attributed string for each row's content, indent it appropriately, and then join all rows into one attributed string to then present in a text view. 

With this pull request, it appears that we've solidified a pattern for adding components:
1. Ensure that `isEmpty` returns the correct value for your `ArticleComponent` case
2. Return the component's appropriate size in the `size(fittingWitdh:)` function in `ArticleComponentPresenter`
3. Add or modify an overloaded `present(component:in:...)` function in `ArticleComponentPresenter`
4. Create and register any necessary collection view cell classes
5. Return the correct cell for the component, calling the correct presenter function for rendering.

This pull request refactors a teeny bit of code to follow more closely to the above.


PocketKit/Sources/PocketKit/Article/ArticleComponentPresenter.swift
```diff
@@ -13,14 +14,16 @@ private extension Style {
     static let imageCaption: Self = .body.sansSerif.with(size: .p3)
     
     static let codeBlock: Self = .body.monospace
+    static let bulletedPrefix: Self = .body.monospace
+    static let standardPrefix: Self = .body.sansSerif
```
jacobsimeon, line 18: Looks like we can get rid of these declarations since they're not being used anymore

jacobsimeon (approved):

PR #102, opened 2021-12-20
Update Article title & metadata cell in reader

jacobsimeon:
This is the first change in a larger set of changes to design & styling of the native reader view. I'll be creating all PRs against the `marticle-updates` branch and using separate PRs for each component. Once each component PR is merged, we can review all the changes together before merging into `develop`


dskuza (approved):
Awesome. The Textile API has definitely evolved into something nice to use, imo.

PR #104, opened 2021-12-22
Reader: Update image component layout

jacobsimeon:
- Caption and Credit are now combined into a single text view (using different attributes for each).
- Images have a bottom margin of 24pt

Also used a different approach to handling of line height multipliers that is more in line with how UIKit does things.

### Example
image-example


dskuza (approved):
Nice. I like not having to additionally pass in a font descriptor for paragraph styling. I'm glad we also decided to split into per-component presenters; it definitely improves readability, as well as making the value of each presenter more succinct.

PR #105, opened 2021-12-22
Marticle everything else

jacobsimeon:
Take a look at #104 before this one!

The only new commit is the last one: 5deff4d3d9ac38205695b16e09ababc1909760bc


dskuza:
@jacobsimeon Could you post a screenshot of an example blockquote with the new styling?

jacobsimeon:
Here's a few screenshots (including updated block quote styling (which was updated _again_))
example-1
example-1x-images
example-2
example-3
example-4

dskuza (approved):

PR #111, opened 2022-01-19
Fix incorrect background colors in Home and My List

dskuza:
Two simple fixes:

1. The background color of the Home collection view was not set, so the resulting color would be the system background rather than the "Pocket" background.
2. After the "My List" refactor to use the modern APIs, it seems as if the collection view was adding implicit decoration items for each section. By explicitly setting each section's `decorationViews` to `[]`, we see the expected color when scrolling to the bottom bottom of the list, as no additional views are rendered as the "background" to each section.


jacobsimeon (approved):
LGTM (Let's Go Team Mozilla)

PR #117, opened 2022-01-28
Refactor ReadableViewModel to send events over delegation

dskuza:
1. Move from using a ReadableViewModelDelegate to using published events
2. Move tracking from ReadableHostViewController to its related view models
3. Move URL opening into ReadableViewModels
    - Tracking URL opens and sharing selected text were actually broken 🙃 

There's a new pattern I've introduced into the view models for the event publishers:

```swift
private var _events: PassthroughSubject<ReadableEvent, Never> = .init()
var events: AnyPublisher<ReadableEvent, Never> {
    _events.eraseToAnyPublisher()
}
```

The reason I introduced this pattern is to add a little more safety with regard to _who_ can send an event. `events` here is a "read-only" publisher, whereas its underlying implementation lets us send events appropriately (whose implementation detail is also now hidden).

### DIFF
--- a/PocketKit/Sources/PocketKit/Article/Cells/ArticleComponentTextCell.swift
+++ b/PocketKit/Sources/PocketKit/Article/Cells/ArticleComponentTextCell.swift
@@ -11,7 +11,7 @@ protocol ArticleComponentTextCellDelegate: AnyObject {
 // An object that conforms to this protocol is capable of delegating actions
 // commonly performed within the cell, typically interactions with a PocketTextView.
 protocol ArticleComponentTextCell: ArticleComponentTextViewDelegate {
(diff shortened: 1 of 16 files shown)


jacobsimeon (changes requested):
As discussed over zoom:

1. Move delete and share actions into the view model
2. Reduce event enum to just `.archive` and `.delete`
3. Avoid duplicating code by using a protocol w/ extension

jacobsimeon (approved):
Nice work! Let's squash these commits into a single commit before merging.

I'm not actually familiar with Github's "Squash and merge" feature. It's currently disabled but might be ok to enable that one? Do you know if it creates a merge commit?

PR #118, opened 2022-02-01
ItemsList: supply item actions from view model

jacobsimeon:
This lays the ground work so that `ArchivedItemsListViewModel` can supply different actions from what `MyListItemsListViewModel` uses.

One big question is around the re-use of `ReadableAction`. I went ahead and adopted it for this use case but I'm also wondering if 1) a separate struct should be use to model the actions or 2) if we can simply supply `UIAction` objects directly.


PocketKit/Sources/PocketKit/MyList/ItemsList/ItemsListViewModel.swift
```diff
@@ -35,10 +35,11 @@ protocol ItemsListViewModel: AnyObject {
...
     func selectCell(with: ItemsListCell<ItemIdentifier>)
-    func shareItem(with: ItemsListCell<ItemIdentifier>)
 
-    func toggleFavorite(_ cell: ItemsListCell<ItemIdentifier>)
-    func archive(_ cell: ItemsListCell<ItemIdentifier>)
-    func delete(_ cell: ItemsListCell<ItemIdentifier>)
+    func shareAction(for objectID: ItemIdentifier) -> ReadableAction?
+    func favoriteAction(for objectID: ItemIdentifier) -> ReadableAction?
+    func overflowActions(for objectID: ItemIdentifier) -> [ReadableAction]?
+    func swipeActions(for objectID: ItemIdentifier) -> [UIContextualAction]
```
dskuza, line 42:
```suggestion
    func trailingSwipeActions(for objectID: ItemIdentifier) -> [UIContextualAction]
```

A small detail, but at least in our current usage, the swipe actions we provide are for the trailing side of the cell. Being a little more explicit in the API might just help in clarity, especially if we may want to add leading actions one day.

PocketKit/Sources/PocketKit/MyList/ItemsList/ItemsListItemCell.swift
```diff
@@ -241,11 +202,24 @@ extension ItemsListItemCell {
...
-        favoriteButton.configuration?.image = state.model?.favoriteButtonImage
-        favoriteButton.accessibilityLabel = state.model?.favoriteButtonAccessibilityLabel
 
-        thumbnailView.image = nil
+        favoriteButton.accessibilityLabel = state.model?.favoriteAction?.title
+        favoriteButton.accessibilityIdentifier = state.model?.favoriteAction?.accessibilityIdentifier
+        favoriteButton.configuration?.image = state.model?.favoriteAction?.image
+
+        if let favoriteAction = UIAction(state.model?.favoriteAction) {
+            favoriteButton.addAction(favoriteAction, for: .primaryActionTriggered)
```
dskuza, line 211: Do we need to remove all actions before adding an action? I just want to verify that we won't be stacking actions for the same control event.

dskuza (commented):
Overall, I like these changes. 

I don't mind having these models return `ReadableAction` types, primarily because it keeps `UIKit` as an implementation detail of the view (controller) using it. It also allows us to do things like this:

```swift
favoriteButton.accessibilityLabel = state.model?.favoriteAction?.title
favoriteButton.accessibilityIdentifier = state.model?.favoriteAction?.accessibilityIdentifier
```

and keep these definitions within a single type, where `UIAction`'s API doesn't expose these sorts of details.

The above is considering that I'm understanding the PR's description correctly, and the question is whether to have the _models_ return these types. 

Added comment: if we are reusing these actions, maybe we need to have a conversation about the `Readable` name, and where / whether it makes sense to use.

Happy to go over this with you live, as always!

PocketKit/Sources/PocketKit/MyList/ItemsList/ItemsListItemCell.swift
jacobsimeon: From Apple's docs:

> UIActions are uniqued based on their identifier, and subsequent actions with the same identifier replace previously added actions.

So I take that to mean that we are safe here. This is why the `.favorite()` and `.unfavorite()` helpers use the same identifier.

jacobsimeon:
>  maybe we need to have a conversation about the Readable name

Yeah I agree. Ideas for names? I was thinking maybe `ItemAction`

dskuza:
`ItemAction` works great - that's what I was thinking, as well.

PocketKit/Sources/PocketKit/ItemAction.swift
```diff
@@ -0,0 +1,125 @@
...
+            image: UIImage(asset: .favoriteFilled)
+                .withTintColor(UIColor(.branding.amber4), renderingMode: .alwaysOriginal),
+            handler: handler
+        )
+    }
+
+    static func share(_ handler: @escaping () -> ()) -> ItemAction {
+        return ItemAction(
+            title: "Share",
+            // intentionally the same as `favorite()` since we want to replace
```
dskuza, line 86: Perhaps this is a byproduct of a copy/paste for (un)favorite actions.

dskuza (approved):

PocketKit/Sources/PocketKit/ItemAction.swift
jacobsimeon: yup

PR #119, opened 2022-02-01
Add support for fetching favorited archived items

dskuza:
This pull request adds support for viewing one's archive of favorited items.

The flow here is that once a user toggles `Favorite`, a blank snapshot will be rendered by the list, until items are fetched, at which point the (un)favorited items will be displayed. The implementation is very similar to that of `SavedItemsListViewModel`.

Behind the scenes, `Source` and its conforming types have been updated with the ability to supply `isFavorite` to the query used when fetching archived items.

Unit tests have been updated to ensure that the correct calls are being made, both through the `PocketArchiveService`, and `PocketSource`.

UI tests have been written to ensure that the toggle works as expected.

I've used the term "favorited" here as the term wrapping the `isFavorite` filter, but I'm happy to change the wording to be more consistent with the filter (which _does_ apply to a single item, but we're fetching a set of items).

### DIFF
--- a/PocketKit/Sources/PocketKit/MyList/ArchivedItemsList/ArchivedItemsListViewModel.swift
+++ b/PocketKit/Sources/PocketKit/MyList/ArchivedItemsList/ArchivedItemsListViewModel.swift
@@ -25,7 +25,6 @@ class ArchivedItemsListViewModel: ItemsListViewModel {
     }
     private var archivedItemsByID: [String: ArchivedItem] = [:]
 
-    @Published
     private var selectedFilters: Set<ItemsListFilter> = .init()
     private let availableFilters: [ItemsListFilter] = ItemsListFilter.allCases
 
--- a/PocketKit/Sources/PocketKit/MyList/ArchivedItemsList/ArchivedItemsListViewModel.swift
+++ b/PocketKit/Sources/PocketKit/MyList/ArchivedItemsList/ArchivedItemsListViewModel.swift
@@ -37,7 +37,8 @@ class ArchivedItemsListViewModel: ItemsListViewModel {
     }
 
     func fetch() async throws {
-        archivedItems = try await source.fetchArchivedItems()
+        let favorited = selectedFilters.contains(.favorites)
+        archivedItems = try await source.fetchArchivedItems(isFavorite: favorited)
 
         let snapshot = buildSnapshot()
         events.send(.snapshot(snapshot))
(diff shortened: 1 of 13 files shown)


PocketKit/Sources/PocketKit/MyList/ArchivedItemsList/ArchivedItemsListViewModel.swift
```diff
@@ -103,20 +117,43 @@ class ArchivedItemsListViewModel: ItemsListViewModel {
...
+        snapshot.appendItems(
+            ItemsListFilter.allCases.map { ItemsListCell<ItemIdentifier>.filterButton($0) },
+            toSection: .filters
+        )
+        
+        return snapshot
     }
+    
+    private func apply(filter: ItemsListFilter, from cell: ItemsListCell<ItemIdentifier>) {
+        if selectedFilters.contains(filter) {
```
jacobsimeon, line 134: So I like this approach. In `SavedItemsListViewModel` we actually have `selectedFilters` publish its changes and then we subscribe to those changes and do the actual fetch within that subscription.
It would be nice to take a consistent approach in both view models, would you mind updating `SavedItemsListViewModel` to follow this simpler implementation?

I could also see doing something like this:
```
var selectedFilters: [ItemsListFilter] {
  didSet { applyFilters() }
}

func applyFilters() {
  // apply the blank snapshot
  // fetch()
}
```

PocketKit/Sources/Sync/Source.swift
```diff
@@ -31,7 +31,7 @@ public protocol Source {
 
     func archive(recommendation: Slate.Recommendation)
 
-    func fetchArchivedItems() async throws -> [ArchivedItem]
+    func fetchArchivedItems(favorited: Bool) async throws -> [ArchivedItem]
```
jacobsimeon, line 34: Two thoughts here:

1. How about renaming `favorited` to `isFavorite` to stay consistent
2. How about making the `Bool` argument a `Bool?`. That way we'd be able to *theoretically* request items that are not favorited. Then when we ultimately build the `SavedItemsFilter`, we would only set the `isFavorite` argument if it is not nil.

PocketKit/Tests/SyncTests/PocketArchiveServiceTests.swift
```diff
@@ -15,15 +15,23 @@ class PocketArchiveServiceTests: XCTestCase {
...
+        
+        let unfavoritedCall: MockApolloClient.FetchCall<UserByTokenQuery> = apollo.fetchCall(at: 0)
+        XCTAssertEqual(unfavoritedCall.query.savedItemsFilter?.isArchived, true)
+        XCTAssertEqual(unfavoritedCall.query.savedItemsFilter?.isFavorite, false)
+        
+        _ = try await subject().fetch(accessToken: "the-access-token", favorited: true)
+        
+        let favoritedCall: MockApolloClient.FetchCall<UserByTokenQuery> = apollo.fetchCall(at: 1)
+        XCTAssertEqual(favoritedCall.query.savedItemsFilter?.isArchived, true)
+        XCTAssertEqual(favoritedCall.query.savedItemsFilter?.isFavorite, true)
```
jacobsimeon, line 29: 😎

Tests iOS/MyList/ArchiveFiltersTests.swift
```diff
@@ -0,0 +1,81 @@
...
+
+        let uiApp = XCUIApplication()
+        app = PocketAppElement(app: uiApp)
+
+        server = Application()
+
+        server.routes.post("/graphql") { request, _ in
+            let apiRequest = ClientAPIRequest(request)
+
+            if apiRequest.isForSlateLineup {
```
jacobsimeon, line 26: I could see ClientAPIRequest having a `requestType` (or some such) property that returns an enum. Then we could turn these `if else` statements into a `switch`. Not necessary now, but food for thought

Tests iOS/MyList/ArchiveFiltersTests.swift
```diff
@@ -0,0 +1,81 @@
...
+        myList.itemView(matching: "Favorited Archived Item 1").wait()
+        myList.itemView(matching: "Favorited Archived Item 2").wait()
+        
+        app.myListView.favoritesButton.tap()
+        
+        waitForDisappearance(of: myList.itemView(matching: "Favorited Archived Item 1"))
+        waitForDisappearance(of: myList.itemView(matching: "Favorited Archived Item 2"))
+        
+        myList.itemView(matching: "Archived Item 1").wait()
+        myList.itemView(matching: "Archived Item 2").wait()
```
jacobsimeon, line 78: 😎😎😎😎

jacobsimeon (approved):
Looks great, just a couple of (optional) thoughts/suggestions.

jacobsimeon (approved):
😎 cool

PR #120, opened 2022-02-03
Show offline view in Archive based on network connection

dskuza:
This pull request adds support for showing an offline view within the Archive list based on network connection.

A `fetch` call is made (currently) under two conditions:
1. Archive is loaded for the first time
2. Favorites are toggled on / off

**If there is no network connection**, then an "offline" snapshot will be sent to be rendered by `ItemsListViewController`. The action button ("Try Again") can be tapped, at which point another call to `fetch` will be made. If no network connection is available, the previous applies again.

**If there is a network connection**, a blank snapshot will be sent (i.e showing only the filter carousel), followed by sending an updated snapshot once the items are fetched from the backend. 

`fetch` calls are made when we want to fully reload the list with new data, which is why a blank snapshot is first sent, as a visual cue that items are being loaded. The two cases above are both fetches of new data that fully replaces what's on-screen, which is why I believe it's best to first send a blank snapshot before loading more items in. This also ensures that once a network connection is available, we do not see the "offline" view while data is being downloaded.

Light:
image

Dark:
image


jacobsimeon (approved):
Looks goot

PR #121, opened 2022-02-04
Use better navigation modeling

jacobsimeon:
- Eliminates the need to pass `MainViewModel` into everything
- Coordinators can simply observe their more focused view model, e.g.:
  CompactHomeCoordinator only has to care about the HomeViewModel.
- Simplifies top level coordinators by using nested coordinators


dskuza:
Reviewed live. ✅

PR #126, opened 2022-02-14
Sync: Pause operations while user is offline

jacobsimeon:
Move NetworkPathMonitor protocol to Sync framework and add some better
support updating the current network path within tests.

https://getpocket.atlassian.net/browse/IN-310


PocketKit/Sources/Sync/NetworkPathMonitor.swift
```diff
@@ -0,0 +1,39 @@
...
+
+    public var updateHandler: UpdateHandler? {
+        get {
+            return { (path: NetworkPath) in
+                guard let nwpath = path as? NWPath else {
+                    fatalError("Attempt to use \(path) as path for real \(Self.self). This is not allowed")
+                }
+
+                self.pathUpdateHandler?(nwpath)
+            }
```
jacobsimeon, line 34: Because of the way the protocols are set up for mocking, this is _theoretically_ possible. But in reality, we should never hit this since the system always calls the update handler with a proper `NWPath`

PocketKit/Sources/Sync/PocketSource.swift
```diff
@@ -84,6 +91,20 @@ public class PocketSource: Source {
...
+            switch path.status {
+            case .unsatisfied, .requiresConnection:
+                self.syncQ.isSuspended = true
+            case .satisfied:
+                self.syncQ.isSuspended = false
+            @unknown default:
+                self.syncQ.isSuspended = false
+            }
+        }
+    }
```
jacobsimeon, line 107: This bit here is really the crux of the whole thing. Since all of our operations are executed in order via `self.syncQ`, when the user goes offline we simply pause the queue. Local changes will take effect immediately while the remote operations wait in the queue.
When the user goes back online, we resume the queue and the operations are executed on the server in order.

jacobsimeon: I'm not sure whether we should pull down any changes _before_ executing the local changes or if that will be necessary?

dskuza: I think we can resort to a K.I.S.S approach and send our changes up _to_ the server, and let the next "sync" with data _from_ the server update the local entries as necessary. TL;DR - Local is the source-of-truth until we request that the server tell us otherwise.

dskuza (approved):

PR #130, opened 2022-02-18
Refactor session into single keychain-based struct

dskuza:
This pull request tacks onto the previous FxA work by saving the user's session after fetching a guid and logging in via the FxA web session. `Session` refers to the combination of guid, access token, and user identifier. `SessionController` now takes an `AppSession` object, which itself contains a `session`. This is done so we have an in-memory representation of the app's session state, which can be shared between login and settings (via `SessionController`), and other necessary objects, such as the Apollo client, which utilizes the session to fetch guid and access token. `AppSession` uses a new `KeychainStorage` property wrapper that encodes `session` and stores it in the Keychain, adding / updating / deleting as necessary.


jacobsimeon (approved):
LGTM (pending session caching via KeychainStorage)

PR #131, opened 2022-02-23
Implement full login flow via FxA

dskuza:
This pull request finalizes support for logging in via FxA. The major change here is the removal of `SessionController` and `RootViewModel`, and the increased usage of `AppSession`.

Rather than having view models send events upwards to the `RootCoordinator`, the `RootCoordinator` uses the state of the `AppSession` to determine which state the application should be in - `main`, or `loggedOut` - by subscribing to its change in `currentSession`. `AppSession` is passed into both the login flow and settings flow, which update the `currentSession` to either a new value, or nil, which in turn triggers the state change, as previously mentioned.

Since we do need to do some cleanup if the session changes, there is a new class that is introduced - `SessionListener`. This listener is initialized within `Services` with its `AppSession` "sibling", and is immediately put to work. This class subscribes to any changes to `currentSession`  and, for example, resets user defaults or tracker. With this change, we can also easily use `AppSession` when we pass in the `clearKeychain` argument on launch by `nil`ing out the session, or setting a new session with credentials supplied by tests.


jacobsimeon (changes requested):
as discussed, look into:
1. Using the factory pattern to remove transient dependencies from root coordinator
2. (re)introduce root view model to give business logic of logging in/out a home

dskuza:
@jacobsimeon Updated. Check out https://github.com/Pocket/pocket-ios/pull/131/commits/7039ea0b07d3ce347138f68d461e1fded36520a5 and https://github.com/Pocket/pocket-ios/pull/131/commits/5aca740f4c205cc0cd9bbde63c8c61cb132ec430 for the two changes you requested.

jacobsimeon (approved):

PR #136, opened 2022-02-25
Obtain session guid and user id from FxA redirect

dskuza:
The FxA redirect will soon be updated to add `guid` and `id` query parameters to the preexisting `access_token`. This lets us remove the solo request for the `guid` preceding login, simplifying the network requests greatly from a previously-required 3 (one each for: guid, fxa, user ID), to 1 (guid, fxa, and user ID in a single response).


jacobsimeon (approved):
💣 💥 BOOM

PR #140, opened 2022-03-02
Implement offline screen within authentication screen

dskuza:
This pull request implements an offline screen when a user requests "log in" or "sign up" while offline. This offline screen supports explicit dismissal (tapping "X" or swiping down), as well as automatic dismissal upon reconnection. If automatically dismissing, the appropriate auth flow will then appear based on the initiating button press.

## Offline Screen Presentation

Given the user is offline,
when the user opens the app,
they should see the standard authentication screen.

Given the user is offline and is on the authentication screen,
when the user taps sign up,
the offline splash screen should appear.

Given the user is offline and is on the authentication screen,
when the user taps log in,
the offline splash screen should appear.

Given the user is offline and the splash screen is visible,
when the user caps the X button,
the view should dismiss.

## Automatic Offline Screen Dismissal

Given the user is offline, tapped log in, and the offline splash screen is visible,
when the device reconnects to a network,
then the splash screen should disappear and the login flow should appear.

Given the user is offline, tapped signup and the offline splash screen is visible,
when the device reconnects to a network,
then the splash screen should disappear and the signup flow should appear.

## Analytics

- Engagement event when "Log in" is tapped
- Engagement event when "Sign up" is tapped

## Preview

image


PocketKit/Tests/PocketKitTests/PocketLoggedOutViewModelTests.swift
```diff
@@ -5,40 +5,46 @@ import AuthenticationServices
...
 
 extension PocketLoggedOutViewModelTests {
     func test_logIn_onFxAError_setsPresentedAlert() {
-        let urlSession = MockURLSession()
-        urlSession.stubData { request in
-            self.validGUIDResponse(for: request)
-        }
-
-        let client = AuthorizationClient(
+        let failingClient = AuthorizationClient(
```
jacobsimeon: Unfortunate that we need to re-create the `AuthorizationClient` here so that we can inject a different type of mock. It would be nice if we just had one `MockAuthenticationSession` that we can manipulate instead of having to rebuild the whole dependency graph.

I see why we need this: we only pass in the _type_ of auth session, not a concrete instance. The factory pattern can help us here. If we inject a function that creates the `AuthenticationSession` instead just a simple type.

I'd like to see this get done before merging the PR, but I could be convinced to leave it as-is for now if it's going to be a time sink.

PocketKit/Tests/PocketKitTests/PocketLoggedOutViewModelTests.swift
```diff
@@ -5,40 +5,46 @@ import AuthenticationServices
...
+        )
         appSession = AppSession(keychain: MockKeychain())
         subscriptions = []
     }
 
     override func tearDown() {
         subscriptions = []
     }
+
+    func subject(
```
jacobsimeon: ohh yeah 😎

I like this pattern

PocketKit/Sources/PocketKit/LoggedOut/PocketLoggedOutViewModel.swift
```diff
@@ -1,35 +1,60 @@
...
 
+    private var isOffline: Bool {
+        networkPathMonitor.currentNetworkPath.status == .unsatisfied
+    }
+
     @Published
     var presentedAlert: PocketAlert? = nil
 
+    @Published
+    var presentOfflineView: Bool = false
```
jacobsimeon: `isPresentingOfflineView` would be a more conventional name here

PocketKit/Tests/PocketKitTests/PocketLoggedOutViewModelTests.swift
```diff
@@ -106,6 +111,36 @@ extension PocketLoggedOutViewModelTests {
...
+        viewModel.$presentOfflineView.dropFirst().sink { present in
+            XCTAssertTrue(present)
+            offlineExpectation.fulfill()
+        }.store(in: &subscriptions)
+
+        viewModel.logIn()
+        wait(for: [offlineExpectation], timeout: 1)
+    }
+
+    func test_signUp_whenOffline_setsPresentOfflineView() {
```
jacobsimeon: So awesome to see these tests for the view model layer!

PocketKit/Sources/PocketKit/LoggedOut/LoggedOutViewModel.swift
```diff
@@ -36,6 +38,23 @@ class LoggedOutViewModel: ObservableObject {
...
+            }
+        }
+    }
+
+    private func updateStatus(_ status: NWPath.Status) {
+        guard status != currentNetworkStatus else {
+            return
+        }
+
+        if currentNetworkStatus == .unsatisfied, status == .satisfied, presentOfflineView == true {
```
jacobsimeon, line 65: Curious if we actually need all these checks here?

If `presentOfflineView` is already false, would setting it to false again have any effect?

jacobsimeon (commented):
So this all looks great, just a few suggestions throughout.

The one question I have is around the behavior of automatically showing the previously requested `LoggedOutAction`. Is this really something we need to worry about? I worry that keeping track of all the state will get hairy and lead to bugs. Is there a way that we could send a message back to the main logged in view(model) when the offline view is automatically dismissed instead of keeping track of the state?

dskuza:
@jacobsimeon 

> The one question I have is around the behavior of automatically showing the previously requested LoggedOutAction. Is this really something we need to worry about?

This was something that was part of the acceptance criteria for the legacy app, so it was moved over into iOS Next as well. Happy to discuss whether we _should_ do it, but it's a byproduct of "cloning" the legacy app functionality.

PocketKit/Tests/PocketKitTests/PocketLoggedOutViewModelTests.swift
dskuza: @jacobsimeon I'd like to discuss this one live, if possible.

jacobsimeon (approved):
Nice!

PR #143, opened 2022-03-09
Force web view when item is not an article, is an image, or is a video

dskuza:
TL;DR - If an article is an article (that may or may not contain images or videos), then present it in the reader. For all other cases, present the web view.

## Cases

1. `isArticle` == false (source of truth: parser), then present the web view upon selection
2. `hasImage` == `IS_IMAGE`, then present the web view upon selection
3. `hasVideo` == `IS_VIDEO`, then present the web view upon selection
4. Present the reader as normal

### Rationale 
The rationale behind points 1 and 2 above is that the "standard" web view will present more options than we currently do. For example, a YouTube video will present the entire YouTube page, which provides more valuable information than just the video itself (especially since we do not provide an offline option for the video). Another example is an image - web views handle image interaction (tap & hold, drag gestures) more than we currently do now as well.

### `Item`

+ `isArticle` (Core Data)
+ `imageness` (Core Data)
+ `videoness` (Core Data)
+ `hasImage` (to match the backend property; computed based on `imageness`)
+ `hasVideo` (to match the backend property; computed based on `videoness`)

### `UnmanagedItem`

+ `isArticle` (in-memory)
+ `imageness` (in-memory)
+ `videoness` (in-memory)
+ `hasImage` (to match the backend property; computed based on `imageness`)
+ `hasVideo` (to match the backend property; computed based on `videoness`)

### `ItemParts` (`list.graphql`)
+ `isArticle`
+ `hasImage`
+ `hasVideo`

### UI Tests

UI tests were added across the UI test suite to verify that under three conditions (1-3 above), the web view is presented on tap.


jacobsimeon (approved):
All the `imageness` and `videoness` nomenclature from the backend is a bit...different. But this looks great!

At some point we'll want to download/store images locally, so we might want to change the policy for images when time comes. That way users can still view images while offline.

PR #144, opened 2022-03-10
Periodically refresh My List while app is in background

jacobsimeon:
https://getpocket.atlassian.net/browse/IN-279


dskuza (changes requested):
Let's add an `expirationHandler` to the task, which should set the task as complete (even if the refresh didn't finish).

dskuza (approved):

PR #146, opened 2022-03-16
Create initial Share Extension target

dskuza:
Code signing is not yet set up as we do not yet have the correct provisioning profile(s) generated.

The activation rule is carried over from the legacy app.


Pocket.xcodeproj/project.pbxproj
```diff
@@ -73,9 +75,27 @@
 			remoteGlobalIDString = 166A81AF2637406C0015AA1D;
 			remoteInfo = "Pocket (iOS)";
 		};
+		F204A76827E22EC50010E155 /* PBXContainerItemProxy */ = {
+			isa = PBXContainerItemProxy;
+			containerPortal = 166A81A42637406B0015AA1D /* Project object */;
+			proxyType = 1;
+			remoteGlobalIDString = F204A75F27E22EC50010E155;
+			remoteInfo = "Share Extension";
```
jacobsimeon, line 83: Can we please verify that this is correct and not just a remnant of the original name?

PR #151, opened 2022-03-26
Save extension service

jacobsimeon:
This is the same code that we wrote together with the addition of storing the `SavedItem` locally when saving the item and updating it via the graphQL response.

Also, I noticed a layout bug when testing out the extension. It looked wonky on my tiny iPhone mini ;). Added a few adjustments to the layout code but could probably be improved.

Screenshot of layout bug:
IMG_1822


jacobsimeon:
I don't think we need to do it as part of this PR, but we should rethink where some of this code lives to avoid having to make so many things `public`

dskuza (approved):

PR #155, opened 2022-03-29
SaveToPocket: Support automatic dismissal

dskuza:
This pull request introduces automatic dismissal of the `SaveToPocket` extension after `n` number of seconds. For the app implementation, this interval is 2 seconds, which is consistent with the legacy app. For tests, this interval is 0 seconds.

There are primary cases at this time:
1. The user is not logged in
2. The user is logged in, but there is no URL to share
3. The user is logged in, and there is a URL to share

All three of these cases will result in the view being automatically dismissed.


jacobsimeon (approved):
Sweet! Nice looking tests

PR #159, opened 2022-04-04
Fix display settings not dismissable in landscape

dskuza:
There are a few extra properties we can set on `UISheetPresentationController` to ensure that the view controller is dismissable in landscape. One oddity here is that we have to set both `.medium()` and `.large()` detents to support autorotate, as if we only set `.medium()`, then the landscape presentation using the new properties causes an exception, where having both detents is the workaround 🙃 

This does not change the iPad implementation, which uses a menu.

## Preview

image


jacobsimeon (approved):
approved (pending deletion of commented out code)

PR #161, opened 2022-04-05
SaveTo: Add "Log In to Pocket" button when logged out

dskuza:
This pull request adds support for presenting a "Log in to Pocket" button when opening the extension if the user is logged out. 

# TL;DR

## MainViewModel

This view model has been updated to return a presenter (type: `MainViewPresenter` based on state and origin. This presenter is then used to conditionally show (or not show) an additional view at the bottom of `MainViewController` based on the usability of the presenter. In a logged out state, this is based on whether there is a `UIResponder` in the chain that can handle opening URLs. 

## MainViewPresenter

This type has been added as a protocol with one required property: `view: UIView?`. This view is what is conditionally shown at the bottom of `MainViewController`. Currently, there are two presenters: `SavedViewPresenter` (for when an item has been saved) and `LoggedOutViewPresenter` (for when the user is logged out). These presenters are also used to update the state and text of the `MainInfoView` presented by the main view controller.

## URL Schemes

The `pocket-next` URL scheme has been registered with the main app so that it can be opened from the extension (by opening `pocket-next:`. 

## Preview

image

### DIFF
--- a/Info.plist
+++ b/Info.plist
@@ -22,6 +22,19 @@
 	<string>$(PRODUCT_BUNDLE_PACKAGE_TYPE)</string>
 	<key>CFBundleShortVersionString</key>
 	<string>1.0</string>
+	<key>CFBundleURLTypes</key>
+	<array>
+		<dict>
+			<key>CFBundleTypeRole</key>
+			<string>Editor</string>
+			<key>CFBundleURLName</key>
+			<string>com.mozilla.pocket.next</string>
+			<key>CFBundleURLSchemes</key>
+			<array>
+				<string>pocket-next</string>
+			</array>
+		</dict>
+	</array>
 	<key>CFBundleVersion</key>
 	<string>1</string>
 	<key>ITSAppUsesNonExemptEncryption</key>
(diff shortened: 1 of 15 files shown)


jacobsimeon (commented):
So glad to see these UI tests exercising the extension!

I like the use of presenter here but I wonder if we're trying to fit too much into a single view model. I think we could actually have two separate view models, one for each state (logged in, logged out). At initialization, we can decide which view model to use. 

If `MainViewController`'s `viewModel` property were actually a protocol `MainViewModel` then we could have two implementations of that protocol: `LoggedInMainViewModel` and `LoggedOutMainViewModel`.

Then, each view model can be responsible for telling the view controller what buttons should be present.

```swift
// in viewController's init
// decide which view model to use
if appSession.currentSession != nil {
  self.viewModel = LoggedInMainViewModel()
} else {
  self.viewModel = LoggedOutMainViewModel()
}

// we can subscribe to a `buttons` publisher on the view model
viewModel.actionButtons.sink { buttons in 
  // buttons here can be [UIButton]
  // or perhaps an a array of `ActionButton` ( a custom struct)
  // that contains all the data necessary to construct an actual button
  // assuming our custom view has a setter that can handle insertion/layout of the action buttons
  view.actionButtons = button
}.store(in: subscriptions)

// in viewController.viewDidAppear
// allow the viewModel to walk the responder chain
// it can then update it's internal `buttons` property and the view controller will respond
viewModel.viewDidAppear(self)

// The implementation of `viewDidAppear` in the "logged in" implementation of the view model would be a no-op
```

This would be a pattern similar to how `ItemsListViewController` can use either `ArchivedItemsListViewModel` or `SavedItemsListViewModel`. 

Using the `MainViewPresenter` protocol was the right idea for sure. I just want to take that idea and elevate it up another level. I think it would clarify the code here and keep from having to do things like check the type of a presenter (as in isPresenterUsable).

PocketKit/Tests/SaveToPocketKitTests/MainViewModelTests.swift
```diff
@@ -1,171 +1,173 @@
...
-        context.stubCompleteRequest { _, _ in
-            completeRequestExpectation.fulfill()
-        }
-
-        await viewModel.save(from: context)
-
-        wait(for: [completeRequestExpectation], timeout: 1)
-    }
-}
+//import XCTest
```
jacobsimeon, line 1: Does this file need to be deleted now that we have `SavedItemViewModelTests`?

jacobsimeon (approved):
Nice! I really like where we landed here.

I'd recommend squashing all the commits since most of them are just code churn that would be nice to keep out of our main branch. I'll leave that decision up to you though

dskuza:
Squashed all commits :+1:

PR #163, opened 2022-04-06
Fix list switcher layout issue after toggling lists

dskuza:
Fixes an issue where the navigation bar state could be out-of-sync with the expected state based on the current visible collection view's content offset after toggling list selection.

## Example Issue

image


jacobsimeon (approved):
cool cool

PR #164, opened 2022-04-06
ShareTo: Support saving items while offline

jacobsimeon:
# Source
- Listen for "unresolvedSavedItemCreated" notification
  When this notification is recieved, we enqueue a `saveItem` sync task
  that will be executed the next time a network connection becomes
  available.
- Add `resolveUnresolvedSavedItems` method that can be called from
  higher level code (such as PocketKit.RefreshCoordinator when the app
  enters foreground)

# PocketSaveService
- Send `.unresolvedSavedItemCreated` notification when:
  - the expiring activity expires
  - the request to client-api fails for any reason (such as a network timeout
    or airplane mode being enabled)

# RefreshCoordinator
- tell `source` to `resolveUnresolvedSavedItems` when app enters
  foreground

### DIFF
--- a/PocketKit/Tests/SyncTests/PocketSourceTests+savedItemEvents.swift
+++ b/PocketKit/Tests/SyncTests/PocketSourceTests+savedItemEvents.swift
@@ -46,4 +46,26 @@ extension PocketSourceTests {
         XCTAssertEqual(notifications, [])
         XCTAssertFalse(space.context.hasChanges)
     }
+
+    func test_events_whenOSNotificationCenterPostsUnresolvedItemCreatedNotification_enqueuesASaveItemOperation() throws {
+        let operationStarted = expectation(description: "operationStarted")
+        operations.stubSaveItemOperation { _, _, _, _, _ in
+            return TestSyncOperation {
+                operationStarted.fulfill()
+            }
+        }
+
+        let source = subject()
+
+        let savedItem = try! space.seedSavedItem()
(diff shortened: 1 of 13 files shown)


PocketKit/Tests/SyncTests/PocketSourceTests+savedItemEvents.swift
```diff
@@ -46,4 +46,26 @@ extension PocketSourceTests {
...
+        let savedItem = try! space.seedSavedItem()
+        let unresolved: UnresolvedSavedItem = space.new()
+        unresolved.savedItem = savedItem
+        try space.save()
+
+        osNotificationCenter.post(name: .unresolvedSavedItemCreated)
+
+        wait(for: [operationStarted], timeout: 1)
+
+        try XCTAssertEqual(space.fetchUnresolvedSavedItems(), [])
```
dskuza, line 69: I feel as if the test name (`enqueuesASaveItemOperation`) implies more than just checking if we have no unresolved saved items. Can we dig into the persisted tasks to make sure we have a `saveItem` task ready?

jacobsimeon: I see your point but line 67 above is actually the "assertion" that a saveItem task is created (and started). The operationFactory is stubbed out to provide a `TestSyncOperation` that fulfills the expectation when the operation is started.

dskuza (approved):

PR #165, opened 2022-04-07
Add support for "Move to My List" via swipe in Archive

dskuza:
This pull request adds support for the "Move to My List" action via swipe when viewing one's archive.

## Details

- `trailingSwipeActions(for objectID: ItemIdentifier)` now returns `[ItemAction]?` instead of `UIContextualAction`, following a similar pattern to overflow actions.
- `Re-add` has been renamed to `Move to My List`, and all occurrences have been replaced.
- Any APIs that utilize a `Move to My List` action have been renamed accordingly for consistency.
- `(Pocket)Source` APIs have not been updated, and maintain `unarchive` as a naming convention.
- UI tests have been added for both swiping to archive from "My List", and swiping to "Move to My List" when viewing "Archive"

## Preview

Simulator Screen Shot - iPhone 13 Pro - 2022-04-07 at 10 51 46

### DIFF
--- a/PocketKit/Sources/PocketKit/MyList/SavedItemsList/SavedItemsListViewModel.swift
+++ b/PocketKit/Sources/PocketKit/MyList/SavedItemsList/SavedItemsListViewModel.swift
@@ -135,9 +135,9 @@ class SavedItemsListViewModel: NSObject, ItemsListViewModel {
         sharedActivity = PocketItemActivity(url: item.url, sender: sender)
     }
 
-    func overflowActions(for objectID: NSManagedObjectID) -> [ItemAction]? {
+    func overflowActions(for objectID: NSManagedObjectID) -> [ItemAction] {
         guard let item = bareItem(with: objectID) else {
-            return nil
+            return []
         }
 
         return [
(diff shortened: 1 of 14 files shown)


dskuza:
@jacobsimeon once this is approved, I plan on squashing the commits.

jacobsimeon (changes requested):
As discussed in slack, let's create `ItemContextualAction` for the purposes of driving swipe actions from the view model (as opposed to re-using `ItemAction` which has a slightly different use case)

PocketKit/Sources/PocketKit/MyList/SavedItemsList/SavedItemsListViewModel.swift
```diff
@@ -150,18 +150,17 @@ class SavedItemsListViewModel: NSObject, ItemsListViewModel {
         ]
     }
 
-    func trailingSwipeActions(for objectID: NSManagedObjectID) -> [UIContextualAction] {
+    func trailingSwipeActions(for objectID: NSManagedObjectID) -> [ItemContextualAction]? {
```
jacobsimeon, line 153: I dig that we're moving away from `UIContextualAction` here. You can adjust return type to non-optional array if you are so inclined ;)

PR #171, opened 2022-04-20
Fixup bundle versioning

jacobsimeon:
Xcode seems to want us to use $(MARKETING_VERSION) and
$(CURRENT_PROJECT_VERSION) build settings instead of explicitly setting
the corresponding values in SaveToPocket/Info.plist.

In order to keep the versions in sync between our app and the extension,
we adjust the main app's Info.plist to use those build settings as
well.

This also requires a change on Bitrise to override
$(CURRENT_PROJECT_VERSION) when creating an archive for submitting to
the app store. (I've already made this change)


cyndichin (approved):
Looks good! Builds successfully on device and simulator.

dskuza (approved):

PR #172, opened 2022-04-22
Omit gray image background when caption is not present

cyndichin:
# IN-559 
**Summary**
In our article reader view, we have images that have a grey background if the image does not span to the entire width of the screen. We only want the grey background to appear if there is an image with a small image size and has an associated caption. If there is no caption, then we want to set the background of the image view to `.clear` instead of grey.

**Testing Steps**
1. Navigate to Home tab on an iPad and tap on the following article: [Night Shifts, by Michael W. Clune](https://harpers.org/archive/2022/04/night-shifts-dream-incubation-technology-sleep-research/)
2. Scroll through the article and verify that images have a clear background
3. Navigate to this following article: [How to Train Your Brain to Get What You Want in 60 Days] (https://getpocket.com/explore/item/how-to-train-your-brain-to-get-what-you-want-in-60-days)
4. Scroll through the article and verify that images have a grey background with captions

**Screenshot**
Has no caption
image

Has caption
image

### DIFF
--- a/PocketKit/Sources/PocketKit/Article/Cells/ImageComponentCell.swift
+++ b/PocketKit/Sources/PocketKit/Article/Cells/ImageComponentCell.swift
@@ -1,6 +1,12 @@
 import UIKit
 import Kingfisher
 
+protocol ImageComponentCellModel {
+    var caption: NSAttributedString? { get }
+    var image: ImageComponentCell.ImageSpec? { get }
+    var shouldHideCaption: Bool { get }
+    func imageViewBackgroundColor(imageSize: CGSize) -> UIColor
+}
 
 class ImageComponentCell: UICollectionViewCell {
     enum Constants {
@@ -12,12 +18,7 @@ class ImageComponentCell: UICollectionViewCell {
         let source: URL
         let size: CGSize
     }
-
-    struct Model {
-        let caption: NSAttributedString?
-        let image: ImageSpec?
-    }
-
+    
     let imageView: UIImageView = {
         let imageView = UIImageView()
         imageView.contentMode = .center
@@ -57,9 +58,9 @@ class ImageComponentCell: UICollectionViewCell {
 }
 
 extension ImageComponentCell {
-    func configure(model: Model, imageLoaded: ((UIImage) -> Void)? = nil) {
+    func configure(model: ImageComponentCellModel, imageLoaded: ((UIImage) -> Void)? = nil) {
         captionTextView.attributedText = model.caption
-        captionTextView.isHidden = model.caption == nil || model.caption?.string.isEmpty == true
+        captionTextView.isHidden = model.shouldHideCaption
 
         imageView.image = nil
         guard let imageSpec = model.image else {
@@ -81,14 +82,7 @@ extension ImageComponentCell {
         ) { [weak self] result in
             switch result {
             case .success(let result):
-
-                if let idealWidth = model.image?.size.width,
-                   result.image.size.width >= idealWidth {
-                    self?.imageView.backgroundColor = UIColor(.clear)
-                } else {
-                    self?.imageView.backgroundColor = UIColor(.ui.grey7)
-                }
-
+                self?.imageView.backgroundColor = model.imageViewBackgroundColor(imageSize: result.image.size)
                 imageLoaded?(result.image)
             case .failure:
                 break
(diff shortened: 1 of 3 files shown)


cyndichin:
not sure if scope with this ticket and may be too small, but in the future, do we think this logic (`model.caption == nil || model.caption?.string.isEmpty == true)` should fit in a view model so we can just utilize `viewModel.hasCaption` and makes this more testable?

PocketKit/Sources/PocketKit/Article/Cells/ImageComponentCell.swift
```diff
@@ -86,7 +86,8 @@ extension ImageComponentCell {
                    result.image.size.width >= idealWidth {
                     self?.imageView.backgroundColor = UIColor(.clear)
                 } else {
-                    self?.imageView.backgroundColor = UIColor(.ui.grey7)
+                    let isCaptionHidden = self?.captionTextView.isHidden ?? false
```
jacobsimeon, line 89: This works, but I actually really like your idea of putting this in the view model.
How about we add `let imageViewBackgroundColor` (or some better name) to the view model. That way we can just say `imageView.backgroundColor = model.imageViewBackgroundColor` and put the logic in the view model?

cyndichin: sounds good, will update! just to be clear, did you meant `viewModel.imageViewBackgroundColor` ?

jacobsimeon: yes `viewModel` not `model` :)

jacobsimeon (changes requested):
Love the usage of tests for the view model!

Just added a couple of notes inline

PocketKit/Sources/PocketKit/Article/Cells/ImageComponentCell.swift
```diff
@@ -17,6 +17,18 @@ class ImageComponentCell: UICollectionViewCell {
         let caption: NSAttributedString?
         let image: ImageSpec?
     }
+    
+    struct ViewModel {
```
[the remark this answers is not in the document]
cyndichin: updated with the protocol!

PocketKit/Sources/PocketKit/Article/Presenters/ImageComponentPresenter.swift
```diff
@@ -68,6 +68,31 @@ class ImageComponentPresenter: ArticleComponentPresenter {
...
+            )
+        }
+    }
+    
+    var shouldHideCaption: Bool {
+        return caption == nil || caption?.string.trimmingCharacters(in: .whitespaces).isEmpty == true
+    }
+    
+    func imageViewBackgroundColor(imageSize: CGSize) -> UIColor {
+        if let idealWidth = image?.size.width,
```
jacobsimeon, line 88: I'm wondering what happens here if `image?.size.width` happens to be `nil` but `shouldHideCaption` is still true. We'd still want a `.clear` background color but I think in this (very edgey) case, we'd end up with a grey background.

PocketKit/Tests/PocketKitTests/ImageComponentPresenterTests.swift
```diff
@@ -0,0 +1,98 @@
...
+            height: 1,
+            width: 2,
+            id: 3,
+            source: URL(string: "http://example.com")!
+        )
+        let presenter = ImageComponentPresenter(component: component, readerSettings: ReaderSettings()) { }
+        let imageSize = CGSize(width: -1, height: 0)
+        
+        XCTAssertEqual(presenter.imageViewBackgroundColor(imageSize: imageSize), UIColor(.clear))
+    }
```
jacobsimeon, line 97: In light of my above comment, try initializing `ImageComponent` with `caption: nil, credit: nil, source: nil` and make sure we still get a clear background color.

cyndichin: updated!

PocketKit/Sources/PocketKit/Article/Presenters/ImageComponentPresenter.swift
cyndichin: updated so that if no image, we return `.clear` background color

jacobsimeon (approved):
Nice work! Thanks for all your patience in iterating on this idea. I think we ended up with an improved design in addition to a nice new bit of functionality. Well done!

audie1:
@cyndichin thank you for the clear test steps! Apologies for taking so long getting the build process figured out here, great work. I verified your test steps and checked images are looking good in general for other articles, no issues found.

PR #174, opened 2022-04-27
IN-591: Constrain Logout Screen to Portrait View

cyndichin:
# IN-591

**Issue:** Progress indicator for the Login slideshow onboarding overlaps the readable text in Landscape view for iPhone
**Fix:** Disabled landscape view for iPhone and leave iPad view as is

### DIFF
--- a/PocketKit/Sources/PocketKit/LoggedOut/LoggedOutViewController.swift
+++ b/PocketKit/Sources/PocketKit/LoggedOut/LoggedOutViewController.swift
@@ -7,6 +7,11 @@ class LoggedOutViewController: UIHostingController<LoggedOutView> {
     convenience init(viewModel: LoggedOutViewModel) {
         self.init(rootView: LoggedOutView(viewModel: viewModel))
     }
+    
+    override var supportedInterfaceOrientations: UIInterfaceOrientationMask {
+        guard traitCollection.userInterfaceIdiom == .phone else { return .all }
+        return .portrait
+    }
 }
 
 struct LoggedOutView: View {


PocketKit/Sources/PocketKit/LoggedOut/LoggedOutViewController.swift
```diff
@@ -7,6 +7,11 @@ class LoggedOutViewController: UIHostingController<LoggedOutView> {
     convenience init(viewModel: LoggedOutViewModel) {
         self.init(rootView: LoggedOutView(viewModel: viewModel))
     }
+    
+    override var supportedInterfaceOrientations: UIInterfaceOrientationMask {
+        guard UIDevice.current.userInterfaceIdiom == .phone else { return .all }
```
dskuza, line 12: 🤔 Would using the view controller's `traitCollection.userInterfaceIdiom` work here as well? I feel it may be better to check the idiom within the context of the current view controller over the device (we also won't have an implicit dependency on `UIDevice`).

dskuza (commented):
@cyndichin just one quick question for ya

PocketKit/Sources/PocketKit/LoggedOut/LoggedOutViewController.swift
cyndichin: good point on the dependency, thanks!

dskuza (approved):

audie1:
Good work @cyndichin - the iPhone is now locked into Portrait view, as expected, I also checked iPad for regressions and found none, no issues found during testing.

PR #180, opened 2022-05-06
IN-569: Fix List Item Action Buttons Sizing

cyndichin:
# IN-569
**Issue:** My List/Archive item action buttons sizing does not match designs. This is due to the image being set as the background image and not the foreground image of the button. Also, the image were not conforming to 20x20 as specified in the Figma file.
**Fix:** Add resizable function to adjust image size and set images to the foreground instead of background of the button.

Reference: https://nshipster.com/image-resizing/

## Testing Steps
1. Navigate to MyList
2. Verify action button items (favorite, share, more) resemble the sizes in Figma file for iPhone and iPad. 

## Screenshots
Simulator Screen Shot - iPhone 13 Pro Max - 2022-05-06 at 09 45 27

### DIFF
--- a/PocketKit/Sources/PocketKit/MyList/ItemsList/ItemsListItemCell.swift
+++ b/PocketKit/Sources/PocketKit/MyList/ItemsList/ItemsListItemCell.swift
@@ -31,6 +31,7 @@ class ItemsListItemCell: UICollectionViewCell {
         static let textStackSpacing: CGFloat = 8
         static let topLevelStackSpacing: CGFloat = 14
         static let actionButtonHeight: CGFloat = 28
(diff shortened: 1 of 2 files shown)


jacobsimeon (approved):
Nice! 

Ideally, we wouldn't have to manually resize the images and we could just set the `contentInsets` for the button configuration. This would be analogous to setting the `contentEdgeInsets` on the button itself (in the old days, before the button configuration api). 
But, for whatever reason, button's don't scale their images to fit within the content insets when using `UIButtonConfiguration`. 🫤

All this to say: looks good to me!

audie1:
Button sizing looks good on phone/tablet. I did some basic device rotations and navigation, I'll do some stress testing at a later time, for now these changes are looking good to go 🟢

PR #186, opened 2022-05-24
Home: implement actions for recommendations in reader

dskuza:
This pull request adds support for actions for recommendations opened in the reader.

Recommendations have three states:
1. Unsaved (not in "My List" or "Archive")
2. Saved, but not archived
3. Saved, but archived

Each of these states have their own actions available to them (as reviewed in Slack with product).

Points 1. and 2. above share the same available actions:
1. Display Settings
2. Favorite / Unfavorite
3. Move to My List / Archive
4. Delete
5. Share

Point 3 has a much more simplified set of actions:
1. Display Settings
2. Save
3. Share

For point 3, upon saving, the available actions will update to the actions shared by points 1. and 2 (acting as a "saved item"). When deleting an item via action, the available actions will be updated to the actions for point 3.

A change to `PocketSource.delete(item:)` had to be made. First, the data model was updated to `nullify` the `item` relationship on delete. Second, the `SavedItem.item` is deleted _if not part of a recommendation_. Otherwise, it remains stored.

### DIFF
--- a/PocketKit/Sources/PocketKit/Article/ReadableViewModel/RecommendationViewModel.swift
+++ b/PocketKit/Sources/PocketKit/Article/ReadableViewModel/RecommendationViewModel.swift
@@ -37,12 +37,13 @@ class RecommendationViewModel: ReadableViewModel {
         self.source = source
         self.tracker = tracker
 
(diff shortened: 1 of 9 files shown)


jacobsimeon (approved):
Very nice

PR #188, opened 2022-05-26
Add authentication safeguards against multiple attempts

dskuza:
## Summary

This pull request adds authentication safeguards to ensure there is only one active authentication session while the user is logged out

## References

IN-608

## Details

- Ensure the AuthorizationClient can only start one session at a time
- No-op LoggedOutViewModel.logIn and LoggedOutViewModel.signUp if the user is already logged in

## Testing

Unit tests have been added, and self-QA was performed to ensure on-device authentication still worked as intended.

### DIFF
--- a/PocketKit/Sources/PocketKit/Authorization/AuthorizationClient.swift
+++ b/PocketKit/Sources/PocketKit/Authorization/AuthorizationClient.swift
@@ -9,6 +9,8 @@ import AuthenticationServices
 public class AuthorizationClient {
     typealias AuthenticationSessionFactory = (URL, String?, @escaping ASWebAuthenticationSession.CompletionHandler) -> AuthenticationSession
 
+    private var isAuthenticating = false
+
     private let consumerKey: String
     private let authenticationSessionFactory: AuthenticationSessionFactory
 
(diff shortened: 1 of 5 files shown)


PocketKit/Sources/PocketKit/Authorization/AuthorizationClient.swift
```diff
@@ -22,11 +24,25 @@ public class AuthorizationClient {
 
     @MainActor
     func logIn(from contextProvider: ASWebAuthenticationPresentationContextProviding?) async throws -> Response {
+        defer { isAuthenticating = false }
+
+        guard isAuthenticating == false else {
+            throw AuthorizationClient.Error.alreadyAuthenticating
+        }
+
+        isAuthenticating = true
```
cyndichin, line 33: looks good, just curious - any reason why we defer first here and defer after the guard in `signUp`

cyndichin (approved):

PocketKit/Sources/PocketKit/Authorization/AuthorizationClient.swift
dskuza: Just preference - to me, it just makes it more obvious that it's happening. Though, in theory, this `defer` does nothing if the guard fails since the value will already be `false`.

PR #189, opened 2022-06-02
Add image prefetching and cache deletion for Image objects

dskuza:
## Summary

This pull request adds prefetching of all `Image`s that are added to Core Data, typically when an `Item` is updated when sync'd. This prefetching begins at startup, and continues fetching images as `Image` objects are loaded into Core Data. Alternatively, if an `Image` is deleted, any related images are deleted from the image cache.

## References

IN-596
IN-612

## Implementation Details

`ImageManager` is the main object of this pull request. It listens and responds to Core Data changes (via callback from an `ImagesController` instance, which uses `NSFetchedResultsController` under the hood). After that, if an `Image` is inserted or updated, we will fetch the image (or not, if already cached). Alternatively, if an `Image` is deleted, we will use the image cache to remove the image from cache (using `ImageRetriever.imageCache`).

When fetching images, we use `imageCacheURL` to ensure we are hitting the Pocket image cache for the image. 

Checking the cache and downloading (as necessary) is handled by `KingfisherManager.shared`, which is passed in as the `imageRetriever` to `ImageManager`; it has been extended to conform to the `ImageRetriever` protocol. There, we use `retrieveImage`, with no completion handler. This will download the image as necessary and cache it (or not, if already in the cache). `KingfisherManager` uses both an `ImageCache` and `ImageDownloader` under the hood to download as necessary. One question I was asking myself during development was whether we should use these independent types to check the cache and download / store accordingly, or just use `KingfisherManager`, which chains these operations as we would need, anyways. I chose the latter because there is a cache hit anyways if we use `ImageCache.isCached`; the result of not using a completion handler to use a retrieved image equates to a cache hit to me.

Another question we may find ourselves asking is whether or not `KingfisherManager.shared` is the right instance to use. In this case, it is appropriate, because we use `.kf` extensions on `UIImageView` (e.g `imageView.kf.setImage`), which implicitly use `KingfisherManager.shared` under the hood, so we know we are sharing both the same cache and downloader across all of our uses of Kingfisher.

The final place where the cache should be cleared is upon sign out, but that is already handled in `RootViewModel` by calling `Textiles.clearImageCache` when a user signs out.


cyndichin (approved):
Looks good :) thanks for the live review

cyndichin:
> @cyndichin re-requesting a review from you because of two new commits:
> 
> [Manage and prefetch only undownloaded images](https://github.com/Pocket/pocket-ios/pull/189/commits/410d30f483f37d722b48fdaf6d75d6a8a6b79db4)
> 
> [Begin prefetching undownload images on launch](https://github.com/Pocket/pocket-ios/pull/189/commits/82fd483d8f06c316b27bfc0a725a1557c50bb539)

thanks! new changes looks good to me and images are loading in app ✅

cyndichin (approved):

PR #193, opened 2022-06-22
Support reporting a recommendation from the reader

dskuza:
## Summary

This pull request adds support for reporting an (unsaved) item from the Home feed, i.e a recommendation, from the reader when opening the action menu.

## References

IN-432

## Implementation Details

A new `ItemAction` has been added to `RecommendationViewModel`, which is the `ReadableViewModel` generated when an item from the Home feed is opened. This action has only been added when an item is unsaved (i.e not in "My List" or "Archive"). This introduces a new `selectedRecommendationToReport` property that mimics that of `HomeViewModel`. This property is them sinked to within both regular and compact coordinators, at which point the report modal is presented for the recommendation. Additionally, the `report` function in the coordinators has been refactored to accept a closure that will be called on dismissal of the report view. This cleans up the way in which the `selectedRecommendationToReport` properties are unset upon dismissal.

## Screenshots

image

## Test Steps

- [ ] Given I am on the Home tab and I have not previously saved or archived an item, when I open a recommendation in the reader and I tap the overflow menu, then I should see the report button.
- [ ] Given I am on the Home tab and I have previously saved or archived an item, when I open a recommendation in the reader and I tap the overflow menu, then I should not see the report button.
- [ ] Given I am on the Home tab and I have previously saved or archived an item, when I open a recommendation in the reader and I tap the overflow menu and delete the item and reopen the menu, then I should see the report button.

### DIFF
--- a/PocketKit/Sources/PocketKit/Article/ReadableViewModel/RecommendationViewModel.swift
+++ b/PocketKit/Sources/PocketKit/Article/ReadableViewModel/RecommendationViewModel.swift
@@ -25,6 +25,9 @@ class RecommendationViewModel: ReadableViewModel {
 
     @Published
     var isPresentingReaderSettings: Bool?
+
+    @Published
+    var selectedRecommendationToReport: Recommendation?
     
     private let recommendation: Recommendation
     private let source: Source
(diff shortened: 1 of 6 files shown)


cyndichin (approved):
Checked all cases - report button looks good!

audie1:
Tested across iPad/iPhone no issues found, QA Pass looking good!

PR #197, opened 2022-07-06
IN-513: Add Recent Saves Section

cyndichin:
## Summary
This PR adds the Recent Saves Section to the Home Tab and the user can view up to 5 of their items from MyList.

## References 
IN-513

## Implementation Details
* Utilize existing `ItemsListItemCell` as the cell design for Recent Saves Section
* Added `RecentSavesController` to manage fetching the appropriate number of objects from Core Data
* Added Archive, Delete and Share to Overflow Actions

## Test Steps
Scenario: User has at least one saved item
- [x] Given I have saved at least one item, and the item is not archived, When I navigate to the Home tab, Then I should see a section at the top that shows a horizontally scrolling list of recently saved items (up to 5)
- [x] Given that the item has overflow actions, the actions to archive, delete and share should appear. 

Scenario: User has no saved items
- [x] Given I do not have any items in my list of unarchived items, When I navigate to the Home tab, Then I should not see a the “recent saves” section

Note: Re-use as much existing design as possible (matching Figma File will be addressed in the future)
Note: There is a bug that was discovered while testing and it is a production bug that will be addressed in this ticket:
https://getpocket.atlassian.net/browse/IN-693

This bug can be triggered when saving the first recommendation item and perform a pull-to-refresh.

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

## Screenshots
image
Simulator Screen Shot - iPad Pro (11-inch) (3rd generation) - 2022-07-07 at 11 09 30


PocketKit/Sources/PocketKit/Home/HomeViewController.swift
```diff
@@ -170,6 +173,28 @@ extension HomeViewController {
...
+            let cell: ItemsListItemCell = collectionView.dequeueCell(for: indexPath)
+            
+            guard let presenter = model.presenter(for: objectID) else {
+                return cell
+            }
+
+            cell.backgroundConfiguration = .listPlainCell()
+            cell.layer.borderWidth = 0.5
+            cell.layer.borderColor = UIColor(.ui.grey6).cgColor
+            cell.layer.cornerRadius = 16
```
jacobsimeon, line 187: What do you think about putting this bit of styling code in the view itself? Then we could add a `style` property to the view model and set it to `.bordered` (or some such name) for this use case.

Something along the lines of 

```swift
extension ItemsListItemCell {
  enum CellStyle {
    case bordered
    case plain
  }
  struct Model {
    // ... omitting existing properties
    let style: CellStyle // default this to .plain
  }
}
```

jacobsimeon (approved):
This looks really good! I'm glad we got a chance to pair on it as well. Just had one thought for ya to consider

PocketKit/Sources/Sync/RecentSavesController.swift
```diff
@@ -0,0 +1,50 @@
...
+        super.init()
+        
+        savedItemsController.delegate = self
+    }
+}
+
+extension RecentSavesController: NSFetchedResultsControllerDelegate {
+    public func controllerDidChangeContent(_ controller: NSFetchedResultsController<NSFetchRequestResult>) {
+
+        if let _recentSaves = savedItemsController.fetchedObjects.flatMap({ Array($0.prefix(5)) }),
```
dskuza, line 29:
```
        if let _recentSaves = savedItemsController.fetchedObjects.flatMap({ Array($0.prefix(5)) }),
```

^ this logic is reused, so it could be extracted into a computed property, for example, but a non-blocker.

dskuza (approved):

audie1:
Looking good, one issue found: https://pocket.slack.com/archives/C01UH57EJKD/p1657312098192519

Overall I like the new feature and the function when the user removes all items from that list, how the recs move to the top. Nice work, QA Pass with the above as a known issue.

PR #199, opened 2022-07-07
Fix crash when reloading Home tab with items to be deleted

dskuza:
## Summary

Fix crash when reloading Home tab with items to be deleted due to SlateService.fetchSlateLineup(_:) deleting any unsaved items, which would publish changes upwards, resulting in erroneous snapshots being published.

## References 

IN-693

## Implementation Details

This one's a little… weird. The major change here is that we empty out `HomeViewModel.viewModels` and `HomeViewModel.viewModelSubscriptions` when refreshing the Home feed (with visible items saved). By emptying out these values, we ensure that we are (primarily) not observing changes to Core Data related to unsaved items being deleted from the database. Below is I _believe_ what was happening:

1. On first load (refresh), a new slate lineup is fetched and added to Core Data, resulting in a new snapshot being generated and applied for the slate lineup and its initial slates / recommendations.
2. On second load (pull-to-refresh), the initial slate lineup is deleted, along with any unsaved items, triggering a new snapshot being generated and applied.
3. While the snapshot from `2` is being applied, unsaved recommendations and items are being deleted from Core Data
4. Deleting unsaved items triggers cell view model refreshes, which publish _another_ (reloaded) snapshot (referencing a recommendation that no longer exists within the database)
5. These snapshots could potentially be applied rapidly enough where one snapshot was not yet completely applied before another snapshot was requested to be applied.
6. *_Crash_*

Emptying out the aforementioned values would mean that steps `4` and `5` do not occur (due to the subscription(s) in `HomeRecommendationCellViewModel` being cancelled), preventing `6`.

## Test Steps

- [x] Given I am on the home tab with the feed loaded, 
when I save the first available recommendation and pull to refresh,
then the list should refresh and the app should not crash.

- [x] Given I am on the home tab with the feed loaded, 
when I unsave the first available recommendation and pull to refresh,
then the list should refresh and the app should not crash.

## PR Checklist:

- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/PocketKit/Home/HomeViewModel.swift
+++ b/PocketKit/Sources/PocketKit/Home/HomeViewModel.swift
@@ -70,6 +70,15 @@ class HomeViewModel {
     }
 
     func refresh(_ completion: @escaping () -> Void) {
+        // By emptying out our view models and their subscriptions,
+        // we are ensuring that no previously existing view models
+        // publish changes when source.fetchSlateLineup cascades and deletes any
+        // unsaved items (see HomeRecommendationCellViewModel.init(recommendation:)).
+        // This would cause an erroneous snapshot to be published
+        // (potentially while another snapshot was still being applied).
+        viewModels = [:]
+        viewModelSubscriptions = []
+
         Task {
             try await source.fetchSlateLineup(Self.lineupIdentifier)
             completion()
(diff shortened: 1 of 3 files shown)


PocketKit/Sources/PocketKit/Home/HomeViewModel.swift
```diff
@@ -126,15 +135,21 @@ extension HomeViewModel {
         }
 
         slates.forEach { slate in
+            // Create a topic and slate section _only_ if any
+            // recommendations exist for the given slate
+            let recs = slate.recommendations?
```
jacobsimeon, line 140: It's unfortunate that these core data relationships aren't sets with proper generic types, forcing us to iterate over each item and downcast to the related item type :/

What do you think about adding an extension to `Slate` that encapsulates the downcasting?

Something like
```swift
extension Slate {
  // open to different names here
  // but need to avoid collision with the existing `recommendations` property
  var recommendationsArray: [Recommendation] {
    recommendations.compactMap { $0 as? Recommendation }
  }
}
```

jacobsimeon (approved):
Nice explanation of the cause of the bug and the solution makes sense. Just have one (optional) suggestion for ya.

cyndichin (approved):
Solves the issue when I cherry-pick this commit onto recent saves branch! Thanks for the writeup under implementation details. ⭐

audie1:
Issue resolved, I encountered a few other crashes during testing but they seem unrelated so we will keep an eye out for those during future tests. QA Pass 👍

PR #201, opened 2022-07-11
Show loading indicator when loading slate detail content

dskuza:
## Summary

Adds an activity indicator when first loading a selected slate.

## References 

IN-692

Similar to pull request #196 

## Implementation Details

`SlateDetailViewModel.Section` and `SlateDetailViewModel.Cell` each have a new `loading` case. When `SlateDetailViewModel` is first initialized, it sets its initial snapshot to that of a "loading" snapshot - a snapshot that only contains a single `loading` section with a single `loading` item. `SlateDetailViewController` returns a single section/group/item, which is then used to return a single `LoadingCell`. This cell is simple - it's a `UICollectionViewCell` subclass that houses a `UIActivityIndicatorView` that is always animating (since the loading cell is only ever present for a "loading" section and is not reused otherwise). 

## Test Steps

- [x] Given I have opened the Pocket app,
when I view the Home tab and tap on a slate topic and data has not yet been loaded,
then I should see a spinner in the middle of the screen.

- [x] Given I viewed the Home tab and viewing a slate topic,
when data has been loaded,
then the spinner should disappear and the Home content should be visible.

## PR Checklist:
- [x] Basic Self QA

### Light Mode

Simulator Screen Shot - iPhone 13 - 2022-07-11 at 13 32 35

### Dark Mode

Simulator Screen Shot - iPhone 13 - 2022-07-11 at 13 32 58

### DIFF
--- a/PocketKit/Sources/PocketKit/Home/SlateDetailViewController.swift
+++ b/PocketKit/Sources/PocketKit/Home/SlateDetailViewController.swift
@@ -12,17 +12,11 @@ class SlateDetailViewController: UIViewController {
     }
 
     private lazy var dataSource: UICollectionViewDiffableDataSource<SlateDetailViewModel.Section, SlateDetailViewModel.Cell> = {
-        let registration = UICollectionView.CellRegistration<RecommendationCell, SlateDetailViewModel.Cell> { [weak self] cell, indexPath, cellViewModel in
-            self?.configure(cell, at: indexPath, viewModel: cellViewModel)
-        }
-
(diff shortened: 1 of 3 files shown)


cyndichin (approved):
Looks good on iPad and iPhone. I notice that in the other PR we have a test for creating a loading snapshot. Should we add a test here in `SlateDetailViewModelTests` to cover that as well?

audie1:
Tested via iPad/iPhone in light/dark mode for both devices, no issues found, looking good to go 👍

PR #203, opened 2022-07-12
[WIP] IN-515: Add My List Button for Recent Saves Section

cyndichin:
## Summary
Add My List button for Recent Saves Section. Tapping on this button redirects the user to their MyList view on MyList tab

## References 
IN-515

## Implementation Details
Modified the `SlateHeaderView` to include a button that the user can tap on. `SlateHeaderPresenter` takes on another argument that sets the button title, which is "My List" in this case.  A published property was added to the `HomeViewModel` called `tappedSeeAll` which is updated when the user taps on the button in the section header. The coordinators monitor this property and updates the `selectedSection` which handles the presentation logic. The logic to present "My List" is mainly in `MyListContainerViewController` under `selectMyList`.

## Test Steps
- [ ] Given I am on the Home tab and Recent Saves is visible, when I tap “My List”, then I should be switched to the “My List” tab.
- [ ] Given I am on the Home tab and Recent Saves is visible and the My List tab is switched to Archive, when I tap “My List”, then I should be switched to the “My List” tab and the list should be switched to “My List”

For both cases, it is expected that the user lands on the "MyList" view with the filter selection they previously had.

## PR Checklist:
- [x] Added UI tests
- [ ] Basic Self QA


PocketKit/Sources/PocketKit/Home/HomeViewController.swift
```diff
@@ -216,17 +216,22 @@ extension HomeViewController {
...
+        case SectionHeaderView.kind:
+            let header: SectionHeaderView = collectionView.dequeueReusableView(forSupplementaryViewOfKind: kind, for: indexPath)
             let section = dataSource.sectionIdentifier(for: indexPath.section)
-            
             switch section {
             case .recentSaves:
-                let presenter = SlateHeaderPresenter(name: "Recent Saves")
-                header.attributedHeaderText = presenter.attributedHeaderText
+                let presenter = SectionHeaderPresenter(name: "Recent Saves", buttonTitle: "My List")
+                header.configure(model: .init(attributedHeaderText: presenter.attributedHeaderText,
```
dskuza, line 225: Idea: consolidate the presenter and header model into one. Then, when we configure the header with a model, we can initialize the model with the slate as well as the action, rather than its individual properties (currently set at initialization). To me, the presenter kind of feels like an extra type for sake of extracting computed properties.

PocketKit/Sources/PocketKit/MyList/MyListContainerViewController.swift
```diff
@@ -46,6 +48,21 @@ class MyListContainerViewController: UIViewController {
         fatalError()
     }
     
+    private func resetTitleView() {
```
dskuza, line 51: I remember why we do this - to reset the title view to have "My List" selected, but I'm wondering if it's worth going into the title view and adding something similar to the `selectedIndex` property. Non-blocker.

PocketKit/Sources/PocketKit/MyList/MyListContainerViewModel.swift
```diff
@@ -1,4 +1,15 @@
...
+
+
 class MyListContainerViewModel {
+    enum Selection {
+        case myList
+        case archive
+    }
+
+    @Published
+    var selection: Selection = .myList
```
dskuza, line 11: 👍

dskuza (approved):
Nice! Great work. I left a few comments for things we _could_ address, but don't necessarily have to. We can have a conversation around if we want to tackle any of them.

PocketKit/Sources/PocketKit/Home/HomeViewController.swift
cyndichin: updated!

PocketKit/Sources/PocketKit/MyList/MyListContainerViewController.swift
cyndichin: discussed that this will be addressed if needed in the future and not part of this work

PocketKit/Sources/PocketKit/Home/SectionHeaderView.swift
```diff
@@ -0,0 +1,103 @@
...
+        myListButton.addAction(buttonAction, for: .primaryActionTriggered)
+        myListButton.isHidden = false
+    }
+}
+
+extension SectionHeaderView {
+    static func height(width: CGFloat, slate: SectionHeaderPresenter) -> CGFloat {
+        let adjustedWidth = width - 16
+
+        let rect = slate.attributedHeaderText.boundingRect(
```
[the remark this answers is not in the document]
cyndichin: updated!

dskuza (approved):

PocketKit/Sources/PocketKit/Home/SectionHeaderView.swift
```diff
@@ -0,0 +1,100 @@
...
+        let buttonTitle: String
+        var buttonAction: (() -> ())? = nil
+        
+        var attributedHeaderText: NSAttributedString {
+            NSAttributedString(string: name, style: .sectionHeader)
+        }
+        
+        func height(width: CGFloat) -> CGFloat {
+            attributedHeaderText.sizeFitting(availableWidth: width).height + 16
+        }
```
jacobsimeon, line 75: nice 👍

audie1:
One issue was found during testing but deemed unrelated to the PR, discussion here:
https://pocket.slack.com/archives/C4DL4QSU8/p1658338436498559

Looking good, flawless core functionality of the expected results for each scenario... QA Pass 🎸

PR #206, opened 2022-07-20
Present context menu when tapping and holding links in reader

dskuza:
## Summary

Presents a context menu to the user when tapping and holding links from text views in the reader.

## References 

IN-618

### Presentation

`ArticleComponentTextView` handles adding a `UIContextMenuInteraction` _to_ the text view, so that we can handle long presses. However, a `UITextView` adds its own context menu interaction upon initialization, so we have to filter out these existing context menu interactions and remove them, so we can handle our own. `ArticleComponentTextView` then handles the context menu interaction, only requesting a context menu when the location of the tap is _on_ a link. A request for the context menu to present is then requested up the delegate chain, finally reaching `ReadableViewController`. This view controller requests "external actions" (actions for links aside from the item url itself) from a `ReadableViewModel`, which provides 4 `ItemAction`s: save, open, copy, and share. These `ItemAction`s are used when presenting other context menus within the app, so it made sense (to me) to utilize them again. 

### Save

Save updates `PocketSource` to add a generic `save(url:)` function. This function will check if a `SavedItem` exists for the provided url, and if so, unarchive it (which will mark `isArchived` as false, and update `createdAt` to move the item to the top of the list). Unarchive performs a save operation on an item just as saving does, so it seemed fair to just call "unarchive" to capture both unarchive and move-to-the-top-of-the-list behaviors. If a `SavedItem` does _not_ exist, one will be created, and `save(item:)` will be called, which will perform a save operation, resolving the save via GraphQL query and updating Core Data (thus reflecting the change in My List).

### Open

The open item action piggybacks off of a `ReadableViewModel`'s `presentedWebReaderURL`, which is already subscribed to in a coordinator to appropriately present a `SFSafariViewController` for a provided url.

### Copy

The copy action simply uses `UIPasteboard.general` to set its current value to a url.

### Share

The share action piggybacks off of a `ReadableViewModel`'s `sharedActivity`, which is already subscribed to in a coordinator to appropriately present a share sheet.

## Test Steps

- [x] Given I am reading an article, when I tap and hold a link, a context menu should appear.
- [x] Given I have opened the context menu, when I look at the menu, I should see a preview of the link.
- [x] Given I have opened the context menu, when I tap "Save", the link should be saved to "My List"
- [x] Given I have opened the context menu, when I tap "Save" and the link had previously been saved, it should move to the top of My List
- [x] Given I have opened the context menu, when I tap "Save" and the link had been archived, it should move to the top of My List
- [x] Given I have opened the context menu, when I tap "Open", it should open the article in a browser view
- [x] Given I have opened the context menu, when I tap "Copy link", it should copy the link to the system clipboard
- [x] Given I have opened the context menu, when I tap "Share", it should open a share sheet for the link

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

## Screenshots

Simulator Screen Shot - iPhone 13 Pro - 2022-07-19 at 15 58 07

### DIFF
--- a/PocketKit/Sources/PocketKit/Article/Cells/ArticleComponentTextCell.swift
+++ b/PocketKit/Sources/PocketKit/Article/Cells/ArticleComponentTextCell.swift
@@ -6,6 +6,7 @@ import UIKit
 protocol ArticleComponentTextCellDelegate: AnyObject {
     func articleComponentTextCell(_ cell: ArticleComponentTextCell, didShareText: String?)
     func articleComponentTextCell(_ cell: ArticleComponentTextCell, shouldOpenURL url: URL) -> Bool
+    func articleComponentTextCell(_ cell: ArticleComponentTextCell, contextMenuConfigurationForURL url: URL) -> UIContextMenuConfiguration?
 }
 
 // An object that conforms to this protocol is capable of delegating actions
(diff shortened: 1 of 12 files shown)


PocketKit/Sources/PocketKit/Article/Cells/ArticleComponentTextCell.swift
```diff
@@ -48,6 +57,11 @@ class ArticleComponentTextView: UITextView {
         isEditable = false
         isScrollEnabled = false
         delegate = self
+
+        interactions
+            .filter { type(of: $0) == UIContextMenuInteraction.self }
+            .forEach { removeInteraction($0) }
+        addInteraction(UIContextMenuInteraction(delegate: self))
```
jacobsimeon, line 64: 😎

PocketKit/Sources/Sync/PocketSource.swift
```diff
@@ -499,3 +499,19 @@ extension PocketSource {
...
+            unarchive(item: savedItem)
+        } else {
+            let savedItem: SavedItem = space.new()
+            savedItem.url = url
+            savedItem.createdAt = Date()
+            try? space.save()
+
+            save(item: savedItem)
+        }
+    }
```
jacobsimeon, line 516: 👍

PocketKit/Sources/PocketKit/Article/Cells/ArticleComponentTextCell.swift
```diff
@@ -1,11 +1,13 @@
 import UIKit
+import SafariServices
```
jacobsimeon, line 2: Is this import still needed? Can't tell what we're using from this framework

PocketKit/Sources/PocketKit/Article/Cells/ArticleComponentTextCell.swift
```diff
@@ -25,19 +27,26 @@ extension ArticleComponentTextCell {
...
+        return delegate?.articleComponentTextCell(self, contextMenuConfigurationForURL: url)
+    }
 }
 
 // An object that conforms to this protocol is able to respond to (overridden)
 // events that occur within a PocketTextView.
 protocol ArticleComponentTextViewDelegate: AnyObject {
     func pocketTextViewDidSelectShareAction(_ textView: ArticleComponentTextView)
     func pocketTextView(_ textView: ArticleComponentTextView, shouldOpenURL url: URL) -> Bool
+    func pocketTextView(_ textView: ArticleComponentTextView, contextMenuConfigurationForURL url: URL) -> UIContextMenuConfiguration?
```
jacobsimeon, line 41: Looks like we've had some naming drift when it comes to these methods. The class is `ArticleComponentTextView` but all the methods start with `pocketTextView(_:)`. Whatdya say we rename one or the other to keep things consistent? (I think renaming the methods makes the most sense here)

jacobsimeon (commented):
Nicely done! 

This is 99.9% of the way there, just a couple minor things to address while we're in this code.

PocketKit/Sources/PocketKit/Article/Cells/ArticleComponentTextCell.swift
jacobsimeon: Also `$0 is UIContextMenuInteraction` could work. Not necessary by an means, but would eliminate a touch of noise in the filter condition

dskuza:
@jacobsimeon Changes here: 7aa75228bc28cc9a7fc54fdf5140d9f327fcfde7

I'm thinking once this is approved, I'll squash the two commits.

jacobsimeon (approved):
v nice

PR #209, opened 2022-07-20
Retry immediately on refresh

jacobsimeon:
## Summary
Sometimes network operations fail due to lack of connectivity or server errors. When this happens, those operations go into a "paused" state until they receive a signal to try again. That signal is sent at various times in the app lifecycle.

We consider the user-initiated "Pull to refresh" to be a strong signal that the network connection is probably available, so we initiate any retries immediately upon that interaction.

## References 
https://getpocket.atlassian.net/browse/IN-685

## Implementation Details
* `PocketSource` gains new method `retryImmediately`
* `SavedItemsListViewModel` calls this method when user initiates pull-to-refresh

## Test Steps
1. Use [network link conditioner][0] to cause network traffic to fail 100% of the time
2. Favorite an item
3. Configure network link conditioner to have a solid connection
4. On a second device, not that the item from step 2 is not favorited
5. Pull-to-refresh
6. Note that the item from steps 2 and 4 is now favorited on the second device

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

[0]: https://nshipster.com/network-link-conditioner/

### DIFF
--- a/PocketKit/Sources/PocketKit/MyList/SavedItemsList/SavedItemsListViewModel.swift
+++ b/PocketKit/Sources/PocketKit/MyList/SavedItemsList/SavedItemsListViewModel.swift
@@ -83,6 +83,7 @@ class SavedItemsListViewModel: NSObject, ItemsListViewModel {
 
     func refresh(_ completion: (() -> ())? = nil) {
         source.refresh(completion: completion)
+        source.retryImmediately()
     }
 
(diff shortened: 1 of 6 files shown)


dskuza (approved):

audie1:
Tested this on iPad and iPhone, core functionality is looking good. I did come across one issue while testing that was deemed unrelated to this work:
https://pocket.slack.com/archives/C4DL4QSU8/p1658438889749399

PR #210, opened 2022-07-21
Home: open item when tapping on recent save

dskuza:
## Summary

Add support for opening a recent save (either in Safari or in the reader) on tap.

## References 

IN-690

## Implementation Details

The biggest change here is moving from the `selectedReadableViewModel` to a more generic `selectedReadableType` enum, which supports two cases: recent save, and recommendation. I was hoping that we could just repurpose `ReadableViewModel` and publish a property of this type so that we could subscribe to all published properties in one spot, but there is one difference that prevented me from doing so: the fact that a `RecommendationViewModel` has to support reporting a recommendation, which I did not want to just add to the `ReadableViewModel` protocol since it is not shared with a `SavedItemViewModel`. Otherwise, everything is handled the same - tapping an item (recent save or recommendation) updates `selectedReadableType` as the corresponding case with the appropriate view model. For either case, the coordinators are subscribed to this new property, and then subscribe to the correct published properties of the new value's associated type (i.e view model), also presenting the correct view as necessary.

## Test Steps

- [x] Given I have a recent save and it has been parsed for offline viewing, when I tap on the item, then it should open in the reader view.
- [x] Given I have a recent save and it cannot be parsed for offline viewing (e.g Pocket collection), when I tap on the item, then it should open in a Safari view.

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/PocketKit/Home/CompactHomeCoordinator.swift
+++ b/PocketKit/Sources/PocketKit/Home/CompactHomeCoordinator.swift
@@ -52,7 +52,6 @@ class CompactHomeCoordinator: NSObject {
             case .none:
                 self?.readerSubscriptions = []
             }
-            return
         }.store(in: &subscriptions)
 
         model.$selectedSlateDetailViewModel.receive(on: DispatchQueue.main).sink { [weak self] viewModel in
@@ -198,7 +197,6 @@ class CompactHomeCoordinator: NSObject {
         activityVC.popoverPresentationController?.sourceView = navigationController.splitViewController?.view
 
         activityVC.completionWithItemsHandler = { [weak self] _, _, _, _ in
-//            self?.model.selectedReadableViewModel?.sharedActivity = nil
             self?.model.selectedSlateDetailViewModel?.selectedReadableViewModel?.sharedActivity = nil
         }
 
@@ -245,7 +243,6 @@ extension CompactHomeCoordinator: UINavigationControllerDelegate {
         if viewController === homeViewController {
             model.selectedSlateDetailViewModel?.resetSlate(keeping: 5)
 
-//            model.selectedReadableViewModel = nil
             model.selectedRecommendationToReport = nil
             model.selectedSlateDetailViewModel = nil
         }
@@ -259,7 +256,6 @@ extension CompactHomeCoordinator: UINavigationControllerDelegate {
 
 extension CompactHomeCoordinator: SFSafariViewControllerDelegate {
     func safariViewControllerDidFinish(_ controller: SFSafariViewController) {
-//        model.selectedReadableViewModel?.presentedWebReaderURL = nil
         model.selectedSlateDetailViewModel?.selectedReadableViewModel?.presentedWebReaderURL = nil
     }
 }
--- a/PocketKit/Sources/PocketKit/Home/CompactHomeCoordinator.swift
+++ b/PocketKit/Sources/PocketKit/Home/CompactHomeCoordinator.swift
@@ -43,8 +43,16 @@ class CompactHomeCoordinator: NSObject {
         navigationController.popToRootViewController(animated: false)
         isResetting = true
 
-        model.$selectedReadableViewModel.receive(on: DispatchQueue.main).sink { [weak self] readable in
-            self?.show(readable)
+        model.$selectedReadableType.receive(on: DispatchQueue.main).sink { [weak self] readableType in
+            switch readableType {
+            case .savedItem(let viewModel):
+                self?.show(viewModel)
+            case .recommendation(let viewModel):
+                self?.show(viewModel)
+            case .none:
+                self?.readerSubscriptions = []
+            }
+            return
         }.store(in: &subscriptions)
 
         model.$selectedSlateDetailViewModel.receive(on: DispatchQueue.main).sink { [weak self] viewModel in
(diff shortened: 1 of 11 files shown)


PocketKit/Sources/PocketKit/Home/CompactHomeCoordinator.swift
```diff
@@ -167,7 +198,7 @@ class CompactHomeCoordinator: NSObject {
         activityVC.popoverPresentationController?.sourceView = navigationController.splitViewController?.view
 
         activityVC.completionWithItemsHandler = { [weak self] _, _, _, _ in
-            self?.model.selectedReadableViewModel?.sharedActivity = nil
+//            self?.model.selectedReadableViewModel?.sharedActivity = nil
```
jacobsimeon, line 201: shall we nix this commented out code?

PocketKit/Sources/PocketKit/Home/CompactHomeCoordinator.swift
```diff
@@ -214,7 +245,7 @@ extension CompactHomeCoordinator: UINavigationControllerDelegate {
         if viewController === homeViewController {
             model.selectedSlateDetailViewModel?.resetSlate(keeping: 5)
 
-            model.selectedReadableViewModel = nil
+//            model.selectedReadableViewModel = nil
```
jacobsimeon, line 248: ✂

PocketKit/Sources/PocketKit/Home/CompactHomeCoordinator.swift
```diff
@@ -228,7 +259,7 @@ extension CompactHomeCoordinator: UINavigationControllerDelegate {
 
 extension CompactHomeCoordinator: SFSafariViewControllerDelegate {
     func safariViewControllerDidFinish(_ controller: SFSafariViewController) {
-        model.selectedReadableViewModel?.presentedWebReaderURL = nil
+//        model.selectedReadableViewModel?.presentedWebReaderURL = nil
```
jacobsimeon, line 262: 🔪

PocketKit/Sources/PocketKit/Home/HomeViewModel.swift
```diff
@@ -323,11 +330,48 @@ extension HomeViewModel {
...
+        tracker.track(
+            event: SnowplowEngagement(type: .general, value: nil),
+            contexts(for: cell, at: indexPath)
+        )
+
+        let item = savedItem.item
+        if let isArticle = item?.isArticle, isArticle == false
+            || item?.hasImage == .isImage
+            || item?.hasVideo == .isVideo {
+            presentedWebReaderURL = item?.bestURL
```
jacobsimeon, line 362: Is this now the second place where this complex conditional logic lives? I think it could be pulled out into an extension `Item`.

```swift
extension Item {
  var shouldOpenInWebReader: Bool {
    // complex logic goes here
  }
```

What do you think?

PocketKit/Sources/PocketKit/Main/RegularMainCoordinator.swift
```diff
@@ -302,7 +304,7 @@ class RegularMainCoordinator: NSObject {
 
         slate.$selectedReadableViewModel.receive(on: DispatchQueue.main).sink { [weak self] readable in
             if readable != nil {
-                self?.model.home.selectedReadableViewModel = nil
+//                self?.model.home.selectedReadableViewModel = nil
```
jacobsimeon, line 307: ✁

PocketKit/Sources/PocketKit/Main/RegularMainCoordinator.swift
```diff
@@ -410,7 +412,7 @@ extension RegularMainCoordinator: SFSafariViewControllerDelegate {
     func safariViewControllerDidFinish(_ controller: SFSafariViewController) {
         model.myList.savedItemsList.selectedItem?.clearPresentedWebReaderURL()
         model.myList.archivedItemsList.selectedItem?.clearPresentedWebReaderURL()
-        model.home.selectedReadableViewModel?.presentedWebReaderURL = nil
+//        model.home.selectedReadableViewModel?.presentedWebReaderURL = nil
```
jacobsimeon, line 415: ✄

jacobsimeon (changes requested):
Looks good for the most part. Just some commented out code that needs deleted and a suggestion about extracting some business logic

PocketKit/Sources/PocketKit/Main/RegularMainCoordinator.swift
```diff
@@ -148,14 +148,16 @@ class RegularMainCoordinator: NSObject {
...
-                self?.model.myList.savedItemsList.selectedItem = nil
-                self?.model.myList.archivedItemsList.selectedItem = nil
+        model.home.$selectedReadableType.receive(on: DispatchQueue.main).sink { [weak self] readableType in
+            switch readableType {
+            case .recommendation(let viewModel):
+                self?.show(viewModel)
+            case .savedItem(let viewModel):
+                self?.show(viewModel)
+            case .none:
+                return
```
cyndichin, line 158: do we also want to do`readerSubscriptions = []` here like we do in `CompactHomeCoordinator`?

Tests iOS/HomeTests.swift
```diff
@@ -55,8 +55,6 @@ class HomeTests: XCTestCase {
         }
 
         try server.start()
-
-        app.launch()
```
cyndichin, line 59: curious - why did we decide to remove this and add it to the individual tests?

dskuza: for certain Home tests, we want use different responses than those set up in `setUp` (e.g we want the initial "My List" sync to return a different list for recent saves in web view). if we launch the app during `setUp`, though, our sync requests will already be fulfilled with the responses assigned in `setUp` since the app is launched (initiating the initial sync request chain, for example, for home, my list, etc) and started before our tests are run (after the app is launched, and already loaded data), where we attempted to override the responses. it's similar to a "race condition", where our app is already loaded and running before a test sets up its expected API responses.

cyndichin: got it, thank you!

PocketKit/Sources/PocketKit/Main/RegularMainCoordinator.swift
dskuza: good catch!

dskuza:
@cyndichin @jacobsimeon comments have been addressed in https://github.com/Pocket/pocket-ios/pull/210/commits/9fd554bbadaad7494fa258aa16f8f68a899cbe4e. can you re-review? once approved, i'll squash before merging.

PocketKit/Sources/PocketKit/Item/Item+Extensions.swift
```diff
@@ -19,3 +19,11 @@ public extension SavedItem {
...
     }
 }
+
+public extension Item {
+    var shouldOpenInWebView: Bool {
+        return isArticle == false
+        || hasImage == .isImage
+        || hasVideo == .isVideo
+    }
+}
```
jacobsimeon, line 29: 😎

jacobsimeon (approved):

cyndichin (approved):

audie1:
Looks good! Items are opening into the correct Reader/Safari UI as explained in the test steps. No issues found on iPad/iPhone

PR #212, opened 2022-07-23
Home crasher

jacobsimeon:
## Summary
Immediately processing changes to the `isSaved` property of a recommendation cell's view model can be problematic due to the possibility that other significant changes are being made to the underlying recommendation in core data (i.e. it's being deleted).

Instead, we observe changes made at the CoreData level so that we don't send collection view snapshots too often.

## References 
https://getpocket.atlassian.net/browse/IN-711

## Implementation Details
Instead of using Key-value observation to respond to changes to a recommendation, we use an NSFetchedResultsController. This means we get notified of changes when the context is saved, as opposed to immediately when the property changes.

## Test Steps
- Launch the app
- Make sure it doesn't crash ;)

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/PocketKit/Home/HomeRecommendationCellViewModel.swift
+++ b/PocketKit/Sources/PocketKit/Home/HomeRecommendationCellViewModel.swift
@@ -2,45 +2,39 @@ import Foundation
 import Sync
 import Combine
 import Textile
+import CoreData
 
 
-class HomeRecommendationCellViewModel {
-    @Published
-    private(set) var isSaved: Bool
-
-    private var subscriptions: Set<AnyCancellable> = []
-
+class HomeRecommendationCellViewModel: NSObject {
+    let updated: PassthroughSubject<Void, Never> = .init()
     let recommendation: Recommendation
+
+    var isSaved: Bool {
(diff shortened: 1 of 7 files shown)


dskuza (changes requested):
Code-wise: approved. PR-wise: requesting changes, since there are build errors in CI. Other than that, with the live review and the code review, I think this all makes sense!

cyndichin (commented):
Testing out the save actions - seems like sometimes pressing on the save button for one cell will cause an issue to save another cell too.

https://user-images.githubusercontent.com/6743397/180850023-7e914054-4bcb-4e99-adea-1ce6febcf16c.MP4

audie1:
No crashes found on iPad/iPhone, looking good!

jacobsimeon:
> Testing out the save actions - seems like sometimes pressing on the save button for one cell will cause an issue to save another cell too.

Nice catch @cyndichin. I'm actually seeing that issue on `develop` as well so I don't think it's caused by this change.
Created [IN-724][in-724] to track this issue.

[in-724]: https://getpocket.atlassian.net/browse/IN-724

dskuza (approved):

PR #213, opened 2022-07-25
IN-700: Home Design Polish

cyndichin:
## Summary
Polishes the Home Design to better match the Figma File and added horizontal carousel for the recommendation cells.

## References 
IN-700, IN-707

## Implementation Details
Separated the Slate section to use two types of section `.slateHero` and `.slateCarousel`. Also, separated the slate cells from `.recommendation` to using `.recommendationHero` and `.recommendationCarousel`. Modified the `RecommendationCell` which is used for the hero card and added a new cell called `HomeCarouselItemCell` that is used for both recent saves and the horizontal recommendation cells. Made a couple of design tweaks such as adding section titles to the See All views, changing report button flow and made sure design worked for both dark / light mode. 

## Test Steps
- [ ] Given I am on the home tab, when I navigate between the slates and recent saves, the cells should behave as previously but with the new design
- [ ] Given I am on the home tab, When I swipe from right to left on the cells below the hero cell, Then the cells should scroll and reveal more recommendation cells
- [ ] Given I am on the home tab, when I tap on the see All button for the slate headers, then I should be able to see the slate section title in the next view
- [ ] Given I am on the home tab, when I tap on overflow button for the recommendation cells, then I should be able to see a menu option that shows report

Note: This work does not take care of dynamic fonts. That work will be addressed in a new ticket: https://getpocket.atlassian.net/browse/IN-723

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

## Screenshots

https://user-images.githubusercontent.com/6743397/180838659-fe2ec5ff-08b0-4b7e-871b-9a8980b5b447.mp4

### DIFF
--- a/PocketKit/Sources/PocketKit/Home/HomeCarouselItemCell.swift
+++ b/PocketKit/Sources/PocketKit/Home/HomeCarouselItemCell.swift
@@ -1,0 +1,221 @@
+import UIKit
+import Kingfisher
+import Textile
+
+protocol HomeCarouselItemCellModel {
+    var thumbnailURL: URL? { get }
+    var saveButtonMode: RecommendationSaveButton.Mode? { get }
+    var favoriteAction: ItemAction? { get }
+    var overflowActions: [ItemAction]? { get }
+    var saveAction: ItemAction? { get }
+    var attributedTitle: NSAttributedString { get }
+    var attributedDomain: NSAttributedString { get }
+    var attributedTimeToRead: NSAttributedString { get }
+}
+
+class HomeCarouselItemCell: UICollectionViewCell {
+    enum Constants {
+        static let cornerRadius: CGFloat = 16
+        static let maxTitleLines = 3
+        static let maxDetailLines = 2
+        static let actionButtonImageSize = CGSize(width: 20, height: 20)
+        static let layoutMargins = UIEdgeInsets(top: Margins.normal.rawValue, left: Margins.normal.rawValue, bottom: Margins.normal.rawValue, right: Margins.normal.rawValue)
+    }
+
+    private let titleLabel: UILabel = {
+        let label = UILabel()
+        label.numberOfLines = Constants.maxTitleLines
+        label.setContentCompressionResistancePriority(.required, for: .vertical)
+        return label
+    }()
+
+    private let domainLabel: UILabel = {
+        let label = UILabel()
+        label.numberOfLines = Constants.maxDetailLines
+        return label
+    }()
+    
+    private let timeToReadLabel: UILabel = {
+        let label = UILabel()
+        label.numberOfLines = Constants.maxDetailLines
+        return label
+    }()
+
+    private let thumbnailView: UIImageView = {
+        let imageView = UIImageView()
+        imageView.layer.cornerRadius = Constants.cornerRadius
+        imageView.layer.masksToBounds = true
+        imageView.backgroundColor = UIColor(.ui.grey6)
+        imageView.contentMode = .center
+        return imageView
+    }()
+
+    private let favoriteButton: UIButton = {
+        var config = UIButton.Configuration.plain()
+        config.contentInsets = .zero
+
+        let button = UIButton(configuration: config, primaryAction: nil)
+        button.accessibilityIdentifier = "favorite"
+        return button
+    }()
+    
+    let saveButton: RecommendationSaveButton = {
+        let button = RecommendationSaveButton()
+        button.accessibilityIdentifier = "save-button"
+        return button
+    }()
+
+    private let overflowButton: UIButton = {
+        let button = RecommendationOverflowButton()
+        button.accessibilityIdentifier = "overflow-button"
+        button.showsMenuAsPrimaryAction = true
+        return button
+    }()
+
+    private let mainContentView = UIView()
+    
+    private let mainContentStack: UIStackView = {
+        let stack = UIStackView()
+        stack.alignment = .top
+        stack.distribution = .equalSpacing
+        stack.spacing = 20
+        stack.axis = .horizontal
+        return stack
+    }()
+
+    private let bottomStack: UIStackView = {
+        let stack = UIStackView()
+        stack.distribution = .equalSpacing
+        stack.axis = .horizontal
+        return stack
+    }()
+    
+    private let subtitleStack: UIStackView = {
+        let stack = UIStackView()
+        stack.distribution = .fillProportionally
+        stack.axis = .vertical
+        stack.spacing = 4
+        return stack
+    }()
+    
+    private let buttonStack: UIStackView = {
+        let stack = UIStackView()
+        stack.axis = .horizontal
+        stack.spacing = 0
+        return stack
+    }()
+
+    private var thumbnailWidthConstraint: NSLayoutConstraint!
+
+    override init(frame: CGRect) {
+        super.init(frame: frame)
+        accessibilityIdentifier = "home-carousel-item"
+        
+        contentView.addSubview(mainContentStack)
+        contentView.addSubview(bottomStack)
+        contentView.layoutMargins = Constants.layoutMargins
+
+        mainContentStack.translatesAutoresizingMaskIntoConstraints = false
+        thumbnailView.translatesAutoresizingMaskIntoConstraints = false
+        bottomStack.translatesAutoresizingMaskIntoConstraints = false
+
+        thumbnailWidthConstraint = thumbnailView.widthAnchor.constraint(
+            equalToConstant: StyleConstants.thumbnailSize.width
+        ).with(priority: .required)
+
+        contentView.layoutMargins = Constants.layoutMargins
+        NSLayoutConstraint.activate([
+            mainContentStack.topAnchor.constraint(equalTo: contentView.layoutMarginsGuide.topAnchor),
+            mainContentStack.leadingAnchor.constraint(equalTo: contentView.layoutMarginsGuide.leadingAnchor),
+            mainContentStack.trailingAnchor.constraint(equalTo: contentView.layoutMarginsGuide.trailingAnchor),
+
+            thumbnailView.heightAnchor.constraint(equalToConstant: StyleConstants.thumbnailSize.height).with(priority: .required),
+            thumbnailWidthConstraint!,
+
+            bottomStack.leadingAnchor.constraint(equalTo: mainContentStack.leadingAnchor),
+            bottomStack.trailingAnchor.constraint(equalTo: mainContentStack.trailingAnchor),
+            bottomStack.bottomAnchor.constraint(equalTo:  contentView.layoutMarginsGuide.bottomAnchor).with(priority: .required),
+        ])
+        
+        [UIView(), domainLabel, timeToReadLabel, UIView()].forEach(subtitleStack.addArrangedSubview)
+        [favoriteButton, saveButton, overflowButton].forEach(buttonStack.addArrangedSubview)
+        [titleLabel, thumbnailView].forEach(mainContentStack.addArrangedSubview)
+        [subtitleStack, UIView(), buttonStack].forEach(bottomStack.addArrangedSubview)
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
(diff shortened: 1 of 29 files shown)


PocketKit/Sources/PocketKit/Home/HomeCollectionViewSectionProvider.swift
```diff
@@ -18,194 +29,87 @@ class HomeViewControllerSectionProvider {
...
-                layoutSize: NSCollectionLayoutSize(
-                    widthDimension: .fractionalWidth(1),
-                    heightDimension: .absolute((miniCardHeight + dividerHeight) * 2)
-                ),
-                subitem: twoUpInner,
-                count: numberOfRows
-            ),
-            height: miniCardHeight
+        let groupSize = NSCollectionLayoutSize(widthDimension: .fractionalWidth(0.80), heightDimension: .absolute(Constants.carouselHeight))
+        let group = NSCollectionLayoutGroup.horizontal(layoutSize: groupSize, subitems: [item])
```
cyndichin, line 103: calculating the number of items for both the carousel and recent saves caused a couple of crashes, but using the method where we rely on the data sources vs count seems to work best in both for recent saves / carousel. Happy to pair on this to clarify. 

Alternative that led to issues:

```
let groupSize = NSCollectionLayoutSize(widthDimension: .fractionalWidth(0.8*Double(numberOfRecentSaves)), heightDimension: .absolute(Constants.carouselHeight))
 let group = NSCollectionLayoutGroup.horizontal(layoutSize: groupSize, subitem: item, count: numberOfRecentSaves)
```

PocketKit/Sources/PocketKit/Home/RecommendationCell.swift
```diff
@@ -132,111 +113,44 @@ class RecommendationCell: UICollectionViewCell {
...
-        + Mini.textStackTopMargin
-        + titleHeight
-        + Mini.textStackSpacing
-        + detailHeight
-        + Mini.buttonStackTopMargin
-        + Self.saveButtonHeight
-        + Self.layoutMargins.bottom
-    }
-
     static func fullHeight(viewModel: RecommendationCellViewModel, availableWidth: CGFloat) -> CGFloat {
```
cyndichin, line 154: leaving height calculation code just in case useful context for when tackling dynamic fonts

PocketKit/Sources/PocketKit/Home/HomeCarouselItemCell.swift
```diff
@@ -0,0 +1,285 @@
...
+        init(viewModel: HomeRecommendationCellViewModel) {
+            self.title = viewModel.title
+            self.domain = viewModel.domain
+            self.timeToRead = viewModel.timeToRead
+            self.thumbnailURL = viewModel.imageURL
+            self.saveButtonMode = viewModel.saveButtonMode
+        }
+
+        var favoriteAction: ItemAction? = nil
+        var overflowActions: [ItemAction]? = nil
```
jacobsimeon, line 174: I found myself looking above the initializer for these properties since they're not computed. Would you mind moving them?

Generally speaking, here's how we like to approach ordering property and function declarations:
```swift
class Foo {
    // declare public properties at the top
    var bar: Bar?
    let baz: Baz

    // private properties come next
    private let qux: Qux

    // then comes any initializers (public first, then private)
    public init() {
        self.qux = Qux.shared
    }
    
    private init(qux: Qux) {
        self.qux = qux
    }

    // method declarations come after the initializer
    func quux() {
    }

    // also, complex computed properties can be placed along with public functions
    // since, practically speaking, they are just parameterless functions
    var quuz: Quuz {
        if foo != nil {
            foo.bar().map { $0.otherThing }
        } else {
            []
        }
    }
}

// also can be nice to group like functions in an extension w/ a `MARK: - Group Title` comment
// MARK: - Network calls
extension Foo {

    func fetchSomething() async {
        // network related code
    }
}
```

There is probably some debate between @dskuza and I regarding placing computed properties above or below the initializer, so this isn't set in stone. Also there are reasons to deviate from this convention (like if a property is closely related to a single function) but this is our general approach.

PocketKit/Sources/PocketKit/Home/HomeCarouselItemCell.swift
```diff
@@ -0,0 +1,285 @@
...
+        
+        if let mode = model?.saveButtonMode {
+            saveButton.isHidden = false
+            saveButton.mode = mode
+        } else {
+            saveButton.isHidden = true
+        }
+
+        let menuActions = model?.overflowActions?.compactMap(UIAction.init) ?? []
+        overflowButton.menu = UIMenu(children: menuActions)
```
jacobsimeon, line 221: Let's update the recommendation cell (hero) to follow this same pattern of setting the overflow button's menu within the view (as opposed to doing so in the view controller)

PocketKit/Sources/PocketKit/Home/HomeViewController.swift
```diff
@@ -190,10 +178,26 @@ extension HomeViewController {
...
+            
+            guard let viewModel = model.recommendationCarouselViewModel(for: objectID) else {
+                cell.model = nil
+                return cell
+            }
+            
+            cell.model = viewModel
+            
+            if let action = model.saveAction(for: item, at: indexPath), let uiAction = UIAction(action) {
+                cell.saveButton.addAction(uiAction, for: .primaryActionTriggered)
```
jacobsimeon, line 196: Same idea here, let's set the save button's action in the view's `reconfigure()` method. Ideally, setting the cell's `model` property should be the only thing we need to do in order for the view to be fully configured.

PocketKit/Sources/PocketKit/Home/HomeViewModel.swift
```diff
@@ -279,33 +302,72 @@ extension HomeViewModel {
...
+        }
+
+        return count - 1
+    }
+    
+    func slateName(with id: NSManagedObjectID) -> String? {
+        (slateLineupController
+            .slateLineup?
+            .slates?
+            .first { ($0 as? Slate)?.objectID == id } as? Slate)?.name
```
jacobsimeon, line 358: could refactor to `slate(with: id)?.name` to avoid duplicating the code for finding the slate

PocketKit/Sources/PocketKit/Home/HomeViewModel.swift
```diff
@@ -279,33 +302,72 @@ extension HomeViewModel {
...
+    }
+    
+    func slate(with id: NSManagedObjectID) -> Slate? {
+        (slateLineupController
+            .slateLineup?
+            .slates?
+            .first { ($0 as? Slate)?.objectID == id } as? Slate)
```
jacobsimeon, line 365: we could use `(source.object(id: id) as? Slate)` here (since it's a O(1) operation) as opposed to iterating through all the slates.

PocketKit/Sources/PocketKit/Home/SectionHeaderView.swift
```diff
@@ -80,6 +86,17 @@ extension SectionHeaderView {
...
     }
     
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        super.traitCollectionDidChange(previousTraitCollection)
+        myListButton.configurationUpdateHandler = { button in
+            var config = button.configuration
+            config?.image = UIImage(asset: .chevronRight)
+                .withTintColor(UIColor(.ui.lapis1), renderingMode: .alwaysOriginal)
+                .resized(to: SectionHeaderView.buttonImageSize)
+            button.configuration = config
```
jacobsimeon, line 96: Wondering why this code needs to live within `traitCollectionDidChange` and not in `updateButtonConfiguration`?

PocketKit/Sources/PocketKit/Home/SlateDetailViewController.swift
```diff
@@ -50,10 +50,8 @@ class SlateDetailViewController: UIViewController {
         super.init(nibName: nil, bundle: nil)
 
         view.accessibilityIdentifier = "slate-detail"
-
-        title = nil
-        navigationItem.largeTitleDisplayMode = .never
-
+        navigationItem.title = model.slateName
+        navigationItem.setValue(true, forKey: "__largeTitleTwoLineMode")
```
jacobsimeon, line 54: Hmm, this seems a bit...unconventional 😆 
Generally speaking using KVC to set a private property is going to be brittle and/or dangerous since the underlying property could go away or get renamed (which would cause a crash here).

What are we trying to accomplish here? Wondering if there's a safer way to get the desired result without having to use a private API

PocketKit/Sources/PocketKit/Home/SlateDetailViewController.swift
```diff
@@ -253,7 +222,7 @@ private extension SlateDetailViewController {
 
             if let action = self.model.reportAction(for: viewModelCell, at: indexPath),
                let uiAction = UIAction(action) {
-                cell.overflowButton.addAction(uiAction, for: .primaryActionTriggered)
+                cell.overflowButton.menu = UIMenu(children: [uiAction])
             }
```
jacobsimeon, line 226: I know the code that came before was setting the action here in the view controller, but I really would prefer to move it into the cell if possible. (similar idea to what I was getting at w/ HomeViewController et al)

jacobsimeon (commented):
Looking really good! Just a few comments about odds and ends.

The main thing I want to call out is that I want to move towards setting all view properties, including actions and menus, within the view itself. Anything you need to configure the view should be passed in via the view model.

The other suggestion is around configuring all the view models with their appropriate actions at the time they are initialized, as opposed to asking for the `overflowActions` and `saveAction` after the fact. Let me know if you have questions about that or if it doesn't seem doable right now.

This looks great! Very nicely done thus far

PocketKit/Sources/PocketKit/Home/SectionHeaderView.swift
cyndichin: updateButtonConfiguration wasn't being called when switching from dark mode to light mode (and vice versa) so the button colors weren't being updated

PocketKit/Sources/PocketKit/Home/HomeCollectionViewSectionProvider.swift
cyndichin: @jacobsimeon not sure why I can't reply to your message on 

```
I actually think the crash that you were seeing is probably resolved per https://github.com/Pocket/pocket-ios/pull/212. I understand that this works but having a layout group with a different number of items from the datasource's snapshot seems risky. Can you test out using a repeating item now that the other bug has been fixed?
```

The crash still exists after rebasing from develop - will see what's going on, but maybe we can pair on this :)

PocketKit/Sources/PocketKit/Home/SectionHeaderView.swift
cyndichin: discussed with @jacobsimeon, the issue was that we were resizing after applying tint color

PocketKit/Sources/PocketKit/Home/SlateDetailViewController.swift
cyndichin: updated to using: 

```
        let largeTitleTwoLineMode = "_largeTitleTwoLineMode"
        if class_getProperty(UINavigationItem.self, largeTitleTwoLineMode) != nil {
            navigationItem.setValue(true, forKey: largeTitleTwoLineMode)
        }
```

PocketKit/Sources/PocketKit/Home/HomeViewModel.swift
```diff
@@ -286,33 +318,66 @@ extension HomeViewModel {
...
+        }
+
+        return count - 1
+    }
+    
+    func numberOfRecentSavesItem() -> Int {
+//        guard snapshot.indexOfSection(.recentSaves) != nil else { return 0 }
+//        return snapshot.itemIdentifiers(inSection: .recentSaves).count
+        return 5
+    }
```
cyndichin, line 372: temporary implementation until we refactor home view model

jacobsimeon (approved):
Amazing work! This is going to be a huge improvement to the Home experience!

PR #214, opened 2022-07-27
Use consistent action identifiers when attaching button actions

jacobsimeon:
When adding a button action *that should replace another button action*,
it's crucial that the two actions have the same identifier.

Per Apple's documentation:
> UIActions are uniqued based on their identifier, and subsequent
> actions with the same identifier replace previously added actions.

In the case of save/archive actions, the save action should be replaced
by the archive action, so the two need to have the same identifier.

https://getpocket.atlassian.net/browse/IN-724

### DIFF
--- a/PocketKit/Sources/PocketKit/Home/HomeRecommendationCellViewModel.swift
+++ b/PocketKit/Sources/PocketKit/Home/HomeRecommendationCellViewModel.swift
@@ -8,7 +8,7 @@ import CoreData
 class HomeRecommendationCellViewModel {
     let recommendation: Recommendation
     let overflowActions: [ItemAction]?
-    let saveAction: ItemAction?
+    let primaryAction: ItemAction?
     
     var isSaved: Bool {
         recommendation.item?.savedItem != nil &&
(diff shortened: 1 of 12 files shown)


dskuza:
@jacobsimeon We had a similar in another screen. I'm not opposed to this fix, but conceptually, the archive action is separate from the save action, so part of me would like to keep the identifiers unique, but I understand why you consolidated them. I'm just wondering if there's another solution. We did add a `recommendationPrimary` item action that could be reused here if we'd like. These actions are used in menus around the app. Will this affect those menus?

In `HomeViewModel`:
```swift
return .recommendationPrimary { [weak self] _ in
    if viewModel.isSaved {
        self?.archive(cell, at: indexPath)
    } else {
        self?.save(cell, at: indexPath)
    }
}
        ```

jacobsimeon:
I see what you're saying @dskuza and I like it. I'll change this up to use the recommendationPrimary action

dskuza (approved):

audie1:
No issues found on iPad/iPhone during testing, I did a lot of toggling and galumphing around the save button toggle, was never able to get the issue to resurface, QA Pass.

PR #216, opened 2022-08-01
Improve stability and reliability of Home and other code cleanup

jacobsimeon:
## Summary
This change is tantamount to a rewrite of `HomeViewModel`. The impetus for the change was to eliminate the various usages of multiple `NSFetchedResultsController` objects to observe changes to core data models. Doing all of this observation within individual objects causes unexpected behavior due to an unpredictable order in which the change handlers are invoked.
Instead of having multiple handlers, the HomeViewModel now simply responds to the `NSManagedObjectContext.didSaveObjectsNotification` and determines any cells that need to be updated (e.g. a SavedItem that has been favorited or a Recommendation that has been saved).

See notes below for other changes to adjacent code.

## References 
https://getpocket.atlassian.net/browse/IN-713
https://getpocket.atlassian.net/browse/IN-724
https://getpocket.atlassian.net/browse/IN-728

## Implementation Details
- HomeViewModel
    - Observe changes to recommendations and recent saves by directly handling core `NSManagedObjectContext.didSaveObjectsNotification`
    - Add `SeeAll` enum that encapsulates which “See All” button users tap and change the `tappedSeeAll` property to be of this new type. Coordinators can observe this property and update the navigation stack directly without having to circle back around to set `model.home.selectedSlateDetail`. 
    - Remove the `selectedSlateDetail` property entirely.
    - Simplify building of recommendation view models and their actions
        - Instead of passing around the cell, and then having to unwrap it to get to the associated `Recommendation` object, we unwrap the model one time and then pass the recommendation directly to various actions and analytics code.
    - provide `SectionHeaderViewModel` directly instead of forcing clients to create it on their own
    - Eliminate need to keep ahold of:
        - viewModels and viewModelSubscriptiosn (since we don’t need to observe changes anymore)
        - RecentSavesController and recentSavesViewModels
    - Re-organize methods by grouping related methods together in extensions labeled with `MARK: - Name of method group`
- Create `NSLayoutSection.empty()` method that provides a section that contains a single item w/ zero height & width
- Return `.empty()` section for recommendation carousels if the view model reports `numberOfCarouselItems` == 0
- HomeRecommendationViewModel:
    - Remove usage of fetched results controller to observe changes to `isSaved` since this is now done by HomeViewModel
- SlateDetailViewModel
    - Directly accept a `Slate` during initialization instead of passing in the ID and name
    - Observe core data changes via `NSManagedObjectContext.didSaveObjectsNotification` (similar to HomeViewModel)
    - Don’t reset slate to only five recommendations when popping back to Home (since that view knows to limit the number of carousel items to a max of 5)
- Remove RecentSavesController, SlateLineupController, and SlateController
- SlateService
    - Don’t delete unsaved items when fetching slate details
- Add RecentSavesCellElement that supports interacting with buttons
- Extract “SeeAll” enum so we can encapsulate the selected slate view model
- CompactHomeCoordinator: delegate to parent coordinator when tapping “My List” on Home
- SyncTests and PocketKitTests
    - Use consistent pattern for creating prebuilt core data models by adding extension methods to `Space` and eliminating static `.build()` methods (e.g. `SavedItem.build()`).

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA


PocketKit/Sources/PocketKit/Home/HomeViewModel.swift
```diff
@@ -44,391 +43,209 @@ class HomeViewModel {
...
-            } else {
-                self?.save(cell, at: indexPath)
+                guard !recs.isEmpty else {
+                    continue
+                }
+                snapshot.appendSections([.slateCarousel(slate.objectID)])
+                snapshot.appendItems(
+                    recs[0...min(3, recs.endIndex - 1)].map { .recommendationCarousel($0.objectID) },
+                    toSection: .slateCarousel(slate.objectID)
+                )
```
cyndichin, line 171: previously, we didn't limit it to 3, why not just rely on passing the right data and use `recs.endIndex - 1`?

PocketKit/Sources/PocketKit/Home/HomeViewModel.swift
```diff
@@ -440,111 +257,255 @@ extension HomeViewModel {
...
+        case .loading, .slateCarousel:
+            return nil
+        }
+    }
+}
+
+// MARK: - Recent Saves Model & Actions
+extension HomeViewModel {
+    func numberOfRecentSavesItem() -> Int {
+        return 5
```
cyndichin, line 291: do we want to use the `recentSaves.count` from the fetchRequest or do we know that our limit is 5, so it's okay to put 5 here? Maybe add a constant to be used in both places?

cyndichin (commented):
Looks great so far, def seems a lot cleaner than what we had before! Few comments and I noticed that there was an issue when navigating to see all from the slates. Sometimes the items reload in different order and shows up in different collections as well.  See video attached. 

https://user-images.githubusercontent.com/6743397/182214910-4f520e52-38c4-429a-8faa-4da0e3c3a643.MP4

PocketKit/Sources/PocketKit/Home/HomeCollectionViewSectionProvider.swift
```diff
@@ -96,7 +95,10 @@ class HomeViewControllerSectionProvider {
     
     func carouselSection(for slateID: NSManagedObjectID, in viewModel: HomeViewModel, width: CGFloat) -> NSCollectionLayoutSection? {
         let numberOfCarouselItems = viewModel.numberOfCarouselItemsForSlate(with: slateID)
-        guard numberOfCarouselItems > 0 else { return nil }
+        guard numberOfCarouselItems > 0 else {
+            return .empty()
```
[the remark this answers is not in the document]
jacobsimeon: Yeah anywhere that we are returning `nil` for a `NSLayoutSection` should probably be changed to return `.empty()`. I can make that change.

PocketKit/Sources/PocketKit/Home/HomeViewModel.swift
jacobsimeon: Good catch, we should be returning the actual number of recents saves in the database

PocketKit/Sources/PocketKit/Home/HomeViewModel.swift
jacobsimeon: So we're accounting for this scenario:

1. User launches app and lands on Home
2. We fetch the entire slate lineup, with a maximum number of recommendations per slate of 5
3. User taps "See All" for a given slate
4. We fetch up to 25 recommendations for the slate the user tapped on
5. User pops back to Home
6. Now we have 25 recommendations in the database but we don't want that many in the carousel

We were previously accounting for this by using some more brute force logic to delete all the extra recommendations when popping back to home (there was a line in `CompactHomeCoordinator` that did this). But I thought it made more sense to leave the recommendations in the database until a full refresh of the slate lineup (that way we show all the recommendations if the user taps "See All" again).

PocketKit/Sources/PocketKit/Home/HomeViewModel.swift
```diff
@@ -44,391 +43,209 @@ class HomeViewModel {
...
-                .compactMap { $0 as? Recommendation }
-            ?? []
-
-            guard recs.isEmpty == false else {
+    func select(cell: HomeViewModel.Cell, at indexPath: IndexPath) {
+        switch cell {
+        case .loading:
+            return
+        case .recentSaves(let objectID):
+            guard let savedItem = source.mainContext.object(with: objectID) as? SavedItem else {
```
dskuza, line 186: We also have `source.object(id:)` at our disposal.

dskuza (commented):
I don't have much to add to @cyndichin's review, so I'll await those changes before I slap an approval on this pull request.

PocketKit/Sources/PocketKit/Home/HomeViewModel.swift
cyndichin: got it thanks @jacobsimeon  :) This is the function you are referring to right? `resetSlate(keeping: 5)`

PocketKit/Sources/PocketKit/Home/HomeCollectionViewSectionProvider.swift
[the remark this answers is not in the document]
cyndichin: thanks!

cyndichin (approved):
Looking great :D Home is a lot more stable! Thanks for improving it

PR #217, opened 2022-08-02
IN-700: Accessibility - Dynamic Fonts

cyndichin:
## Summary
Allow users to be able to read and view items in their preferred font size for Home, Slate Detail, My List / Archive

## References 
IN-723

## Implementation Details
Home
* Carousel cells are using a static height of 146, which gets scaled using `UIFontMetrics.default.scaledValue`, but because of additional white space for larger fonts, we cap this value to 300
* Hero Recommendation cells are configured based on the height calculations of the margins and views within the cell
* Set border on the view itself vs. content view because the content view was not resizing properly with the height change of the cell
* Added `maxScaleSize` to our Style extension as some labels we don't want to scale larger than a certain size (i.e. time to read, see all button, save)

SlateDetail
* Uses similar height calculation for a single card recommendation as Home 
* Difference is that we need to calculate the total height based on all the items (which was implemented previously)

MyList / Archive
* Remove `selection` view and configure background color on `button` so that view adjusts itself properly when preferred font changes
* Set `label.adjustsFontForContentSizeCategory = true` as text was not updating when switching between sizes

## Test Steps
- [ ] Given that I change my font settings to use small font, when I navigate to the app, then I should be able to have a good user experience in that size.
- [ ] Given that I change my font settings to use extra large font, when I navigate to the app, then I should be able to have a good user experience in that size.

## PR Checklist:
- [x] Basic Self QA

## Screenshots
Simulator Screen Shot - iPhone 8 - 2022-08-02 at 12 22 08
Simulator Screen Shot - iPhone 8 - 2022-08-02 at 12 22 28

### DIFF
--- a/PocketKit/Sources/PocketKit/Home/HomeCarouselItemCell.swift
+++ b/PocketKit/Sources/PocketKit/Home/HomeCarouselItemCell.swift
@@ -25,6 +25,7 @@ class HomeCarouselItemCell: UICollectionViewCell {
     private let titleLabel: UILabel = {
         let label = UILabel()
         label.numberOfLines = Constants.maxTitleLines
+        label.adjustsFontForContentSizeCategory = true
         label.setContentCompressionResistancePriority(.required, for: .vertical)
         return label
     }()
@@ -32,12 +33,14 @@ class HomeCarouselItemCell: UICollectionViewCell {
     private let domainLabel: UILabel = {
         let label = UILabel()
         label.numberOfLines = Constants.maxDetailLines
+        label.adjustsFontForContentSizeCategory = true
         return label
     }()
     
     private let timeToReadLabel: UILabel = {
         let label = UILabel()
         label.numberOfLines = Constants.maxDetailLines
+        label.adjustsFontForContentSizeCategory = true
         return label
     }()
 
@@ -118,6 +121,7 @@ class HomeCarouselItemCell: UICollectionViewCell {
         mainContentStack.translatesAutoresizingMaskIntoConstraints = false
         thumbnailView.translatesAutoresizingMaskIntoConstraints = false
         bottomStack.translatesAutoresizingMaskIntoConstraints = false
+        contentView.translatesAutoresizingMaskIntoConstraints = false
 
         thumbnailWidthConstraint = thumbnailView.widthAnchor.constraint(
             equalToConstant: StyleConstants.thumbnailSize.width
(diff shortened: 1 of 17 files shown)


PocketKit/Sources/PocketKit/Home/HomeCarouselItemCell.swift
```diff
@@ -25,19 +25,22 @@ class HomeCarouselItemCell: UICollectionViewCell {
     private let titleLabel: UILabel = {
         let label = UILabel()
         label.numberOfLines = Constants.maxTitleLines
+        label.adjustsFontForContentSizeCategory = true
```
jacobsimeon, line 28: Did this turn out to be strictly necessary? I would imagine that the font size is scaled when building the text attributes from the style here:
https://github.com/Pocket/pocket-ios/blob/d00f21443d7b4a6f97967d5b223743e06fa4ac8c/PocketKit/Sources/Textile/Style/Style%2BUIKit.swift#L141

PocketKit/Sources/PocketKit/Style/StyleConstants.swift
```diff
@@ -8,6 +8,10 @@ enum Margins: CGFloat {
   case wide = 32
 }
 
-enum StyleConstants {
+struct StyleConstants {
```
jacobsimeon, line 11: I actually like having this be an enum since it mainly serves as a container for a few static properties. Making it a struct would allow creating instances of `StyleConstants` which could get confusing since it has no instance properties or methods.

PocketKit/Sources/Textile/Style/Style.swift
```diff
@@ -63,6 +66,18 @@ public struct Style {
             backgroundColor: backgroundColorAsset
         )
     }
+    
+    public func with(maxScaleSize: CGFloat) -> Style {
```
jacobsimeon, line 70: Nice 👍

PocketKit/Sources/PocketKit/MyList/MyListTitleView.swift
```diff
@@ -88,18 +78,19 @@ class MyListTitleView: UIView {
...
 
     override func layoutSubviews() {
         super.layoutSubviews()
-
-        if selection.frame == .zero,
-           let selectedButton = buttons.first(where: \.isSelected) {
-            selection.frame = selectedButton.frame
+        
+        buttons.forEach { button in
+            button.layer.cornerRadius = 16
```
jacobsimeon, line 87: Digging through view hierarchies like this can lead to unexpected weirdness down the line if the hierarchy changes (i.e. if we add another kind of button).

I'm thinking we can configure these buttons to update the background color when their state changes by setting the `backgroundColorTransformer` on the background configuration.

```swift
class MyListSelectionButton: UIButton {
    // ...
    init(selection: MylistSelection, action: UIAction) {
        // ...
        configuration = UIButton.Configuration.plain()
        configuration?.background.backgroundColorTransformer = UIConfigurationColorTransformer { [weak self] _ in
            return isSelected ? UIColor(.ui.teal6) : .clear
        }
    }
}
```

jacobsimeon (commented):
Looking great. Just a couple of questions/thoughts for ya

jacobsimeon:
PS: I use the mini-sized phone and I have the dynamic text size set to a couple notches larger than the standard size so my eyes thank you for this one!

PocketKit/Sources/PocketKit/Home/HomeCarouselItemCell.swift
cyndichin: The fonts wouldn't scale if I background the app, changed the font size in Settings and then foreground the app. It does scale after closing and re-opening the app. Adding this change allows the fonts to adjust after the user navigates back to the app, without having to relaunch it.

PocketKit/Sources/PocketKit/Style/StyleConstants.swift
cyndichin: updated!

PocketKit/Sources/PocketKit/MyList/MyListTitleView.swift
cyndichin: good point, updated!

jacobsimeon (approved):

PR #219, opened 2022-08-08
IN-661: Save From Clipboard

cyndichin:
## Summary
Support Save From Clipboard by Adding a Banner that allows user to save an item after copying a URL

## References 
IN-661, IN-662, IN-664

## Implementation Details
Add `BannerView` to host the main UI element for the banner and the data to populate the UI elements are handled by `SaveFromClipboardViewModel`. Since we want this banner to appear in various places in the app, the logic to handle animation and presentation is located in `RootCoordinator`.

## Test Steps
- [x] Given a user has a URL copied to their clipboard,
and the url is not already saved,
and the url is not the last URL presented,
when the app is launched,
then I should see a banner asking to save the URL.

- [x] Given a user has a URL copied to their clipboard,
and the url is not already saved,
and the url is the last URL presented,
when the app is launched,
then I should see the banner asking to save the URL.

- [x] Given a user has a URL copied to their clipboard,
and the url is already saved,
when the app is launched,
then I should see the banner asking to save the URL.

- [x] Given there is a url on the clipboard,
and the clipboard banner has been presented,
when the “Save” button is tapped,
the URL should be saved.

- [x] Given there is a url on the clipboard,
and the clipboard banner has been presented,
when the “Save” button is tapped,
the URL should appear in the user’s list.

- [x] Given there is a url on the clipboard,
and the clipboard banner has been presented,
when the “Save” button is tapped,
and the URL save has been synced with the server,
the list should replace the url with the saved item.

- [x] Given the clipboard banner is present,
when the user swipes down on the banner,
the banner should dismiss.

Note: For now, it is okay for the user to continue to see the banner. In the case of the banner, if the user already saved the item - then it will bring the item to the top. 

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

## Screenshots

https://user-images.githubusercontent.com/6743397/183942732-eda575bc-4fac-471c-b961-a369e02bcf71.mp4

### DIFF
--- a/PocketKit/Sources/PocketKit/Banner/SavedFromClipboardViewModel.swift
+++ b/PocketKit/Sources/PocketKit/Banner/SavedFromClipboardViewModel.swift
@@ -1,0 +1,54 @@
+import Foundation
+import Sync
+import Textile
+import UIKit
+
+
+protocol SavedFromClipboardViewModelDelegate: AnyObject {
+    func coordinatorDismissBanner()
+}
+
+class SavedFromClipboardViewModel: BannerViewModel {
+    let prompt = "Add copied URL to your Saves?"
+    let clipboardURL: String?
+    let buttonText: String? = "Save"
+    
+    private let source: Source?
+    weak var delegate: SavedFromClipboardViewModelDelegate?
+    
+    init(clipboardURL: String?, source: Source? = nil) {
+        self.clipboardURL = clipboardURL
+        self.source = source
+    }
+    
+    func action() {
+        guard let clipboardURL = clipboardURL, let url =  URL(string: clipboardURL) else { return }
+        source?.save(url: url)
+        delegate?.coordinatorDismissBanner()
+    }
+    
+    var attributedText: NSAttributedString {
+        return NSAttributedString(string: prompt, style: .main)
+    }
+    
+    var attributedDetailText: NSAttributedString {
+        return NSAttributedString(string: clipboardURL ?? "", style: .detail)
+    }
+    
+    var attributedButtonText: NSAttributedString {
+        return NSAttributedString(string: buttonText ?? "", style: .button)
+    }
+    
+    var backgroundColor: UIColor = UIColor(.ui.teal6)
+    var borderColor: UIColor = UIColor(.ui.teal5)
+}
+
+private extension Style {
+    static let main: Self = .header.sansSerif.p2.with(weight: .semibold).with { paragraph in
+        paragraph.with(lineSpacing: 4)
+    }
+    static let detail: Self = .header.sansSerif.p4.with { paragraph in
+        paragraph.with(lineBreakMode: .byTruncatingTail)
+    }.with(maxScaleSize: 16)
+    static let button: Self = .header.sansSerif.h8.with(color: .ui.white1)
+}
--- a/PocketKit/Sources/PocketKit/Main/MainCoordinator.swift
+++ b/PocketKit/Sources/PocketKit/Main/MainCoordinator.swift
@@ -9,6 +9,10 @@ class MainCoordinator {
     var viewController: UIViewController {
         regular.viewController
     }
+    
+    var compactViewController: UIViewController {
+        compact.viewController
+    }
 
     private let compact: CompactMainCoordinator
     private let regular: RegularMainCoordinator
--- a/PocketKit/Sources/PocketKit/Main/MainCoordinator.swift
+++ b/PocketKit/Sources/PocketKit/Main/MainCoordinator.swift
@@ -10,7 +10,7 @@ class MainCoordinator {
         regular.viewController
     }
     
-    var tabBar: UITabBar {
+    var tabBar: UITabBar? {
         compact.tabBar
     }
 
--- a/PocketKit/Sources/PocketKit/Main/MainCoordinator.swift
+++ b/PocketKit/Sources/PocketKit/Main/MainCoordinator.swift
@@ -10,8 +10,8 @@ class MainCoordinator {
         regular.viewController
     }
     
-    var compactViewController: UIViewController {
-        compact.viewController
+    var tabBar: UITabBar {
+        compact.tabBar
     }
 
     private let compact: CompactMainCoordinator
--- a/PocketKit/Sources/PocketKit/Root/RootCoordinator.swift
+++ b/PocketKit/Sources/PocketKit/Root/RootCoordinator.swift
@@ -59,7 +59,7 @@ class RootCoordinator {
         window?.addSubview(bannerView)
         configureConstraints(bannerView)
         
-        let tabBarHeight: CGFloat = main?.tabBar.frame.height ?? 0
+        let tabBarHeight: CGFloat = main?.tabBar?.frame.height ?? 0
         let translation = window?.traitCollection.userInterfaceIdiom == .phone ? tabBarHeight : 0
       
         UIView.animate(withDuration: 0.5, delay: 0, options: [.allowUserInteraction], animations: {
@@ -84,9 +84,9 @@ class RootCoordinator {
                 bannerView.centerXAnchor.constraint(equalTo: window.centerXAnchor),
             ])
         } else {
-            let tabBar = main?.tabBar ?? window
+            let bottomAnchor = main?.tabBar?.bottomAnchor ?? window.safeAreaLayoutGuide.bottomAnchor
             NSLayoutConstraint.activate([
-                bannerView.bottomAnchor.constraint(equalTo: tabBar.bottomAnchor, constant: 0),
+                bannerView.bottomAnchor.constraint(equalTo: bottomAnchor, constant: 0),
                 bannerView.leadingAnchor.constraint(equalTo: window.leadingAnchor, constant: 12),
                 bannerView.trailingAnchor.constraint(equalTo: window.trailingAnchor, constant: -12),
             ])
--- a/PocketKit/Sources/PocketKit/Root/RootCoordinator.swift
+++ b/PocketKit/Sources/PocketKit/Root/RootCoordinator.swift
@@ -14,6 +14,7 @@ class RootCoordinator {
     private var shouldAnimateTransition = false
 
     private var subscriptions: [AnyCancellable] = []
+    private var bannerView: UIView? = nil
 
     private let rootViewModel: RootViewModel
     private let mainCoordinatorFactory: () -> MainCoordinator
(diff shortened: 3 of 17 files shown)


Tests iOS/MyList/ArchiveFiltersTests.swift
```diff
@@ -41,6 +41,7 @@ class ArchiveFiltersTests: XCTestCase {
         }
 
         try server.start()
+        UIPasteboard.general.string = ""
```
cyndichin, line 44: it seems the banner was appearing because there was a copy of `https://getpocket.com/` in the clipboard, so I nil out the pasteboard so it doesn't break these tests. open to better ideas

jacobsimeon: The pasteboard containing a URL is almost certainly due to another test that is messing with the pasteboard. I took a look around some other tests and it could be [RecommendationViewModelTests](https://github.com/Pocket/pocket-ios/blob/3b3cb4b846660ae73b89ed3640dfa8df578ffaa1/PocketKit/Tests/PocketKitTests/RecommendationViewModelTests.swift#L306).

That test should clear out UIPasteboard in its own `tearDown` method so we don't have to do that here.

(not sure if this is the exact test that causing the issue, there could be others)

jacobsimeon (commented):
Looking great! `BannerView.swift` is especially clean.

As discussed via zoom, look into:
- replacing `SavedFromClipboardViewModelDelegate` with observation of a `$isDismissed` property.
- Possibly remove `SavedFromClipboardViewModel` altogether and use a simpler struct that holds onto closures that represent `dismissAction` and `primaryAction`
- Currently the root coordinator has to do a bit of digging to determine the bottom anchor for the banner view, move this responsibility to the main coordinator and/or the compact coordinator.

Tests iOS/MyList/ArchiveFiltersTests.swift
cyndichin: thanks! I cleared out the pasteboard in the teardown for `RecommendationViewModelTests` and `SavedItemViewModelTests` instead

PocketKit/Sources/PocketKit/Banner/SavedFromClipboardViewModel.swift
```diff
@@ -0,0 +1,36 @@
+import Foundation
+import Sync
+import Textile
+import UIKit
+
+struct BannerViewModel {
```
jacobsimeon, line 6: Nice, I like this! (let's rename this file to reflect the model name change)

PocketKit/Sources/PocketKit/Root/RootCoordinator.swift
```diff
@@ -41,6 +42,64 @@ class RootCoordinator {
...
+        rootViewModel.$bannerViewModel.sink { [weak self] viewModel in
+            if let viewModel = viewModel {
+                self?.setupBanner(with: viewModel)
+            } else {
+                self?.dismissView()
+            }
+        }.store(in: &subscriptions)
+
+        NotificationCenter.default.publisher(for: UIApplication.didEnterBackgroundNotification).sink { [weak self] _ in
+            self?.removeBanner()
```
jacobsimeon, line 55: What do you think about moving this into the root view model? We could listen for this same notification and then set `self.bannerViewModel` to nil to dismiss. I think it gives a nice symmetry with the logic that presents the banner (and it also moves this business logic out of the coordinator)

PocketKit/Sources/PocketKit/Root/RootCoordinator.swift
```diff
@@ -41,6 +42,64 @@ class RootCoordinator {
...
+    private func configureConstraints(_ bannerView: UIView) {
+        guard let window = window else { return }
+        bannerView.translatesAutoresizingMaskIntoConstraints = false
+        if window.traitCollection.userInterfaceIdiom == .pad {
+            bannerView.bottomAnchor.constraint(equalTo: window.bottomAnchor, constant: -12).isActive = true
+            bannerView.widthAnchor.constraint(equalToConstant: 600).isActive = true
+            bannerView.centerXAnchor.constraint(equalTo: window.centerXAnchor).isActive = true
+        } else {
+            let tabBar = main?.tabBar ?? window
+            bannerView.bottomAnchor.constraint(equalTo: tabBar.bottomAnchor, constant: 0).isActive = true
```
jacobsimeon, line 90: It can be cumbersome to have to set `isActive = true` on all these constraints one by one. As an alternative you can go:

```swift
NSLayoutConstraint.activate([
  bannerView.leadingAnchor.constraint(equalTo: ...),
  bannerView.trailingAnchor.constraint(equalTo: ...)
  ...
])
```

Not saying one is better than the other, but I personally prefer the above style

PocketKit/Tests/PocketKitTests/RecommendationViewModelTests.swift
```diff
@@ -23,6 +23,7 @@ class RecommendationViewModelTests: XCTestCase {
     }
 
     override func tearDownWithError() throws {
+        UIPasteboard.general.string = ""
```
jacobsimeon, line 26: nice

Tests iOS/BannerViewTests.swift
```diff
@@ -0,0 +1,221 @@
+import XCTest
+import Sails
+import NIO
+
+
+class BannerViewTests: XCTestCase {
```
jacobsimeon, line 6: First off, I love these tests! So happy to see that setting the pasteboard content is so simple.

That said, I think it's a bit overkill to have a separate pair of tests for each tab. I get that we want to make sure that the banner appears (or doesn't appear) on each tab, but we need to way the cost of the tests against the value they provide.

The "cost" of the tests in this case is the time it takes for them to run. UI tests that require launching the app take quite a while to run (relative to unit tests that are typically very fast). 

I think we could reduce this cost if we simplify down to three tests:
1. save from clipboard banner appears on the home tab when foregrounding the app w/ a URL in the pasteboard
2. same as 1 but with a URL that has already been saved (assert the item is moved to top of list) 
3. save from clipboard banner _does not appear_ when foregrounding the app without a URL in the pasteboard

In tests 1 and 3, you could switch between each tab to make sure the banner is still present (or not as the case may be).

If we organize the tests in that way, then I think we can get the same general value but at a much lower cost of running. What do you think?

Tests iOS/Support/Response+factories.swift
```diff
@@ -31,8 +31,11 @@ extension Response {
         fixture(named: "slate-detail-\(number)")
     }
 
-    static func saveItem(_ fixtureName: String = "save-item") -> Response {
-        fixture(named: fixtureName)
+    static func saveItem(_ fixtureName: String = "save-item", number: Int? = nil) -> Response {
```
jacobsimeon, line 34: I see what you're going for here but I think this API could get a bit confusing. If I pass a number, it's going to add it to a default fixture name, but if I pass a number and a fixture name then it will append the number to the fixture name I just passed. In the latter case, I would probably opt to pass the whole fixture name, e.g.: `saveItem("save-item-2")`

What do you think about taking out the `number: Int` param and just passing in "save-item-2" as the fixture name in your tests?

jacobsimeon (changes requested):
Looking great!

I just have a couple more thoughts for ya (take them or leave them) and one requested change to rename a file

PocketKit/Sources/PocketKit/Main/MainCoordinator.swift
```diff
@@ -9,6 +9,10 @@ class MainCoordinator {
     var viewController: UIViewController {
         regular.viewController
     }
+    
+    var tabBar: UITabBar? {
+        (compact.viewController as? UITabBarController)?.tabBar
```
jacobsimeon, line 14: I don't love that we have to downcast to `UITabBarController` here. The fact that the compact view uses a tab bar controller is meant to be an implementation detail that other coordinators don't know/care about.

How about we go one level deeper and add this same property to `CompactMainCoordinator`?
```swift
class MainCoordinator {
    var tabBar: UITabBar {
        compact.tabBar
    }
}

class CompactMainCoordinator {
    // ...
    var tabBar: UITabBar {
        tabBarController.tabBar
    }
}
```

In fact, we could further hide implementation details of the compact coordinator if we changed this to simply return a view that we can anchor to when setting up constraints for the banner:

```swift
class MainCoordinator {
    var bannerAnchor: UIView {
        compact.bannerAnchor
    }
}

class CompactMainCoordinator {
    // ...
    var bannerAnchor: UIView {
        tabBarController.tabBar
    }
}
```

I'll leave this one up to you. I just see this downcast and a little warning buzzer goes off in my head but I can happily ignore it in this case if you can. 😄

Tests iOS/BannerViewTests.swift
cyndichin: I agree! What do we think about keeping the dismissing of the banner as test 4?

PocketKit/Sources/PocketKit/Root/RootCoordinator.swift
```diff
@@ -41,6 +42,64 @@ class RootCoordinator {
...
+            bannerView.trailingAnchor.constraint(equalTo: window.trailingAnchor, constant: -12).isActive = true
+        }
+    }
+    
+    func dismissView() {
+        let animationDistance = self.window?.frame.height
+        UIView.animate(withDuration: 1.0, delay: 0, options: [.allowUserInteraction], animations: {
+            self.bannerView?.transform = CGAffineTransform(translationX: 0, y: animationDistance ?? CGFloat.greatestFiniteMagnitude)
+        }, completion: {_ in
+            self.removeBanner()
```
ckpocket, line 101: Is [weak self] needed for to avoid holding a strong reference of self?

Tests iOS/Support/Response+factories.swift
cyndichin: makes sense! will update

PocketKit/Sources/PocketKit/Main/MainCoordinator.swift
cyndichin: sounds good to me :) I went with the first one because we also need to use the tabBar to get the height, so not only using it for the anchor constraints

PocketKit/Sources/PocketKit/Root/RootCoordinator.swift
cyndichin: good catch!

cyndichin: good point, updated!

Tests iOS/BannerViewTests.swift
jacobsimeon: oh, right. Yes let's keep that one as well

jacobsimeon (approved):

jacobsimeon:
Looks great! Approved (pending changes to minimize the number of UI tests)

ckpocket (approved):

audie1:
Crash has been resolved.

@cyndichin the only comment I have left is about the Save button in Dark Mode:
https://pocket.slack.com/archives/C4DL4QSU8/p1660763981535209

I'm guessing we can address that later if we need to, for now I'll mark as QA Pass on this PR, great work!

PR #220, opened 2022-08-10
IN-672: Add Loading State to Home / Slate Detail View

cyndichin:
## Summary
Add Loading Spinner to Home + Slate Detail

## References 
IN-672, IN-692

## Implementation Details
Check if the snapshot has any sections, if not, then show loading state via `loadingSnapshot`

## Test Steps
- [x] Given I have opened the Pocket app,
when I view the Home tab and data has not yet been loaded,
then I should see a spinner in the middle of the screen.

- [x] Given I viewed the Home tab,
when data has been loaded,
then the spinner should disappear and the Home content should be visible.

- [x] Given I have opened the Pocket app,
when I view the Home tab and tap on a slate topic and data has not yet been loaded,
then I should see a spinner in the middle of the screen.

- [x] Given I viewed the Home tab and viewing a slate topic,
when data has been loaded,
then the spinner should disappear and the Home content should be visible.

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

## Screenshots
Simulator Screen Shot - iPhone 12 mini - 2022-08-10 at 15 01 04

### DIFF
--- a/PocketKit/Sources/PocketKit/Home/HomeViewModel.swift
+++ b/PocketKit/Sources/PocketKit/Home/HomeViewModel.swift
@@ -109,10 +109,7 @@ class HomeViewModel {
         self.tracker = tracker
 
         self.snapshot = {
-            var snapshot = Snapshot()
-            snapshot.appendSections([.loading])
-            snapshot.appendItems([.loading], toSection: .loading)
-            return snapshot
+            return Self.loadingSnapshot()
         }()
 
         NotificationCenter.default.publisher(
@@ -129,7 +126,9 @@ class HomeViewModel {
 
     func fetch() {
         do {
-            snapshot = try rebuildSnapshot()
+            let snapshot = try rebuildSnapshot()
+            guard snapshot.numberOfSections != 0 else { return }
+            self.snapshot = snapshot
         } catch {
             print(error)
         }
@@ -335,6 +334,16 @@ extension HomeViewModel {
         case .loading, .slateCarousel:
             return nil
         }
+    }
+}
+
+// MARK: - Loading Section
+extension HomeViewModel {
+    static func loadingSnapshot() -> Snapshot {
+        var snapshot = Snapshot()
+        snapshot.appendSections([.loading])
+        snapshot.appendItems([.loading], toSection: .loading)
+        return snapshot
     }
 }
 
--- a/PocketKit/Sources/PocketKit/Home/SlateDetailViewModel.swift
+++ b/PocketKit/Sources/PocketKit/Home/SlateDetailViewModel.swift
@@ -49,7 +49,9 @@ class SlateDetailViewModel {
     }
 
     func fetch() {
-        snapshot = buildSnapshot()
+        let snapshot = buildSnapshot()
+        guard snapshot.numberOfItems != 0 else { return }
+        self.snapshot = snapshot
     }
 
     func refresh(_ completion: @escaping () -> Void) {
--- a/PocketKit/Tests/PocketKitTests/HomeViewModelTests.swift
+++ b/PocketKit/Tests/PocketKitTests/HomeViewModelTests.swift
@@ -50,18 +50,18 @@ class HomeViewModelTests: XCTestCase {
         wait(for: [snapshotExpectation], timeout: 1)
     }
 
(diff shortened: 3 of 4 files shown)


PocketKit/Sources/PocketKit/Home/HomeViewModel.swift
```diff
@@ -173,7 +170,11 @@ extension HomeViewModel {
                 )
             }
         }
-
+        
+        if snapshot.numberOfSections == 0 {
+            return Self.buildLoadingSnapshot()
+        }
```
cyndichin, line 176: is there a case where we don't want to show the loading cell if there are no items?

Tests iOS/LoadingTests.swift
```diff
@@ -0,0 +1,27 @@
...
+        let uiApp = XCUIApplication()
+        app = PocketAppElement(app: uiApp)
+    }
+    
+    override func tearDownWithError() throws {
+        app.terminate()
+    }
+    
+    func test_initialHomeState_showsLoadingState() {
+        app.launch().loadingCell.wait()
```
ckpocket, line 25: Sorry, I'm not verifying this on my system, is this test making sure that the loading starts and then stop/hide after certain secs?

PocketKit/Sources/PocketKit/Home/HomeViewModel.swift
```diff
@@ -287,6 +288,16 @@ extension HomeViewModel {
     }
 }
 
+// MARK: - Loading Section
+extension HomeViewModel {
+    static func buildLoadingSnapshot() -> Snapshot {
```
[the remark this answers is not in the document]
jacobsimeon: I believe it's because we were originally calling it from the initializer before all the other properties on `self` were initialized, so we couldn't pull it out to an instance method (since that would reference self before being fully initialized)

jacobsimeon:
David mentioned something about how the loading cell doesn't appear if there are recent saves while we're fetching the slate lineup. In that case, we'd want to show the recents saves _and also_ show the loading spinner. Are we accounting for that as part of this PR?

Tests iOS/LoadingTests.swift
```diff
@@ -0,0 +1,27 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import XCTest
+import Sails
+import NIO
+
+class LoadingTests: XCTestCase {
+    var server: Application!
```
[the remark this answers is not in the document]
cyndichin: thinking about combining IN-692, which may need that so moving PR to draft to avoid confusion

Tests iOS/LoadingTests.swift
cyndichin: just showing loading cell for now if there's nothing loaded

cyndichin: decided to test with unit test instead of UI tests

Tests iOS/LoadingTests.swift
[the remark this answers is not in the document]
cyndichin: decided to test with unit test instead of UI tests

cyndichin:
> David mentioned something about how the loading cell doesn't appear if there are recent saves while we're fetching the slate lineup. In that case, we'd want to show the recents saves _and also_ show the loading spinner. Are we accounting for that as part of this PR?

After discussing with @moizsyed and @nzeltzer, we eventually want to use skeleton cells to handle it and not the loading spinner. For offline view, we will show an offline state in the slate lineup section but no additional requirements will be handled as part of this ticket. https://pocket.slack.com/archives/C01UH57EJKD/p1660254585025399?thread_ts=1660251042.902629&cid=C01UH57EJKD

dskuza (approved):

audie1:
An issue was discussed during testing:
https://pocket.slack.com/archives/C4DL4QSU8/p1660668113519139?thread_ts=1660659876.886519&cid=C4DL4QSU8

This will be addressed elsewhere, the core functionality of this PR is looking good - I tested it on iPad/iPhone, QA Pass

PR #221, opened 2022-08-11
Refactor coordinators and properly unwind navigational view model properties

jacobsimeon:
## Summary
* Refactor coordinators and properly unwind navigational view model properties

## Implementation Details
The goal of our coordinators is twofold:

1. We want to be able to encapsulate all business logic within our
   view models, and simply observe their state to know when to present
   modal views or push views onto a navigation controller.
2. When users transition from compact to regular size classes (or vice
   versa), we can keep our compact view hiearchy in sync with the split
   view's hiearchy.

The theoretical concept is mostly good —containing state within view
models that are shared between both the regular and compact coordinators
allows us to easily reconstruct the view hiearchy when the user
transitions size classes.

However, the practical reality is...less good. One of the complications
with this approach is that we need to clean up after ourselves when
users dismiss modals or pop the navigation stack. For example, when the
share sheet is dismissed, we need to clear the `sharedActivity` property
of the view model that presented it.

To make this "unwinding" easier to keep track of, a set of methods to
clear certain properties have been added to objects throughout the view
model tree. This means the coordinators can call a single method on
their top level view model to "clearPresentedSharedActivity" and the
message will delegate down the tree to ensure that the correct property
is cleared, regardless of what level the shared activity was shared
from.

In addition to this change (and to make it somewhat easier to reason through), `RegularHomeCoordinator` and `RegularMyListCoordinator` have been extracted from `RegularMainCoordinator`. These sub-coordinators are focused on single parts of the app and delegate to higher level coordinator(s) when necessary to present modal content.

https://getpocket.atlassian.net/browse/IN-701

### DIFF
--- a/PocketKit/Sources/PocketKit/Article/ReadableViewModel/ReadableViewModel.swift
+++ b/PocketKit/Sources/PocketKit/Article/ReadableViewModel/ReadableViewModel.swift
@@ -32,6 +32,7 @@ protocol ReadableViewModel: ReadableViewControllerDelegate {
     func showWebReader()
     func fetchDetailsIfNeeded()
     func externalActions(for url: URL) -> [ItemAction]
+    func clearPresentedWebReaderURL()
 }
 
 // MARK: - ReadableViewControllerDelegate
--- a/PocketKit/Sources/PocketKit/Home/CompactHomeCoordinator.swift
+++ b/PocketKit/Sources/PocketKit/Home/CompactHomeCoordinator.swift
@@ -47,24 +47,11 @@ class CompactHomeCoordinator: NSObject {
         isResetting = true
 
         model.$selectedReadableType.sink { [weak self] readableType in
-            switch readableType {
-            case .savedItem(let viewModel):
-                self?.show(viewModel)
-            case .recommendation(let viewModel):
-                self?.show(viewModel)
-            case .none:
-                self?.readerSubscriptions = []
-            }
-        }.store(in: &subscriptions)
-
-        model.$selectedSlateDetailViewModel.sink { [weak self] viewModel in
-            self?.show(viewModel)
+            self?.show(readableType)
         }.store(in: &subscriptions)
 
         model.$selectedRecommendationToReport.sink { [weak self] recommendation in
-            self?.report(recommendation) {
-                self?.model.selectedRecommendationToReport = nil
-            }
+            self?.report(recommendation)
         }.store(in: &subscriptions)
 
         model.$presentedWebReaderURL.sink { [weak self] url in
@@ -79,7 +66,7 @@ class CompactHomeCoordinator: NSObject {
             self?.present(activity: activity)
         }.store(in: &subscriptions)
 
-        model.$tappedSeeAll.dropFirst().sink { [weak self] seeAll in
+        model.$tappedSeeAll.sink { [weak self] seeAll in
             self?.show(seeAll)
         }.store(in: &subscriptions)
 
@@ -97,9 +84,21 @@ class CompactHomeCoordinator: NSObject {
         homeViewController.handleBackgroundRefresh(task: task)
     }
 
+    func show(_ readableType: ReadableType?) {
+        switch readableType {
+        case .savedItem(let viewModel):
+            show(viewModel)
+        case .recommendation(let viewModel):
+            show(viewModel)
+        case .none:
+            readerSubscriptions = []
+        }
+    }
+
     func show(_ viewModel: SlateDetailViewModel?) {
+        slateDetailSubscriptions = []
+
         guard let viewModel = viewModel else {
-            slateDetailSubscriptions = []
             return
         }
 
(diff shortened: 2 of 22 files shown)


PocketKit/Sources/PocketKit/Home/CompactHomeCoordinator.swift
```diff
@@ -48,40 +46,27 @@ class CompactHomeCoordinator: NSObject {
...
-            case .none:
-                self?.readerSubscriptions = []
-            }
+        model.$selectedReadableType.sink { [weak self] readableType in
+            self?.show(readableType)
         }.store(in: &subscriptions)
 
-        model.$selectedSlateDetailViewModel.receive(on: DispatchQueue.main).sink { [weak self] viewModel in
-            self?.show(viewModel)
+        model.$selectedRecommendationToReport.sink { [weak self] recommendation in
```
cyndichin, line 53: why did we decide not to receive these on main?

PocketKit/Sources/PocketKit/MyList/ArchivedItemsList/ArchivedItemsListViewModel.swift
```diff
@@ -432,3 +432,30 @@ extension ArchivedItemsListViewModel {
...
+        sharedActivity = nil
+        selectedItem?.clearSharedActivity()
+    }
+
+    func clearPresentedWebReaderURL() {
+        switch selectedItem {
+        case .readable(let readable):
+            readable?.clearPresentedWebReaderURL()
+        case .webView:
+            selectedItem = nil
```
cyndichin, line 448: why do we do this for archived items, but not for saved?

PocketKit/Sources/PocketKit/Article/ReadableViewModel/ReadableViewModel.swift
```diff
@@ -32,6 +32,7 @@ protocol ReadableViewModel: ReadableViewControllerDelegate {
     func showWebReader()
     func fetchDetailsIfNeeded()
     func externalActions(for url: URL) -> [ItemAction]
+    func clearPresentedWebReaderURL()
```
cyndichin, line 35: why not add `clearIsPresentingReaderSettings` or `clearSharedActivity`

PocketKit/Sources/PocketKit/Home/SlateDetailViewController.swift
```diff
@@ -51,6 +51,7 @@ class SlateDetailViewController: UIViewController {
 
         view.accessibilityIdentifier = "slate-detail"
         navigationItem.title = model.slateName
+        hidesBottomBarWhenPushed = true
```
cyndichin, line 54: what's this change for?

PocketKit/Sources/PocketKit/Main/RegularMainCoordinator.swift
```diff
@@ -44,42 +47,26 @@ class RegularMainCoordinator: NSObject {
...
         
-        myList = MyListContainerViewController(
-            viewControllers: [
-                ItemsListViewController(model: model.myList.savedItemsList),
-                ItemsListViewController(model: model.myList.archivedItemsList)
-            ]
+        myList = RegularMyListCoordinator(model: model.myList)
+        home = RegularHomeCoordinator(model: model.home, tracker: tracker)
+        account = UINavigationController(
+            rootViewController: AccountViewController(model: model.account)
```
cyndichin, line 53: nice!

PocketKit/Sources/PocketKit/Main/RegularMainCoordinator.swift
```diff
@@ -88,163 +75,75 @@ class RegularMainCoordinator: NSObject {
     }
 
     func showInitialView() {
-        model.$isCollapsed.receive(on: DispatchQueue.main).sink { [weak self] isCollapsed in
+        sizeClassObserver = model.$isCollapsed.sink { [weak self] isCollapsed in
```
cyndichin, line 78: curious to see why this way is preferred

PocketKit/Sources/PocketKit/Home/CompactHomeCoordinator.swift
jacobsimeon: The main reason is because we have a property within these coordinators called `isResetting` that we use to determine if we should animate certain transitions.

If we're resetting –i.e. if we're transitioning size classes– then we want to skip animations. If we receive these messages on the main queue, then the `isResetting` property will already be set to `false` by the time these closures are executed. By removing the `receive(on: ...)`, the closures will be executed immediately so we can count on the `isResetting` property to be correct.

jacobsimeon: The second reasons is that it isn't necessary to explicitly jump to the main queue because all of these properties are already being set on the main thread

PocketKit/Sources/PocketKit/MyList/ArchivedItemsList/ArchivedItemsListViewModel.swift
jacobsimeon: yep, that's an oversight on my part. Thanks!

PocketKit/Sources/PocketKit/Article/ReadableViewModel/ReadableViewModel.swift
jacobsimeon: I don't think it was strictly required in the contexts where we need to clear those things (since we would be dealing with concrete types instead of this protocol). But it might make sense to add them here anyway for consistency's sake.

PocketKit/Sources/PocketKit/Home/SlateDetailViewController.swift
jacobsimeon: When we push the slate detail, we want to hide the bottom tab bar. I'm not sure when that stopped happening, but I noticed it while I was testing things out so I added this line back.

PocketKit/Sources/PocketKit/Main/RegularMainCoordinator.swift
jacobsimeon: Well, we had a property called `longSubscriptions` which was of type `[AnyCancellable]`. I realized that we only had one "long subscription" so we didn't need to have an array, and I felt the name `sizeClassObserver` was more meaningful considering what the property was for.

cyndichin (approved):

PR #223, opened 2022-08-16
IN-227: Add Tags to Overflow Menu

cyndichin:
## Summary
Add tags to overflow menu in each list item (in Saves / Archive)

## References 
IN-227, IN-228

## Implementation Details
Created new `addTags` func in `ItemAction' and add this to the array of actions for the overflow menu. There are three areas that this have been added - 1) Saves Item 2) Archived Item 3) Reader Item (Saves only)

## Test Steps
- [ ] Given a user has navigated to their Saves,
and the user taps on the overflow menu button, 
then the user should see an Add Tags option.

- [ ] Given a user has navigated to their Archive list,
and the user taps on the overflow menu button, 
then the user should see an Add Tags option.

- [ ] Given a user has navigated to the Reader View for a Saved Item,
and the user taps on the overflow menu button on the top corner, 
then the user should see an Add Tags option.

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

## Screenshots
image

image


dskuza (approved):
Thanks for introducing more UI tests!

PR #227, opened 2022-08-24
IN-231: Add Tags Screen Text Field Input

cyndichin:
## Summary
Add Tags Screen: user can input a tag into a text field and save it

Note: Small UI updates (i.e. color) will be addressed as soon as designs are ready

## References 
IN-231

## Implementation Details
* Created Add Tags Screen in SwiftUI which uses the `AddTagsViewModel ` that handles tags management for a specific saved item
* Added `Tag` in Core Data Model as a NSOrderedSet (many to many relationship with a SavedItem)
* Updated `list.graphql` with `tags` field and added `replaceSavedItemTags` mutation for updating a saved item's tags
* Created two requests `fetchTags` and `fetchTag(byName name: String)` to be used to update tags locally by first checking if tags already exists
* Updated `SavedItem+RemoteMapping` so that when we call `fetchList` the save item updates retrieves the remote tags and updates the tags data locally
* Updated `PocketSource` so that when we can call `addTags`to send a mutation query to graphQL. This function updates the tags locally with the new data first and then sends the request to the server.
* User can see the Add Tags Screen from a Saved Item in Reader view, My List and Archive

## Test Steps
- [ ] Given a user has tapped on the Add Tags option for a Saved Item,
and the user views the text field (which is attached to the keyboard),
then the user should be able to enter and add tags.

- [ ] Given a user has tapped on the Add Tags option for a Saved Item,
and the user views the text field, 
then the user should be able to see existing tags.

- [ ] Given a user has tapped on the Add Tags option for a Saved Item,
and the user inputs a tag into the text field, 
then the user should be able save it.

- [ ] Given a user has tapped on the Add Tags option for a Saved Item,
and the user taps on a tag, 
then the user should be able to delete it. 

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

## Screenshots
Simulator Screen Shot - iPhone 12 mini - 2022-08-24 at 11 55 20

### DIFF
--- a/PocketKit/Sources/PocketKit/Article/ReadableViewModel/SavedItemViewModel.swift
+++ b/PocketKit/Sources/PocketKit/Article/ReadableViewModel/SavedItemViewModel.swift
@@ -23,7 +23,7 @@ class SavedItemViewModel: ReadableViewModel {
     var presentedWebReaderURL: URL?
     
     @Published
-    var presentAddTags: AddTagsViewModel?
+    var presentedAddTags: AddTagsViewModel?
 
     @Published
     var sharedActivity: PocketActivity?
@@ -164,12 +164,13 @@ extension SavedItemViewModel {
     }
     
     private func showAddTagsView() {
-        presentAddTags = AddTagsViewModel(item: item,
-                                          source: source,
-                                          saveAction:
-                                            { [weak self] in
-            self?.fetchDetailsIfNeeded()
-        })
+        presentedAddTags = AddTagsViewModel(
+            item: item,
+            source: source,
+            saveAction: { [weak self] in
+                self?.fetchDetailsIfNeeded()
+            }
+        )
     }
 
     private func saveExternalURL(_ url: URL) {
--- a/PocketKit/Sources/PocketKit/Article/ReadableViewModel/SavedItemViewModel.swift
+++ b/PocketKit/Sources/PocketKit/Article/ReadableViewModel/SavedItemViewModel.swift
@@ -21,6 +21,9 @@ class SavedItemViewModel: ReadableViewModel {
 
     @Published
     var presentedWebReaderURL: URL?
+    
+    @Published
+    var presentAddTags: AddTagsViewModel?
 
     @Published
     var sharedActivity: PocketActivity?
@@ -161,7 +164,12 @@ extension SavedItemViewModel {
     }
     
     private func showAddTagsView() {
-        // TODO: Show Add Tags View
+        presentAddTags = AddTagsViewModel(item: item,
+                                          source: source,
+                                          saveAction:
+                                            { [weak self] in
+            self?.fetchDetailsIfNeeded()
+        })
     }
 
     private func saveExternalURL(_ url: URL) {
--- a/PocketKit/Sources/PocketKit/MyList/ArchivedItemsList/ArchivedItemsListViewModel.swift
+++ b/PocketKit/Sources/PocketKit/MyList/ArchivedItemsList/ArchivedItemsListViewModel.swift
@@ -24,6 +24,9 @@ class ArchivedItemsListViewModel: ItemsListViewModel {
 
     @Published
     var presentedAlert: PocketAlert?
+    
+    @Published
+    var presentAddTags: AddTagsViewModel?
 
     @Published
     var selectedItem: SelectedItem?
@@ -250,7 +253,12 @@ extension ArchivedItemsListViewModel {
 // MARK: - Add Tags to an item
 extension ArchivedItemsListViewModel {
     private func showAddTagsView(item: SavedItem) {
-        // TODO: Show Add Tags View
+        presentAddTags = AddTagsViewModel(item: item,
+                                          source: source,
+                                          saveAction:
+                                            { [weak self] in
+            self?.fetch()
+        })
     }
 }
 
--- a/PocketKit/Sources/PocketKit/MyList/ArchivedItemsList/ArchivedItemsListViewModel.swift
+++ b/PocketKit/Sources/PocketKit/MyList/ArchivedItemsList/ArchivedItemsListViewModel.swift
@@ -26,7 +26,7 @@ class ArchivedItemsListViewModel: ItemsListViewModel {
     var presentedAlert: PocketAlert?
     
     @Published
-    var presentAddTags: AddTagsViewModel?
+    var presentedAddTags: AddTagsViewModel?
 
     @Published
     var selectedItem: SelectedItem?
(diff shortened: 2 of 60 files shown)


PocketKit/Sources/Sync/PocketSource.swift
```diff
@@ -303,6 +303,33 @@ extension PocketSource {
...
+        guard let remoteID = item.remoteID else {
+            return
+        }
+        
+        tags.compactMap { $0 }.forEach({ tag in
+            let fetchRequest = Requests.fetchTag(byName: tag)
+            fetchRequest.fetchLimit = 1
+            let fetchedTag = try? space.context.fetch(fetchRequest).first ?? Tag(context: space.context)
+            guard let fetchedTag = fetchedTag else { return }
+            fetchedTag.name = tag
```
jacobsimeon, line 317: I could see moving this logic into `Space` by adding a method `fetchOrCreateByName(name: String)`. Would help keep this method here a bit more focused on building up the mutations.

PocketKit/Sources/Sync/RemoteMapping/SavedItem+RemoteMapping.swift
```diff
@@ -34,7 +34,21 @@ extension SavedItem {
...
             return
         }
-
+        
+        if let tags = tags {
+            removeFromTags(tags)
+        }
+        
+        remote.tags?.compactMap { $0 }.forEach({ remoteTag in
+            let fetchRequest = Requests.fetchTag(byName: remoteTag.name)
```
jacobsimeon, line 43: If we extract this logic out of from the method in `Source` (see above comment), we could reuse it here.

PocketKit/Sources/PocketKit/Tags/AddTagsView.swift
```diff
@@ -0,0 +1,109 @@
...
+            .matchedGeometryEffect(id: tag, in: animation)
+    }
+    
+    func getRows(screenWidth: CGFloat) -> [[String]] {
+        var rows: [[String]] = []
+        var currentRow: [String] = []
+        
+        var totalWidth: CGFloat = 0
+        let safeWidth: CGFloat = screenWidth
+        let padding: CGFloat = 20  // 4 horizontal padding in viewbuilder and 6 padding in hstack
```
jacobsimeon, line 81: How about extracting the horizontal padding values used elsewhere into constants so they can be referenced here? would eliminate the need for the comment and also prevent these values from getting out of sync.

PocketKit/Sources/PocketKit/Tags/AddTagsView.swift
```diff
@@ -0,0 +1,109 @@
...
+            .cornerRadius(4)
+            .lineLimit(1)
+            .onTapGesture {
+                viewModel.removeTag(with: tag)
+            }
+            .accessibilityIdentifier("tag")
+            .matchedGeometryEffect(id: tag, in: animation)
+    }
+    
+    func getRows(screenWidth: CGFloat) -> [[String]] {
```
jacobsimeon, line 75: I see what we're going for here– we need items to flow to a new line as we fill the width of the screen. In UIKit Land, you can build a flow layout using `UICollectionView` + `UICollectionViewFlowLayout`. Is there a SwiftUI analog to those UIKit components?

jacobsimeon (commented):
This looks really good! The overall patterns and structure are all 💯 

I added a couple of small comments regarding style, and some suggestions for minor refactors.

PocketKit/Sources/PocketKit/MyList/ArchivedItemsList/ArchivedItemsListViewModel.swift
```diff
@@ -24,6 +24,9 @@ class ArchivedItemsListViewModel: ItemsListViewModel {
 
     @Published
     var presentedAlert: PocketAlert?
+    
+    @Published
+    var presentAddTags: AddTagsViewModel?
```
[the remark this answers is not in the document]
cyndichin: yes!

PocketKit/Sources/PocketKit/Article/ReadableViewModel/SavedItemViewModel.swift
```diff
@@ -161,7 +164,12 @@ extension SavedItemViewModel {
     }
     
     private func showAddTagsView() {
-        // TODO: Show Add Tags View
+        presentAddTags = AddTagsViewModel(item: item,
+                                          source: source,
+                                          saveAction:
+                                            { [weak self] in
+            self?.fetchDetailsIfNeeded()
+        })
```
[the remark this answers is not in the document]
cyndichin: makes sense! I wonder if we can make this into a SwiftLint rule some how in the future (if one doesn't exist already)

PocketKit/Sources/Sync/RemoteMapping/SavedItem+RemoteMapping.swift
cyndichin: okay! from the current ticket I have, I realize I have to rethink this logic since it doesn't delete old tags (not assigned to save items) from core data. I'll postpone that work in a different PR though

PocketKit/Sources/PocketKit/Tags/AddTagsView.swift
cyndichin: will look into it!

PocketKit/Sources/PocketKit/Article/ReadableViewModel/SavedItemViewModel.swift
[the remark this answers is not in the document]
jacobsimeon: I was wondering the same thing. Not sure if it's available as a rule. We could investigate as part of #228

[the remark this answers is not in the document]
cyndichin: oh I found this - https://realm.github.io/SwiftLint/multiline_arguments_brackets.html
seems like it covers it

PocketKit/Sources/PocketKit/Tags/AddTagsView.swift
cyndichin: The closest thing I'll say to collection views are grids. In iOS 15, only LazyH/VGrid are available and they support only regular structure with items of equal rows/columns.

cc: @kybhamilton

PocketKit/Sources/Sync/RemoteMapping/SavedItem+RemoteMapping.swift
cyndichin: I realize that this is an extension on saved item and doesn't have access to `source`.

jacobsimeon (approved):
➕ 🏷️ 🏷️ 🏷️

PR #233, opened 2022-08-29
[IN-668] Braze SDK Integration

bassrock:
## Summary
* Allow us to use Braze for push notifications and in app messages

## References 
* https://getpocket.atlassian.net/browse/IN-668

## Implementation Details
* Push notifications are registered for a user [provisionally](https://www.braze.com/docs/user_guide/message_building_by_channel/push/ios/notification_options/#provisional-push) as soon as the application is launched (not after they login), this will cause a user to not see a modal for push and allow us to push notifications to the notification center quietly. Braze recommends this, because it allows us to start sending push notifications and give a user a taste of what they can get if they explicitly opt in.
   * An update to the Braze SDK in September will allow us to use in-app messages to also convince a user to give Pocket explicit push action after they perform a high-value action. This is called [push primers](https://www.braze.com/docs/user_guide/message_building_by_channel/push/ios/notification_options/#provisional-push)
* There are no tests because this integration fully uses Apple app delegates and a third-party SDK. I am open to ideas on how to provide test coverage here.
* Deep links while available for Braze, are not setup in this PR and will be a future follow up as we figure out what deep links iOS Next will support and those necessary views are built.

Note: That all the keys used will need to change when we move the Pocket app to the production identifier.

## Test Steps

Note if using a Debug (QA) build you need to use the Development Braze Account because that targets the development Apple Push Notification Service (APNS)

### Provisional Push (Notification Center Quiet Pushes)

> iOS 12+ provides the ability to [deliver push notifications](https://www.braze.com/docs/user_guide/message_building_by_channel/push/ios/notification_options/#provisional-push) to a user's notification center and allow them to opt-out of push notifications or explicitly opt in.

As a user that doesn't explicitly opt-in to push notifications
  I can see push notifications delivered quietly to my notification center.

Have the app backgrounded after launching it once after a fresh install, then send a test push notification in Braze. If you look at your notification center you should see the push notification was delivered quietly there.

### Regular Push Notification - Open Web View
Using the Braze account, send a push test notification that links to a web page in the app and contains an image, title, and excerpt.

As a user that explicitly opts-in to push notifications
  I can see push notifications delivered to my lock screen

* Observe that you can see:
    * An Image
    * A title
    * A excerpt
* If you tap the message it should open in the Pocket App within a modal web view

### In App Message

Note: In-App messages uses push notifications during testing. You must explicitly opt-in to push notifications to test In App messages.

As a user that opens Pocket
  I can see various In App prompts when sent from Braze.

## PR Checklist:
- [ ] Added Unit / UI tests
- [x] Basic Self QA
- [x] Added needed secrets to Bitrise

### DIFF
--- a/Config/secrets.xcconfig.example
+++ b/Config/secrets.xcconfig.example
@@ -1,3 +1,5 @@
 POCKET_API_CONSUMER_KEY=test-consumer-key
 #Note: you need the $() in the url because of how xcconfig expands urls.
 SENTRY_DSN=https:/$()/somesentry@testurl.sentry.io/123456
+BRAZE_API_KEY=123123123123-zcsd-132123-csd
+BRAZE_API_ENDPOINT=someendpoint
--- a/Pocket (iOS).entitlements
+++ b/Pocket (iOS).entitlements
@@ -2,6 +2,8 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>aps-environment</key>
+	<string>development</string>
 	<key>com.apple.security.application-groups</key>
 	<array>
 		<string>group.com.ideashower.ReadItLaterProAlphaNeue</string>
--- a/Pocket.xcodeproj/project.pbxproj
+++ b/Pocket.xcodeproj/project.pbxproj
@@ -44,6 +44,14 @@
 		16EE3F5026CEC80900249AF4 /* PullToRefreshTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 16EE3F4F26CEC80900249AF4 /* PullToRefreshTests.swift */; };
 		16FCADBF26CB04AC00906C03 /* ArchiveAnItemTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 16FCADBE26CB04AC00906C03 /* ArchiveAnItemTests.swift */; };
 		3E8283A428B65ECF00C1DC4A /* SortMenuElement.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3E8283A328B65ECF00C1DC4A /* SortMenuElement.swift */; };
+		650FECCE28BA85F300A3CB0C /* NotificationService.swift in Sources */ = {isa = PBXBuildFile; fileRef = 650FECCD28BA85F300A3CB0C /* NotificationService.swift */; };
+		650FECD228BA85F300A3CB0C /* PushNotificationServiceExtension.appex in Embed App Extensions */ = {isa = PBXBuildFile; fileRef = 650FECCB28BA85F300A3CB0C /* PushNotificationServiceExtension.appex */; settings = {ATTRIBUTES = (RemoveHeadersOnCopy, ); }; };
+		65734C2628BA89B900730661 /* BrazeNotificationService in Frameworks */ = {isa = PBXBuildFile; productRef = 65734C2528BA89B900730661 /* BrazeNotificationService */; };
+		65EC272728BA8AD50075E1DF /* UserNotifications.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 65EC272628BA8AD50075E1DF /* UserNotifications.framework */; };
+		65EC272928BA8AD50075E1DF /* UserNotificationsUI.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 65EC272828BA8AD50075E1DF /* UserNotificationsUI.framework */; };
+		65EC272C28BA8AD50075E1DF /* NotificationViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 65EC272B28BA8AD50075E1DF /* NotificationViewController.swift */; };
+		65EC273328BA8AD50075E1DF /* PushNotificationStoryExtension.appex in Embed App Extensions */ = {isa = PBXBuildFile; fileRef = 65EC272528BA8AD50075E1DF /* PushNotificationStoryExtension.appex */; settings = {ATTRIBUTES = (RemoveHeadersOnCopy, ); }; };
+		65EC273928BA8B120075E1DF /* BrazePushStory in Frameworks */ = {isa = PBXBuildFile; productRef = 65EC273828BA8B120075E1DF /* BrazePushStory */; };
 		8A289CD62899BD820030E5E0 /* BannerViewTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8A289CD52899BD820030E5E0 /* BannerViewTests.swift */; };
 		8A3E3DBC2864DB0500483564 /* EmptyStateTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8A3E3DBB2864DB0500483564 /* EmptyStateTests.swift */; };
 		8A3EAA5128AACC9500392057 /* AddTagsItemTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8A3EAA5028AACC9500392057 /* AddTagsItemTests.swift */; };
@@ -77,6 +85,20 @@
 			remoteGlobalIDString = F204A75F27E22EC50010E155;
 			remoteInfo = SaveToPocket;
 		};
+		650FECD028BA85F300A3CB0C /* PBXContainerItemProxy */ = {
+			isa = PBXContainerItemProxy;
+			containerPortal = 166A81A42637406B0015AA1D /* Project object */;
+			proxyType = 1;
+			remoteGlobalIDString = 650FECCA28BA85F300A3CB0C;
+			remoteInfo = PushNotificationServiceExtension;
+		};
+		65EC273128BA8AD50075E1DF /* PBXContainerItemProxy */ = {
+			isa = PBXContainerItemProxy;
+			containerPortal = 166A81A42637406B0015AA1D /* Project object */;
+			proxyType = 1;
+			remoteGlobalIDString = 65EC272428BA8AD50075E1DF;
+			remoteInfo = PushNotificationStoryExtension;
+		};
 /* End PBXContainerItemProxy section */
 
 /* Begin PBXCopyFilesBuildPhase section */
@@ -87,6 +109,8 @@
 			dstSubfolderSpec = 13;
 			files = (
 				166DF27427E9160A00C03BF4 /* SaveToPocket.appex in Embed App Extensions */,
+				65EC273328BA8AD50075E1DF /* PushNotificationStoryExtension.appex in Embed App Extensions */,
+				650FECD228BA85F300A3CB0C /* PushNotificationServiceExtension.appex in Embed App Extensions */,
 			);
 			name = "Embed App Extensions";
 			runOnlyForDeploymentPostprocessing = 0;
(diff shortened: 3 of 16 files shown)


Pocket (iOS).entitlements
```diff
@@ -2,6 +2,8 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>aps-environment</key>
+	<string>development</string>
```
bassrock, line 6: Docs state xcode will auto change this based on provisioning profiles used.

PocketKit/Sources/PocketKit/PocketNotificationService.swift
```diff
@@ -0,0 +1,217 @@
...
+        // Set up the in app message ui
+        let inAppMessageUI = BrazeInAppMessageUI()
+        inAppMessageUI.delegate = self
+        self.braze.inAppMessagePresenter = inAppMessageUI
+
+
+        self.sessionManager.$currentSession.sink { session in
+            if let session = session {
+                DispatchQueue.main.async {
+                    // Braze SDK docs say this needs to be called from the main thread.
```
jacobsimeon, line 72: We can leverage some conveniences from `Combine` to have this called on the main thread, e.g:

```swift
myPublisher.recieve(on: DispatchQueue.main).sink { val in
    // will be executed on main thread
}
```

PocketKit/Sources/PocketKit/PocketNotificationService.swift
```diff
@@ -0,0 +1,217 @@
...
+                    // https://www.braze.com/docs/developer_guide/platform_integration_guides/ios/analytics/setting_user_ids/#assigning-a-user-id
+                    self.braze.changeUser(userId: session.userIdentifier)
```
jacobsimeon, line 74: References to `self` within a closure can be dangerous because they often create a retain cycle. In this case, `self` has a reference to the session manager, which has a reference to a closure, which holds a strong reference to `self`. So the chain goes `self -> sessionManager -> $currentSession -> closure -> self`

To avoid creating this cycle, use the `[weak self]` annotation in the closure params list, e.g:

```swift
fooPublisher.sink { [weak self] val in
    self?.bar() // self is now available as an optional value (all weakly held objects are optional)
}
```

At this point I pretty much automatically use `[weak self]` whenever I need to reference `self` inside a closure. I recommend just having that be your default approach unless there's a good reason to maintain a strong reference to `self`.

Config/secrets.xcconfig.example
```diff
@@ -1 +1,3 @@
 POCKET_API_CONSUMER_KEY=test-consumer-key
+BRAZE_API_KEY=123123123123-zcsd-132123-csd
```
jacobsimeon, line 2: Just double checking that this is not the real key?

PocketKit/Sources/PocketKit/PocketNotificationService.swift
```diff
@@ -0,0 +1,217 @@
...
+       withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void
+     ) {
+         completionHandler([.list, .banner])
+     }
+}
+
+/**
+ Extend the App Delegate so that we can add in the notification methods
+ */
+extension PocketAppDelegate {
```
jacobsimeon, line 163: I like having this code in an extension but would prefer to have it either 1) within the actual `PocketAppDelegate.swift` file or in another file following the naming convention of `PocketAppDelegate+foo.swift`.

Pocket.xcodeproj/project.pbxproj
```diff
@@ -696,6 +861,7 @@
 				CODE_SIGN_IDENTITY = "Apple Distribution: Read It Later, Inc (EX3VH4YFCH)";
 				CODE_SIGN_STYLE = Manual;
 				DEVELOPMENT_TEAM = EX3VH4YFCH;
+				"DEVELOPMENT_TEAM[sdk=iphoneos*]" = EX3VH4YFCH;
```
jacobsimeon, line 864: Is adding the `sdk=iphoneos*` constraint truly necessary here? I feel like having it adds noise/complication to our configuration. Personally, I'd prefer to keep this out unless it's 100% necessary. (same goes for the configuration of the newly added targets)

Pocket.xcodeproj/project.pbxproj
```diff
@@ -43,6 +43,14 @@
...
+		65734C2628BA89B900730661 /* BrazeNotificationService in Frameworks */ = {isa = PBXBuildFile; productRef = 65734C2528BA89B900730661 /* BrazeNotificationService */; };
+		65EC272728BA8AD50075E1DF /* UserNotifications.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 65EC272628BA8AD50075E1DF /* UserNotifications.framework */; };
+		65EC272928BA8AD50075E1DF /* UserNotificationsUI.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 65EC272828BA8AD50075E1DF /* UserNotificationsUI.framework */; };
+		65EC272C28BA8AD50075E1DF /* NotificationViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 65EC272B28BA8AD50075E1DF /* NotificationViewController.swift */; };
+		65EC273328BA8AD50075E1DF /* PushNotifcationStoryExtension.appex in Embed App Extensions */ = {isa = PBXBuildFile; fileRef = 65EC272528BA8AD50075E1DF /* PushNotifcationStoryExtension.appex */; settings = {ATTRIBUTES = (RemoveHeadersOnCopy, ); }; };
```
jacobsimeon, line 52: typo: `PushNotifcationStoryExtension` -> `PushNotificationStoryExtension`

jacobsimeon (changes requested):
This is super cool! Happy to see that braze makes it so easy to support all these different types of notifications. The basic structure of the code looks great, just a few thoughts on general style and patterns:

### Style for function calls and method declarations 

When parameters are on multiple lines I'm seeing a style here that is very reminiscent of Obj-C. While Obj-C will always occupy a special place in my heart 😆, we have adopted something a bit different:

```swift
// method declarations
// new line after opening param
// indent each line by one level
// closing paren and opening brace on their own line

// Preferred
func foo(
    myBar: Bar,
    aBaz: Baz,
    theQuuxParameterWithAReallyLongName: Qux
) {
    // function body goes here
}

// Not preferred
// aligning the colons creates big swaths of whitespace that can disrupt the natural 
// flow when reading the code
func foo(
                                myBar: Bar,
                                 aBaz: Baz,
  theQuuxParameterWithAReallyLongName: Qux
) {
    // function body here
}
```

A similar style is preferred when calling methods:
```swift
foo(
    myBar: bar,
    aBaz: baz,
    theQuuxParameterWithAReallyLongName: qux
)
```

### Use of comments
From the very beginning, our approach to using comments has been to only use them when they explain _why something exists_ as opposed to _what it does_. Comments that do the later can often become stale and/or inaccurate as code behavior changes over time.
(Not saying any of the comments should be removed, but just a general note for you to consider moving forward.)

### Omit explicit `self` when not required
Throughout the code, we generally omit the use of explicit `self` whenever it is not required, for two reason:
1. It just looks a bit nicer 😉 
2. It makes it easier to move refactor from using a method parameter to a property (or vice versa).

Would you mind updating the style to fit the above notes? (fixing whitespace in method calls and definitions should just be a matter of cmd+A, ctrl+I). Also, whenever we're ready to merge this in, I think it would make sense to consolidate some (or all) of the commits so as to avoid having some commits that don't add a complete piece of work to our main branch.

As for suggestions for testing strategies, my idea would be to extract protocols from our usages of the various types provided by the Braze SDK, change all the properties to use those protocols, then inject mock version of those protocols into the various types that you want to test. However, since most of this code is just wiring in third party components, I don't think we'd get enough value out of those tests to make it worth all the additional work/complexity.

Aside from the style stuff, this is really awesome! So happy to see this getting done.

Config/secrets.xcconfig.example
bassrock: no its not. its an example

Pocket.xcodeproj/project.pbxproj
bassrock: this is xcode autoadding stuff. Will verify that removing it doesn't break things and will update.

Pocket.xcodeproj/project.pbxproj
bassrock: Damn good catch. Will need a new app id and certs generated now 😭

bassrock:
> **Testing strategies**
As for suggestions for testing strategies, my idea would be to extract protocols from our usages of the various types provided by the Braze SDK, change all the properties to use those protocols, then inject mock version of those protocols into the various types that you want to test. However, since most of this code is just wiring in third party components, I don't think we'd get enough value out of those tests to make it worth all the additional work/complexity.

@jacobsimeon I agree here, this feels like pretty boiler plate things calling delegates, etc so the juice isn't worth the squeeze here.

> **Use of comments**
From the very beginning, our approach to using comments has been to only use them when they explain why something exists as opposed to what it does. Comments that do the later can often become stale and/or inaccurate as code behavior changes over time.
(Not saying any of the comments should be removed, but just a general note for you to consider moving forward.)

@jacobsimeon not sure I thoroughly agree here. But it is the iOS team's codebase can you provide examples or comment on the ones you think I should remove?

nzeltzer:
I strongly agree with @jacobsimeon that we should avoid comments describing how something works. I wonder if we couldn't relax the prescription against describing _what_ things do, somewhat: those sorts of comments can be extremely useful for navigating a codebase, especially when the names for things are domain-specific.

jacobsimeon:
> I wonder if we couldn't relax the prescription against describing _what_ things do, somewhat

Two notes on that:
1. It sounds good to me 😄 
2. Before adding a comment to describe what something does, consider whether the name of the variable/method/class/whatever could be changed to better express the same thing.

Not trying to say "COMMENTS AREN'T ALLOWED". I just find that they can add noise and sometimes be misleading as behaviors changes over time.

> Can you provide examples or comment on the ones you think I should remove?

I like all the comments explaining the ins and outs of how the Braze SDK behaves and why certain bits of code exist.

Here's an example of a comment that feels a bit excessive to me, since the variable name (and the fact that it's a private property) express the same thing.
```swift
class PocketNotificationService {
    /**
      Subscriptions that this object listens to.
      */
     private var subscriptions: Set<AnyCancellable> = []
}
```

All of the comments in the PR are fine by me, tbh. In fact, you'd probably get a different take from @cyndichin or @dskuza. Personally, I prefer the aesthetic of sparsely commented code, so that is probably (irrationally) influencing my opinion here 😆

PocketKit/Sources/PocketKit/PocketAppDelegate.swift
```diff
@@ -8,14 +8,15 @@ import Textile
 import Analytics
 import BackgroundTasks
 import SharedPocketKit
-
+import BrazeKit
```
jacobsimeon, line 11: Is this import needed? I don't see any direct references to Braze in the file

PocketKit/Sources/PocketKit/PocketNotificationService.swift
```diff
@@ -0,0 +1,170 @@
...
+        // Enable logging of general SDK information (e.g. user changes, etc.)
+        configuration.logger.level = .info
+        configuration.push.appGroup = "group.com.ideashower.ReadItLaterProAlp"
+        braze = Braze(configuration: configuration)
+        super.init()
+        
+        // Set up the in app message ui
+        let inAppMessageUI = BrazeInAppMessageUI()
+        inAppMessageUI.delegate = self
+        self.braze.inAppMessagePresenter = inAppMessageUI
```
jacobsimeon, line 59: omit `self` here

PocketKit/Sources/PocketKit/PocketNotificationService.swift
```diff
@@ -0,0 +1,170 @@
...
+        
+        sessionManager.$currentSession.receive(on: DispatchQueue.main).sink { [weak self] session in
+            if let session = session {
```
jacobsimeon, line 62: I think you'll see that a lot of swift code would prefer a [guard statement][guard] here to bail out of the function if proper conditions aren't met.

e.g.:

```swift
func foo(bar: String) {
  // make sure all our state and/or inputs are valid before proceeding
  guard !bar.isEmpty else { return }
  doSomething(with: bar)
}
```

The effect is that we eliminate "right drift" and also the main thrust of the function's work is a bit more apparent.

[guard]: https://docs.swift.org/swift-book/ReferenceManual/Statements.html#ID524

PocketKit/Sources/PocketKit/PocketNotificationService.swift
```diff
@@ -0,0 +1,170 @@
...
+     */
+    public func register(deviceToken: Data) {
+        //TODO: Also register the token with the Pocket Instant Sync endpoint
+        braze.notifications.register(deviceToken: deviceToken)
+    }
+    
+    /**
+     Proxy function to braze to handle background notifications
+     */
+    public func handleBackgroundNotifcation(didReceiveRemoteNotification userInfo: [AnyHashable : Any],
```
jacobsimeon, line 95: Let's adopt the style where we have a newline after the opening paren for these params (see below)

PocketKit/Sources/PocketKit/PocketNotificationService.swift
```diff
@@ -0,0 +1,170 @@
...
+        
+        // Manually call the completion handler to let the system know
+        // that the background notification is processed.
+        completionHandler(.noData)
+    }
+    
+    /**
+     Proxy function to Braze to handle user facing notifications
+     */
+    public func didRecieveUserNotifcation( _ center: UNUserNotificationCenter,
```
jacobsimeon, line 120: ditto

PocketKit/Sources/PocketKit/PocketNotificationService.swift
```diff
@@ -0,0 +1,170 @@
...
+        let _ = braze.notifications.handleUserNotification(response: response, withCompletionHandler: completionHandler)
+    }
+}
+
+/**
+ Extend the Notification service to support User Notifications
+ */
+extension PocketNotificationService : UNUserNotificationCenterDelegate {
+    
+    func userNotificationCenter(
```
jacobsimeon, line 132: This is the style we're shooting for (note how the first parameter is on its own line)

jacobsimeon (commented):
Looking really sweet!

Just a couple of minor suggestions (take them or leave them) and a final style tweak.

PocketKit/Sources/PocketKit/PocketAppDelegate.swift
bassrock: ah whoops. nope

PocketKit/Sources/PocketKit/Services.swift
```diff
@@ -65,6 +67,12 @@ struct Services {
             source: source
         )
         imageManager.start()
+        
+        notifcationService = PocketNotificationService(
```
jacobsimeon, line 71: oops, another typo: `notifcationService` -> `notificationService`

jacobsimeon (approved):

jacobsimeon:
Thanks @bassrock! This is a huge contribution!

audie1:
I was unable to verify in-app messages but will do so at a later date, could be a config issue - for now this is PR pass based on previous testing from others around in-app messages.

Discussion here:
https://pocket.slack.com/archives/C03B94CN8KD/p1662658239760329?thread_ts=1662654220.796329&cid=C03B94CN8KD

👍 good to go!

PR #239, opened 2022-09-02
chore(sentry): adding sentry enhancements [IN-817]

bassrock:
## Summary
* Update sentry implementation to always init. This is ok because if sentry has an invalid DSN or none, it will auto disable itself.
* Upgrade the sentry SDK to the latest version
* Only enable sentry debug mode and logging in a development/debug build
* Add sentry to the save extension
* Disabling bitcode since apple is retiring it and it makes uploading symbols much more difficult.
* Switching to deploy pipelines for parallelization (future test improvements ;) )

## References 
* https://getpocket.atlassian.net/browse/IN-817

## Implementation Details

## Test Steps

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/Pocket.xcodeproj/project.pbxproj
+++ b/Pocket.xcodeproj/project.pbxproj
@@ -703,6 +703,7 @@
 				CODE_SIGN_IDENTITY = "Apple Development: David Skuza (A7H8F4CPQQ)";
 				CODE_SIGN_STYLE = Manual;
 				DEVELOPMENT_TEAM = EX3VH4YFCH;
+				ENABLE_BITCODE = NO;
 				ENABLE_PREVIEWS = YES;
 				INFOPLIST_FILE = Info.plist;
 				IPHONEOS_DEPLOYMENT_TARGET = 15.0;
@@ -731,6 +732,7 @@
 				CODE_SIGN_IDENTITY = "Apple Distribution: Read It Later, Inc (EX3VH4YFCH)";
 				CODE_SIGN_STYLE = Manual;
 				DEVELOPMENT_TEAM = EX3VH4YFCH;
+				ENABLE_BITCODE = NO;
 				ENABLE_PREVIEWS = YES;
 				INFOPLIST_FILE = Info.plist;
 				IPHONEOS_DEPLOYMENT_TARGET = 15.0;
@@ -865,6 +867,7 @@
 				CODE_SIGN_IDENTITY = "Apple Development: David Skuza (A7H8F4CPQQ)";
 				CODE_SIGN_STYLE = Manual;
 				DEVELOPMENT_TEAM = EX3VH4YFCH;
+				ENABLE_BITCODE = NO;
 				ENABLE_PREVIEWS = YES;
 				INFOPLIST_FILE = Info.plist;
 				IPHONEOS_DEPLOYMENT_TARGET = 15.0;
@@ -912,6 +915,7 @@
 				CODE_SIGN_IDENTITY = "Apple Development: David Skuza (A7H8F4CPQQ)";
 				CODE_SIGN_STYLE = Manual;
 				DEVELOPMENT_TEAM = EX3VH4YFCH;
+				ENABLE_BITCODE = NO;
 				GENERATE_INFOPLIST_FILE = YES;
 				INFOPLIST_FILE = "$(TARGET_NAME)/Info.plist";
 				INFOPLIST_KEY_CFBundleDisplayName = "Save to Pocket";
@@ -943,6 +947,7 @@
 				CODE_SIGN_IDENTITY = "Apple Development: David Skuza (A7H8F4CPQQ)";
 				CODE_SIGN_STYLE = Manual;
 				DEVELOPMENT_TEAM = EX3VH4YFCH;
+				ENABLE_BITCODE = NO;
 				GENERATE_INFOPLIST_FILE = YES;
 				INFOPLIST_FILE = "$(TARGET_NAME)/Info.plist";
 				INFOPLIST_KEY_CFBundleDisplayName = "Save to Pocket";
@@ -974,6 +979,7 @@
 				CODE_SIGN_IDENTITY = "Apple Distribution: Read It Later, Inc (EX3VH4YFCH)";
 				CODE_SIGN_STYLE = Manual;
 				DEVELOPMENT_TEAM = EX3VH4YFCH;
+				ENABLE_BITCODE = NO;
 				GENERATE_INFOPLIST_FILE = YES;
 				INFOPLIST_FILE = "$(TARGET_NAME)/Info.plist";
 				INFOPLIST_KEY_CFBundleDisplayName = "Save to Pocket";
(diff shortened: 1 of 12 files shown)


PocketKit/Sources/PocketKit/PocketAppDelegate.swift
```diff
@@ -30,9 +30,7 @@ public class PocketAppDelegate: UIResponder, UIApplicationDelegate {
     }
 
     public func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey : Any]? = nil) -> Bool {
-        if !CommandLine.arguments.contains("disableSentry") {
-            Crashlogger.start(dsn: Keys.shared.sentryDSN)
-        }
+        Crashlogger.start(dsn: Keys.shared.sentryDSN)
```
jacobsimeon, line 33: I remember adding this in to avoid sending crash reports while running UI tests and during development. Also, IIRC, the stack traces were sometimes clobbered by sentry during development. I see you've added the conditional compilation bit to set the `debug` option when starting up Sentry. Does that achieve these same results?

PocketKit/Sources/SaveToPocketKit/Services.swift
```diff
@@ -10,11 +10,24 @@ struct Services {
...
+        
+        appSession.$currentSession.sink { session in
+            if let session = session {
+                Crashlogger.setUserID(session.userIdentifier)
+            } else {
+                Crashlogger.clearUser()
+            }
+        }.store(in: &subscriptions)
+        
+
```
jacobsimeon, line 30: The purpose of this `Services` object is to setup and keep ahold of various service objects that need to be used/injected throughout the rest of the app. I think this type of observation code would fit better in one of the view models, probably the `MainViewModel`.

PocketKit/Sources/PocketKit/PocketAppDelegate.swift
bassrock: Ah gotcha, cool I can handle those use cases. We can leverage setting sentry environments here and some other things. I'll look into that.

Pocket.xcodeproj/project.pbxproj
```diff
@@ -699,6 +699,7 @@
 				CODE_SIGN_IDENTITY = "Apple Development: David Skuza (A7H8F4CPQQ)";
 				CODE_SIGN_STYLE = Manual;
 				DEVELOPMENT_TEAM = EX3VH4YFCH;
+				ENABLE_BITCODE = NO;
```
bassrock, line 702: Disabling bitcode. Yay!

PocketKit/Sources/Sync/Crashlogger.swift
```diff
@@ -22,10 +22,25 @@ public struct Crashlogger {
...
+        if isRunningTests() {
+            // We are in a test environment, lets not init sentry.
+            return
+        }
+
         SentrySDK.start { options in
             options.dsn = dsn
-            options.enableAutoPerformanceTracking = false
+            options.enableAutoSessionTracking = true
+            #if DEBUG
```
bassrock, line 33: Enabling debug mode during our debug builds to help see whats going on.

PocketKit/Sources/Sync/Crashlogger.swift
```diff
@@ -22,10 +22,25 @@ public struct Crashlogger {
...
             options.debug = true
+            #endif
         }
     }
+
+    /**
+     Utility to determine if we are in a test environment.
+     */
+    static func isRunningTests() -> Bool {
```
bassrock, line 42: Helper for determining if we are in a test environment. Putting here since it's only used for Sentry, but it may have a better home elsewhere.

bitrise.yml
```diff
@@ -7,12 +7,46 @@ trigger_map:
     workflow: app-store-build
   - pull_request_target_branch: develop
     workflow: qa-build
+  - pull_request_target_branch: develop
```
bassrock, line 10: This splits up the workflow so that QA Build is no longer tied to our tests. This should make bitrise builds move faster by separating the flows.

bitrise.yml
```diff
@@ -7,12 +7,46 @@ trigger_map:
...
+    workflow: test
 workflows:
-  app-store-build:
+  #Reusable workflow that is ran before each workflow to setup the necessary information
+  _setup:
```
bassrock, line 14: Creating a re-usable setup step for the workflows that can run before each workflow.

bitrise.yml
```diff
@@ -53,29 +89,16 @@ workflows:
             - configuration: Debug
       - deploy-to-bitrise-io@2: {}
     before_run:
-      - test
+      - _setup
+    after_run:
+      - _upload_sentry
```
bassrock, line 94: Before the run running setup and after, running upload to sentry!

Pocket.xcodeproj/project.pbxproj
dskuza: Is this valid _now_, or do we have to wait until it's fully deprecated in Xcode 14?

PocketKit/Sources/PocketKit/PocketAppDelegate.swift
dskuza: Re-opening. Perhaps it's personal preference, but I did like having Sentry entirely disabled when doing simulator development, for example. I don't feel there's a need to send things up when I'm doing local debugging.

PocketKit/Sources/Sync/Crashlogger.swift
```diff
@@ -22,10 +22,25 @@ public struct Crashlogger {
...
     public static func start(dsn: String) {
+        if isRunningTests() {
+            // We are in a test environment, lets not init sentry.
+            return
+        }
+
         SentrySDK.start { options in
             options.dsn = dsn
-            options.enableAutoPerformanceTracking = false
+            options.enableAutoSessionTracking = true
```
dskuza, line 32: What kinds of things does this track?

Pocket.xcodeproj/project.pbxproj
bassrock: valid now. Only tvOS and watchOS required it in Xcode 13

PocketKit/Sources/PocketKit/PocketAppDelegate.swift
bassrock: @dskuza If there is no valid DSN in your secrets file it will not initialize.

dskuza: I see, so then the suggestion is to just remove it from our local config and really only keep it present when generating a release build?

bassrock: @dskuza correct!

PocketKit/Sources/Sync/Crashlogger.swift
bassrock: @dskuza Release health https://docs.sentry.io/product/releases/health/

dskuza (approved):
I'm still not _totally_ sold on the SentryDSN environment variable change, but after discussion, I at least understand the thinking. TL;DR: Instead of ticking a checkbox to enable / disable Sentry (which could be a committed change, and have to be captured at code review time), we modify the keys to contain a(n) (in)valid DSN. 

Thanks, @bassrock, for walking me through the bitrise changes. Parallelization / reuse seems to be the major changes there. Excited to see it in action.

PR #245, opened 2022-09-08
Hide chevron in "My List" button

jacobsimeon:
## Summary
Hide the chevron for "My List" button by giving the corresponding view model a `buttonImage` property that can be passed in.

Also:
- A fix for when an edge case where the sidebar is already showing Settings or My List, but the user taps the "My List" button on Home. To handle this case, we explicitly set the navigation stack to what it should be (as opposed to simply pushing the next view controller).
- Better handling for selection status in the sidebar. We reload each cell whenever the selected section changes so that the correct cell is highlighted.

## References 
https://getpocket.atlassian.net/browse/IN-760

### DIFF
--- a/PocketKit/Sources/PocketKit/Home/HomeCollectionViewSectionProvider.swift
+++ b/PocketKit/Sources/PocketKit/Home/HomeCollectionViewSectionProvider.swift
@@ -62,15 +62,15 @@ class HomeViewControllerSectionProvider {
         )
         group.interItemSpacing = .fixed(16)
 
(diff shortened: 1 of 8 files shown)


dskuza (approved):

audie1:
Crash is fixed! One issue found during testing, discussion here:
https://pocket.slack.com/archives/CBKLKAUTC/p1663177862652009?thread_ts=1663173079.434479&cid=CBKLKAUTC

QA Pass 👍

PR #252, opened 2022-09-10
fix(fastlane): switching to recomended flow

bassrock:
## Summary
* Switching to fastlane maintainers recommended solution for internal testflight builds, looks like I originally had it wrong.

https://github.com/fastlane/fastlane/pull/19735#issuecomment-1028398066


nzeltzer (approved):
lgtm

PR #254, opened 2022-09-13
chore(dependencies): updating dependenices to some of the latest versions

bassrock:
## Summary
* Update apollo ios to 0.53.0
* Update Kingfisher from a custom branch to 7.3.2
* Update Sentry to 7.25.0
* Update lottie to 3.4.3  
    * I also enabled CoreAnimation rendering which is in the latest release
* Update Snowplow to 3.2.0
    * Note: I could not update us to 4.0.0 due to an exception crash on launch. Instead of debugging going with 3.2.0 

## Test Steps
* General regression test

## PR Checklist:
- [ ] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/PocketKit/Main/MainCoordinator.swift
+++ b/PocketKit/Sources/PocketKit/Main/MainCoordinator.swift
@@ -29,7 +29,7 @@ class MainCoordinator {
         regular.setCompactViewController(compact.viewController)
 
         model.$selectedSection
-            .sink { section in
+            .receive(on: DispatchQueue.global(priority: .background)).sink { section in
                 let context: UIContext
                 switch section {
                 case .home:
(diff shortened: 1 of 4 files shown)


PocketKit/Sources/PocketKit/Main/MainCoordinator.swift
```diff
@@ -29,7 +29,7 @@ class MainCoordinator {
         regular.setCompactViewController(compact.viewController)
 
         model.$selectedSection
-            .sink { section in
+            .receive(on: DispatchQueue.global(priority: .background)).sink { section in
```
bassrock, line 32: This performs an analytic action and was receiving it on the main thread. This should be in a background thread.

dskuza: @bassrock I understand making this explicit, but does the underlying tracker (Snowplow) automatically do this on a background thread? I feel like the answer is "maybe". Did something show up in Xcode to suggest this change?

bassrock: @dskuza Good call just looked, it does. I reverted this.

dskuza (approved):

audie1:
Did basic manual regression checks on iPad/iPhone, no issues found, good to go 👍

PR #255, opened 2022-09-14
fix(braze): ensure that braze data is wiped and disabled on logout

bassrock:
## Summary

We need to ensure that braze is disabled and user data is wiped when there is no user session active.

## References 
This calls https://braze-inc.github.io/braze-swift-sdk/documentation/brazekit/braze/wipedata() when the session is null.

## Implementation Details
* Add a call to wipe data in braze

Note: This will mean that we can not do push notifications or In App messages unless a user is logged in.

## Test Steps
* Log out of Pocket
* Ensure that you can not send a push notification to the logged out device

## PR Checklist:
- [ ] Added Unit / UI tests
- [x] Basic Self QA


dskuza (approved):

audie1:
Copying test steps from Android:

audie1:
Login to a user
    grab the braze notification ID
    test a push notification
    notification should be received
    log out
    test the same push notification
    no notification should be received
    log back in
    test the same notification
    notification should be received

bassrock:
@audie1 I have made the necessary changes and tested them locally and it seems to be working. 

Note that after the logout and log back in, you may need to force quit the app to trigger apple to give the device a new push notification token to register with braze. As far as I can tell this is an apple quirk when we unregister for push notifications on logout.

audie1:
Looks like the current build is having trouble registering the user with Braze after login to confirm my test steps:

1. Install build 103
2. Login to build 103 via a80 123456
3. Create test Braze campaign
4. Send test message to eW0A0g33d0G6fT454epd6d1p8fTfd07729fw24o76aO223a44e676C58J09umfY4

Discussion thread here:
https://pocket.slack.com/archives/C4DL4QSU8/p1663612145570969?thread_ts=1663611023.455499&cid=C4DL4QSU8

PR #257, opened 2022-09-14
Updates for Xcode 14

jacobsimeon:
## Summary
The biggest change here is the switch from using a regular `UIButton` to trigger a save action from a URL in the clipboard. It also contains a few other fixes to unit and UI tests to account for slight changes in behavior in the new version of XCTest.

## References 
* Links to docs, tickets, designs if available

## Implementation Details
- [Avoid usages of pasteboard in unit tests](https://github.com/Pocket/pocket-ios/commit/70a9be36410387784cd2930a872979a9665e3ba1)
- [Run tests via MainActor when necessary](https://github.com/Pocket/pocket-ios/commit/ef800f851517c477677bb32fa6f6505cd0c25508)
- [Avoid system alert when pasting URLs](https://github.com/Pocket/pocket-ios/commit/83e617aff092a7a49e842383f480a3d2c687c623)
- [Fix a network retry tests](https://github.com/Pocket/pocket-ios/commit/4c1cafbe1ee45b31bbf371d1f87f32a59929737f)
- [Fix UI tests for Xcode 14](https://github.com/Pocket/pocket-ios/commit/4d477d5a1d8ca04b4a80d034a1793752b43fea6c)

### DIFF
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -223,4 +223,4 @@ app:
 meta:
   bitrise.io:
     machine_type_id: g2.12core
-    stack: osx-xcode-13.4.x
+    stack: osx-xcode-14.0.x
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -199,7 +199,7 @@ workflows:
           title: Install test secrets.xcconfig
       - xcode-test@4:
           inputs:
-            - destination: platform=iOS Simulator,name=iPhone 13,OS=latest
+            - destination: platform=iOS Simulator,name=iPhone 14,OS=latest
       - deploy-to-bitrise-io@2: {}
     before_run:
       - _setup
--- a/PocketKit/Sources/PocketKit/Article/ReadableViewModel/RecommendationViewModel.swift
+++ b/PocketKit/Sources/PocketKit/Article/ReadableViewModel/RecommendationViewModel.swift
@@ -30,15 +30,17 @@ class RecommendationViewModel: ReadableViewModel {
 
     private let recommendation: Recommendation
     private let source: Source
+    private let pasteboard: Pasteboard
     let tracker: Tracker
 
     private var savedItemCancellable: AnyCancellable?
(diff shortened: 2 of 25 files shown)


bassrock:
@jacobsimeon Question will the tests here still run on iOS 15 simulators within Xcode 14?  (I'm asking mainly because I see bitrise is failing on Xcode 13, and at some point I think we will want to run tests on iOS 15 and iOS 16 since we will support both.)

jacobsimeon:
@bassrock I just updated `bitrise.yml` to run the tests on the Xcode 14 stack

bitrise.yml
```diff
@@ -191,4 +191,4 @@ app:
 meta:
   bitrise.io:
     machine_type_id: g2.12core
-    stack: osx-xcode-13.4.x
+    stack: osx-xcode-14.0.x
```
bassrock, line 194: Note, I think this _may_ break automatic uploads to app store connect until https://github.com/fastlane/fastlane/pull/20631 is merged. Given we need to start using Xcode 14 is probably ok, though for now.

bassrock:
@jacobsimeon Yep makes sense, but I'm wondering if any of these changes make the tests only work for iOS 16 and above or will we still be able to run the tests on iOS 15 (all within Xcode 14)

bitrise.yml
bassrock: I would bet that upstream fastlane change merges within the next day or so. Its a big change breaking a lot of folks workflows.

jacobsimeon:
The tests should still run and pass on iOS 15 (as long as they're build via Xcode 14)

jacobsimeon:
@bassrock actually, it looks like the tests don't pass on iOS 15 😞 

They definitely run, but they don't pass because of some changes to how XCTest behaves when querying for UI elements.

dskuza (approved):

PR #259, opened 2022-09-15
feat(coverage): adding code coverage reports

bassrock:
## Summary
* Turns on xcode coverage
* Adds in Danger.systems to post code coverage of changed files to the pr comments
    *  Note: we can build more on danger later if we want.

## References 
* https://pocket.slack.com/archives/C01UH57EJKD/p1663186743065739

## Implementation Details
* This implements danger with swift which adds some dependencies.
* Note that code coverage is on for all targets, this is because for some reason Xcode won't generate code coverage  unless it we tell it all.

## PR Checklist:
- [ ] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/Pocket.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/Pocket.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -20,6 +20,15 @@
         }
       },
       {
+        "package": "DangerSwiftCoverage",
+        "repositoryURL": "https://github.com/f-meloni/danger-swift-coverage",
+        "state": {
+          "branch": null,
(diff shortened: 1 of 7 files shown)


dskuza:
@bassrock Is there an example of what this output will look like? This PR somewhat ironically doesn't show anything.

bassrock:
@dskuza funny enough there was, but it removed it because it was all successful. It will look like this. https://github.com/fastlane-community/danger-xcov

dskuza:
@bassrock Cool, and this really only affects CI, right? (Looks like there's just a tiny bit added to `bitrise.yml`)

bassrock:
@dskuza yea thats correct

dskuza (approved):
@bassrock I don't see why we _shouldn't_ do this, but it seems like it might be a good retro topic at some point (just to be on the same page).

PR #260, opened 2022-09-19
fix(rename): fixing bug in rename

bassrock:
## Summary
* The find and replace of UserByToken accidentally took API.swift with it in ways it should not have.  This fixes that bug. 

## Implementation Details
* Note that the tests continued to pass because API.swift was modified and checked in. Moving forward I updated bitrise.yml always to download the latest API Schema which would have caught this bug after the find and replace.

## Test Steps

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA


dskuza (approved):
Good catch!

PR #262, opened 2022-09-20
feat(v3): adding initial v3 client

bassrock:
## Summary
* For IN-857 we need to create an initial client that can be used to run requests against V3.

## References 
* IN-860

## Implementation Details
* I chose to use Alamofire as the base networking client because it allows for RequestInterceptors similarly to how Apollo client has them and it will handle the response encoding/decoding for us. 

## Test Steps
* 

## PR Checklist:
- [ ] Added Unit / UI tests
- [ ] Basic Self QA


jacobsimeon (commented):
I see that this is just a draft PR, so here are just a few thoughts to consider:

1. For me, the use of AlamoFire doesn't seem worth the added dependency. URLSession has come a long way since the days of the old AFNetworking. I'd encourage you to explore using URLSession.
2. I don't think the v3 client needs to be accessed via Source. Source is mostly concerned with (re)enqueuing sync operations. I'd prefer to see the client get injected wherever it is needed.
3. For encoding parameters, do we need to send any GET requests? If not, then we could use the built-in `JSONEncoder`.
4. I'd love to see a protocol extracted from V3 client. This would allow use to create a mock version of the client for testing purposes.
5. Speaking of tests, would love to see some added for the client.

One final thought: I think we may be getting ahead of ourselves by adding code that is somewhat speculative. Would love to see this trimmed down to just the parts that we need for push registration/deregistration.

bassrock:
@jacobsimeon 

> I see that this is just a draft PR, so here are just a few thoughts to consider:
Yep that's on purpose to get early feedback on the approach.

> 1. For me, the use of AlamoFire doesn't seem worth the added dependency. URLSession has come a long way since the days of the old AFNetworking. I'd encourage you to explore using URLSession.

I can explore this, but I'd like to see what others think too, Alamofire is pretty lightweight and has some nice built-in stuff to take advantage of.

> 2. I don't think the v3 client needs to be accessed via Source. Source is mostly concerned with (re)enqueuing sync operations. I'd prefer to see the client get injected wherever it is needed.

That makes sense, can we get some documentation added to that affect? Perhaps a comment on the file to indicate it's only for operations needing sync?

> For encoding parameters, do we need to send any GET requests? If not, then we could use the built-in JSONEncoder.
I can remove the get functionality. 

> I'd love to see a protocol extracted from V3 client. This would allow use to create a mock version of the client for testing purposes.
> Speaking of tests, would love to see some added for the client.
Yep I'm aware this needs tests and protocols, i wanted to ensure this was a path the team was ok with first with how i went about it before i spent the time on tests.

> One final thought: I think we may be getting ahead of ourselves by adding code that is somewhat speculative. Would love to see this trimmed down to just the parts that we need for push registration/deregistration.
The only thing I see as speculative is the GET request functionality, is that your only concern? If not can you point me to what else is speculative?

PocketKit/Sources/SharedPocketKit/Device.swift
```diff
@@ -0,0 +1,26 @@
...
+import UIKit
+
+public class Device {
+
+    // static property to create singleton
+    public static let current = Device()
+
+    public func deviceIdentifer() -> String {
+        guard let identifierForVendor = UIDevice.current.identifierForVendor else {
+            // TODO: Save this string off into the keychain
```
bassrock, line 18: Will come back to this in a future pr, still deciding if this is necessary. Old app had a lot of fallbacks.

PocketKit/Sources/Sync/OEmbedService.swift
```diff
@@ -1,17 +1,17 @@
 import Foundation
 
-class OEmbedService {
-    enum Error: Swift.Error {
+public class OEmbedService {
```
bassrock, line 3: Due to the move to Sync, it needed to be made public.

PocketKit/Sources/Sync/URLSessionProtocol.swift
```diff
@@ -4,7 +4,7 @@
 
 import Foundation
 
-protocol URLSessionProtocol {
+public protocol URLSessionProtocol {
```
bassrock, line 7: Due to the move to sync

jacobsimeon (approved):

PocketKit/Sources/Sync/V3/V3Client.swift
```diff
@@ -0,0 +1,183 @@
...
+    ) {
+        self.sessionProvider = sessionProvider
+        self.consumerKey = consumerKey
+        self.urlSession = urlSession
+    }
+
+    /**
+     Helper function to execute a V3 request so we can re-use error responses and the type decoding.
+     */
+    internal func executeRequest<T>(
```
dskuza, line 67:
```suggestion
    func executeRequest<T>(
```

`func`s and properties are implicitly internal unless marked otherwise; we've only been marking functions that aren't external with their required access. I'd like to stick with that.

PocketKit/Sources/Sync/V3/V3Client.swift
```diff
@@ -0,0 +1,183 @@
...
+        request: URLRequest,
+        decodeable: T.Type
```
dskuza, line 69:
```suggestion
        decodable: T.Type
```

Just to match the Swift protocol naming

PocketKit/Sources/Sync/V3/V3RequestTypes.swift
```diff
@@ -0,0 +1,32 @@
+protocol BasicV3Request {
```
dskuza, line 1: The Sync package _does_ have a `Session` protocol (see `ApolloClient+Extensions.swift` and its usage in `AppSession+Sync.swift`). There might be an opportunity for extension (if / as necessary) and reuse. Maybe even just having this request have a `var session: Session? { get }` requirement rather than explicitly defining other values? Not sure, just an idea.

PocketKit/Sources/PocketKit/PocketNotificationService.swift
```diff
@@ -79,12 +100,37 @@ class PocketNotificationService: NSObject {
         }
     }
 
+    /**
+     Perform the following notification actions on Logout:
+     - Remove instant sync
+     */
+    private func loggedOut() {
```
dskuza, line 107: Do we need the complement to `braze.changeUser(userId:) in this function?

PocketKit/Sources/PocketKit/PocketNotificationService.swift
```diff
@@ -167,3 +213,19 @@ extension PocketNotificationService: BrazeInAppMessageUIDelegate {
         // Executed when `message` is presented to the user
     }
 }
+
+/**
+ V3 Helpers
+ */
+extension PocketNotificationService {
+
+    static func pushType() -> PushType {
```
dskuza, line 222: This can be a property rather than a function since it's not really performing any function; rather, it's defining a property of the service.

PocketKit/Sources/Sync/V3/V3Client.swift
```diff
@@ -0,0 +1,183 @@
...
+            throw Error.serverError
+        default:
+            throw Error.unexpectedError
+        }
+    }
+
+    /**
+     Given a path and parameters build a URL Request to be used on V3
+     */
+    internal func buildRequest(path: String, parameters: Encodable & BasicV3Request)  throws -> URLRequest {
```
dskuza, line 113:
```suggestion
    func buildRequest(path: String, request: Encodable & BasicV3Request)  throws -> URLRequest {
```

dskuza (commented):
Cool beans with the tests! Just leaving a few comments with some ideas (and some minor stylistic nitpicks). Great work adding tests!

PocketKit/Sources/PocketKit/PocketNotificationService.swift
bassrock: That's part of the issue I'm emailing braze about, it doesn't exist atm.

dskuza (approved):
@bassrock Should we be doing this as a feature branch (i.e until we get rid of the comments)?

PR #263, opened 2022-09-20
chore(lint): linting stuff from pre-swiftlint

bassrock:
## Summary
* A few files seem to have gotten merged without lifting. I think it's because Danger was not introduced on those PRs yet. Let's keep an eye out for this for a bit until we dive in.


dskuza (approved):
This seems primarily like whitespace again, which is good. Most of these seem related to the Settings work.

@kybhamilton  do you have this enabled in your Xcode settings?

Untitled

jacobsimeon (approved):

PR #267, opened 2022-09-22
Use correct consumer key on iPad

dskuza:
## Summary

Use the "legacy" consumer key for "pad" idioms. This should *_not_* be merged until all the appropriate bitrise config files have been updated as well.

## References 

- IN-872

## Implementation Details

- [ ] Add new row to `Info.plist` for both `PocketKit` and `SaveToPocketKit`
- [ ] Update `Keys.swift` in both `PocketKit` and `SaveToPocketKit` to obtain _both_ consumer keys, and set the appropriate key based on the device's user interface idiom.

Note: the key contains the word "pad" vs "iPad" because the idiom is "pad", and I wanted the two to match.

## PR Checklist:
- [ ] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/Info.plist
+++ b/Info.plist
@@ -68,6 +68,8 @@
 	<string>$(POCKET_API_BASE_URL)</string>
 	<key>PocketAPIConsumerKey</key>
 	<string>$(POCKET_API_CONSUMER_KEY)</string>
+	<key>PocketAPIConsumerKeyPad</key>
+	<string>$(POCKET_API_CONSUMER_KEY_PAD)</string>
 	<key>PocketPremiumAnnual</key>
 	<string>$(POCKET_PREMIUM_ANNUAL)</string>
 	<key>PocketPremiumMonthly</key>
--- a/PocketKit/Sources/PocketKit/Keys.swift
+++ b/PocketKit/Sources/PocketKit/Keys.swift
@@ -2,7 +2,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-import Foundation
+import UIKit
 
 struct Keys {
     static let shared = Keys()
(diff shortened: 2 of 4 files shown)


Info.plist
```diff
@@ -54,6 +54,8 @@
 	<string>$(POCKET_API_BASE_URL)</string>
 	<key>PocketAPIConsumerKey</key>
 	<string>$(POCKET_API_CONSUMER_KEY)</string>
+	<key>PocketAPIConsumerKeyPad</key>
+	<string>$(POCKET_API_CONSUMER_KEY_PAD)</string>
```
bassrock, line 58: Note add to bitrise env variables

bassrock (commented):
Works for me, what is the experience like going from an old build to this? Does it auto log users out or does it error?

Info.plist
dskuza: @bassrock I believe we discussed this live and determined we would perform no other action at this time.

bassrock: @dskuza right, but would still like to understand what the experience is going to do so it can be communicated.

dskuza: @bassrock Fair, I missed the second part of your comment somehow 😅 . I can test that path, but I believe things will just seem to … not … be … working? Off to verification land…

dskuza: @bassrock Looks like things like pull-to-refresh on My List fail and spin, and empties the list. Looks like at least _one_ resolution would be communicate to iPad users that if they're seeing issues to log out and log back in (which I still think is okay since we're in beta - this isn't something that would affect a production release).

bassrock: @dskuza cool, yep given this would make the next beta if merged today, and @nzeltzer will be emailing folks, let's post that in the slack channel for the team to know.

bassrock (approved):
Approved, pending the slack heads up

dskuza:
Although approved, do *_not_* yet merge this pull request.

jacobsimeon (approved):
lgtm 👍

PR #271, opened 2022-10-04
Fetch all tags on initial sync

jacobsimeon:
## Summary
Adds logic to `FetchList` operation that will download all the user's tags on the initial sync (since subsequent tags will be sync'd as they're added).

## References 
https://getpocket.atlassian.net/browse/IN-866

## Implementation Details
- Add `TagsQuery` to Apollo codegen
- Adds some logic to paginate through all the user's tags in `FetchList`
- Updates `FetchListTests` to properly stub out requests for tags in all methods
- Update UI tests to stub out the request for tags as well

## Test Steps

Scenario: Tag shows up in tag filter menu
1. Create an tag with no associated items on a separate device
2. Log into the app for the first time
3. Navigate to My List
4. Tap "Tagged" in filter carousel
5. The tag created in Step 1 should be present

Scenario: Tag appears in Add Tag screen
1. Create an tag with no associated items on a separate device
2. Log into the app for the first time
3. Navigate to My List (make sure you have at least one item)
4. Tap overflow menu for item, Tap "Add tags"
5. The tag created in Step 1 should be present

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/Sync/Operations/FetchList.swift
+++ b/PocketKit/Sources/Sync/Operations/FetchList.swift
@@ -94,7 +94,7 @@ class FetchList: SyncOperation {
         guard let edges = result.data?.userByToken?.savedItems?.edges else {
             return
         }
-        // TODO: Add deleteOrphanTags here instead of update and refactor below to be in space (IN-788)
+
         for edge in edges {
             guard let edge = edge, let node = edge.node else {
                 return
--- a/PocketKit/Sources/Sync/Operations/FetchList.swift
+++ b/PocketKit/Sources/Sync/Operations/FetchList.swift
@@ -31,7 +31,9 @@ class FetchList: SyncOperation {
 
     func execute() async -> SyncOperationResult {
         do {
-            try await fetchList()
+            async let doneFetchingList: Void = try fetchList()
+            async let doneFetchingTags: Void = try fetchTags()
+            _ = try await (doneFetchingList, doneFetchingTags)
 
             lastRefresh.refreshed()
             return .success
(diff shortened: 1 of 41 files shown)


PocketKit/Sources/Sync/Operations/FetchList.swift
```diff
@@ -73,6 +75,31 @@ class FetchList: SyncOperation {
...
+
+            if let pageInfo = result.data?.user?.tags?.pageInfo {
+                pagination.after = pageInfo.endCursor
+                shouldFetchNextPage = pageInfo.hasNextPage
+            } else {
+                shouldFetchNextPage = false
+            }
+        }
+    }
+
```
jacobsimeon, line 102: This is the main thrust of this change

nzeltzer: Appreciate the guidance: I was drowning in the generated code. 🏊‍♀️

nzeltzer: Will this cover the edge case where …

user’s iOS device goes offline;
on web, user saves item;
on web, user tags item with new tag “kayak”;
on web, user deletes item;
user’s iOS device goes back online

Will iOS still download “kayak” tag?

nzeltzer (commented):
@jacobsimeon, one edge case Q.

PocketKit/Sources/Sync/Operations/FetchList.swift
jacobsimeon: Oh shoot. No it wouldn't.

nzeltzer: Okay, let's move forward with this, and then discuss that edge case separately.

nzeltzer (approved):
lgtm, but we will need to address an edge case separately.

PR #277, opened 2022-10-12
chore(depedencies): upgrading swift & apollo client

bassrock:
## Summary

This upgrades the following on the app:
* Swift 5.5 -> 5.7
* Apollo Client 0.53 -> 1.0 🥳 
* Braze 5.3.2 -> 5.5.1
* Lottie -> 3.5.0

## References 
* https://github.com/apollographql/apollo-ios/blob/main/CHANGELOG.md#v100

## Implementation Details
* A key piece here is that API.swift becomes a set of multiple files that are auto-generated by Apollo Client. They are then put into a package called `PocketGraph` that the Sync library can then import and use.
    * All the GraphQL operations are then also stored within PocketGraph and generated during apollo code generation.   

## Todos
- [x] Wait for https://github.com/apollographql/apollo-ios/pull/2554 to release to include schema download and generation locally
- [x] Ensure CI is updated to download and generate the schema
- [x] Figure out an issue with GraphQL input types and it not building.

## Test Steps

## PR Checklist:
- [ ] Added Unit / UI tests
- [ ] Basic Self QA


PocketKit/Sources/Sync/Operations/FetchList.swift
```diff
@@ -93,16 +94,18 @@ class FetchList: SyncOperation {
     }
 
     private func fetchPage(_ pagination: PaginationSpec) async throws -> GraphQLResult<FetchSavesQuery.Data> {
-        let query = FetchSavesQuery(token: token)
-        query.pagination = PaginationInput(
-            after: pagination.cursor,
-            first: pagination.maxItems
+        let query = FetchSavesQuery(
```
AnthonyMDev, line 97: @bassrock  If this is an example of where you are having the issue with the nil coalescing operator, I think it is because the `PaginationInput` itself should also be wrapped in a `GraphQLNullable`. 

The `pagination` parameter has a type of `GraphQLNullable<PaginationInput>`.

We can’t initialize a `GraphQLNullable` from a literal with a custom object. We do it for literals using the `ExpressibleByStringLiteral`, `ExpressibleByIntegerLiteral`, etc. protocols in Swift, but for our generated objects, there is not a great way to do that. So the solution is to do this:
``` swift
let query = FetchSavesQuery(
  token: token,
  pagination: .some(PaginationInput(
     after: pagination.cursor ?? .none,
     first: pagination.maxItems
  )
)
```

You aren't the first person to be confused by this one though, so I'd love to come up with a better alternative. I'm just not sure what that would be.

bassrock: @AnthonyMDev Yea so we did try this and unfortunately, it still gives us a compiler error with the nill coalescing operator. (Had to also wrap `pagination.maxItems` in `.some()`
  
Screen Shot 2022-10-20 at 5 15 13 PM

.gitignore
```diff
@@ -121,5 +121,7 @@ Config/secrets.xcconfig
 **/*.otf
 **/.swiftpm
 **/.tmp
+
 schema.graphqls
 introspection_response.json
+!PocketKit/Sources/PocketGraph/user-defined-operations
```
dskuza, line 127: Is there a conflicting ignore where this had to specifically be added?

PocketKit/Package.swift
```diff
@@ -33,11 +31,11 @@ let package = Package(
...
                 "Analytics",
-                "Lottie",
-                "YouTubePlayerKit",
                 "SharedPocketKit",
-                "BrazeKit",
-                .product(name: "BrazeUI", package: "BrazeKit")
+                .product(name: "YouTubePlayerKit", package: "YouTubePlayerKit"),
+                .product(name: "Lottie", package: "lottie-ios"),
+                .product(name: "BrazeKit", package: "braze-swift-sdk"),
+                .product(name: "BrazeUI", package: "braze-swift-sdk")
```
dskuza, line 38: Here's a part of SPM I still don't have knowledge in. Why the change from string literals?

README.md
```diff
@@ -16,33 +16,35 @@ cp Config/secrets.xcconfig.example Config/secrets.xcconfig
 
 Then replace values in `Config/secrets.xcconfig` with the values you have received.
 
+After you will need to run the API Generation steps below.
+
 ## Pocket Graph (API) Schema
 
-Pocket for iOS uses an autogenerated API.swift file that is created by the Apollo CodeGen target.
+Pocket for iOS uses Apollo client to autogenerate it's API schema code. You will need to run the following commands everytime the api's you use change or if you change your api queries.
```
dskuza, line 23:
```suggestion
Pocket for iOS uses Apollo client to autogenerate its API schema code. You will need to run the following commands every time the APIs you use change or if you change your API queries.
```

Nitpick.

dskuza (approved):
Looks good, thanks for taking care of this! Just a few comments for curiosities and sake of nitpicking.

timc-mozilla (approved):

PocketKit/Package.swift
bassrock: Because swift 5.7 changed it! yay!

cyndichin (approved):
thanks for the walkthrough! looks great :)

PocketKit/Sources/Sync/Operations/ArchiveService.swift
```diff
@@ -364,13 +365,13 @@ private class FetchArchivePagesOperation: AsyncOperation {
...
             let result = try await apollo.fetch(
                 query: SavedItemSummariesQuery(
-                    pagination: PaginationInput(
-                        after: cursor,
-                        first: pageSize
-                    ),
+                    pagination: .some(PaginationInput(
+                        after: cursor ?? .none,
+                        first: .some(pageSize)
+                    )),
```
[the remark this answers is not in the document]
bassrock: @dskuza yea https://www.apollographql.com/docs/ios/fetching/operation-arguments#working-with-nullable-arguments its a change that apollo made to the API.

.gitignore
bassrock: No, at one point i had ignored the whole graph folder

PR #278, opened 2022-10-12
IN 874 - updating untagged filter

timc-mozilla:
## Summary
* Added `_untagged_` to filter no tags in archive

## References 
* [IN-874](https://getpocket.atlassian.net/jira/software/projects/IN/boards/80?selectedIssue=IN-874)

## Implementation Details
* handle untagged from graphql

## Test Steps
* go to archives
* sort by tag
* select untagged

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/Sync/Operations/ArchiveService.swift
+++ b/PocketKit/Sources/Sync/Operations/ArchiveService.swift
@@ -360,7 +360,7 @@ private class FetchArchivePagesOperation: AsyncOperation {
 
             var tagNames: [String] = []
             if let tagName = tagName {
-                tagNames = tagName == "not tagged" ? ["untagged"] : [tagName]
+                tagNames = tagName == "not tagged" ? ["_untagged_"] : [tagName]
             }
             let result = try await apollo.fetch(
                 query: SavedItemSummariesQuery(
--- a/Tests iOS/MyList/ArchiveFiltersTests.swift
+++ b/Tests iOS/MyList/ArchiveFiltersTests.swift
@@ -104,4 +104,23 @@ class ArchiveFiltersTests: XCTestCase {
         app.myListView.selectedTagChip(for: "not tagged").buttons.element(boundBy: 0).tap()
         XCTAssertEqual(app.myListView.wait().itemCells.count, 2)
     }
+
+    func test_archiveView_sortingNoTagFilter_showFilteredItems() {
+        app.launch().tabBar.myListButton.wait().tap()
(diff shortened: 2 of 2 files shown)


Tests iOS/MyList/ArchiveFiltersTests.swift
```diff
@@ -104,4 +104,23 @@ class ArchiveFiltersTests: XCTestCase {
         app.myListView.selectedTagChip(for: "not tagged").buttons.element(boundBy: 0).tap()
         XCTAssertEqual(app.myListView.wait().itemCells.count, 2)
     }
+
```
timc-mozilla, line 107: I don't think the test above is testing the name of the test.

cyndichin: Good observation! Some background on it:
The naming was to be consistent with "Favorites" and "All" test names and was not created with the idea of only testing `not tagged`, but any tag value. In general, wanted to test the functionality of the `Tagged` filter. I just decided to choose `not tagged` at that time, but we could have implemented using `tag 0` too. For UI tests, they take longer to run than Unit tests typically so there's a balance in covering all cases and in this case just wanted to test generally that filtering works from UI perspective. However, open to changing the name if that makes more sense. If no differences besides the name, then we should only keep one of the test. Hope that clarifies the reasoning :)

PocketKit/Sources/Sync/Operations/ArchiveService.swift
```diff
@@ -360,7 +360,7 @@ private class FetchArchivePagesOperation: AsyncOperation {
 
             var tagNames: [String] = []
             if let tagName = tagName {
-                tagNames = tagName == "not tagged" ? ["untagged"] : [tagName]
+                tagNames = tagName == "not tagged" ? ["_untagged_"] : [tagName]
```
cyndichin, line 363: 🔥

Tests iOS/MyList/ArchiveFiltersTests.swift
cyndichin: @timc-mozilla Let's keep one of the tests so we're not spending time to run dup tests, else ready for approval!

cyndichin (changes requested):
Looks great! - lets move analytics to another PR as don't want to block this ticket. Just one comment in removing one of the duplicate tests

cyndichin (approved):
thanks @timc-mozilla :) like that you decided to test both cases

PR #280, opened 2022-10-13
IN-321: Add Tags to Save Extension

cyndichin:
## Summary
Added the ability for users to add tags to a Saved Item via Save Extension

## References 
IN-321

## Implementation Details
* Move `AddTagsView` into the shared UI Component `Textile` and created a two view models (one for PocketKit and one for SaveTo Extension) which both conform to `AddTagsViewModel` protocol. Modified the `PocketSaveService` to be able to send the GraphQL mutations related to tags and update the saved item accordingly. Added a new `infoViewModel` called `.taggedItem` to show the user a confirmation view.
* Change the copy of existing item from `You've already saved this before. We'll move it to the top of your list.` to `You've already saved this. We'll move it to the top of your list.` as confirmed with product.
* Extended dismiss timer as it disappeared too quickly

## Test Steps
- [ ] Given that a user saves item to pocket using shared menu, 
and the user taps on Add Tags button, 
then the user should be able to view the Add Tags screen. 

- [ ] Given that sees the Add Tags view via shared extension, 
and the user interacts with it, 
then the behavior should be the same as general Add Tags feature.

- [ ] Given that a user saves a tag, 
then the user should view a confirmation page.

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

## Screenshots

https://user-images.githubusercontent.com/6743397/195681326-4f7ac0f0-fa34-4fd8-aaa4-b420dc6553a6.mp4

### DIFF
--- a/PocketKit/Sources/PocketKit/Tags/PocketAddTagsViewModel.swift
+++ b/PocketKit/Sources/PocketKit/Tags/PocketAddTagsViewModel.swift
@@ -15,7 +15,7 @@ class PocketAddTagsViewModel: AddTagsViewModel {
         self.source = source
         self.saveAction = saveAction
 
-        tags = item.tags?.compactMap { $0 as? Tag }.map { $0.name ?? "" } ?? []
+        tags = item.tags?.compactMap { ($0 as? Tag)?.name } ?? []
     }
 
     func addTags() {
--- a/PocketKit/Sources/PocketKit/Tags/PocketAddTagsViewModel.swift
+++ b/PocketKit/Sources/PocketKit/Tags/PocketAddTagsViewModel.swift
@@ -1,0 +1,30 @@
+import Combine
+import Sync
+import Textile
+
+class PocketAddTagsViewModel: AddTagsViewModel {
+    private let item: SavedItem
+    private let source: Source
+    private let saveAction: () -> Void
+
+    @Published
+    var tags: [String] = []
+
+    init(item: SavedItem, source: Source, saveAction: @escaping () -> Void) {
+        self.item = item
+        self.source = source
+        self.saveAction = saveAction
+
+        tags = item.tags?.compactMap { $0 as? Tag }.map { $0.name ?? "" } ?? []
+    }
+
+    func addTags() {
+        source.addTags(item: item, tags: tags)
+        saveAction()
+    }
+
+    func allOtherTags() -> [String]? {
+        let fetchedTags = source.retrieveTags(excluding: tags)
+        return fetchedTags?.compactMap { $0.name }
+    }
+}
--- a/PocketKit/Sources/SaveToPocketKit/MainViewController.swift
+++ b/PocketKit/Sources/SaveToPocketKit/MainViewController.swift
@@ -28,7 +28,7 @@ class MainViewController: UIViewController {
                 viewModel: SavedItemViewModel(
                     appSession: appSession,
                     saveService: services.saveService,
-                    dismissTimer: Timer.TimerPublisher(interval: 2.1, runLoop: .main, mode: .default)
+                    dismissTimer: Timer.TimerPublisher(interval: 3.0, runLoop: .main, mode: .default)
                 )
             )
         }
--- a/PocketKit/Sources/SaveToPocketKit/SaveToAddTagsViewModel.swift
+++ b/PocketKit/Sources/SaveToPocketKit/SaveToAddTagsViewModel.swift
@@ -15,7 +15,7 @@ class SaveToAddTagsViewModel: AddTagsViewModel {
         self.retrieveAction = retrieveAction
         self.saveAction = saveAction
 
-        tags = item?.tags?.compactMap { $0 as? Tag }.map { $0.name ?? "" } ?? []
+        tags = item?.tags?.compactMap { ($0 as? Tag)?.name } ?? []
     }
 
     func addTags() {
--- a/PocketKit/Sources/SaveToPocketKit/SaveToAddTagsViewModel.swift
+++ b/PocketKit/Sources/SaveToPocketKit/SaveToAddTagsViewModel.swift
@@ -1,0 +1,29 @@
+import Combine
+import Sync
+import Textile
+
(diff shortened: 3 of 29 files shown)


timc-mozilla (approved):
tested locally and works as expected.  Code looks good, but I'll defer my judgement to a more senior member on the team.  If no one else has time, code wise, looks good to me.

PocketKit/Sources/SaveToPocketKit/SaveToAddTagsViewModel.swift
```diff
@@ -0,0 +1,29 @@
+import Combine
+import Sync
+import Textile
+
+class SaveToAddTagsViewModel: AddTagsViewModel {
+    private let item: SavedItem?
+    private let retrieveAction: ([String]) -> [Tag]?
+    private let saveAction: ([String]) -> Void
```
dskuza, line 8: What's the thinking behind having these as functions that are injected? An immediate thought is so that we don't have to further pass down dependencies. Perhaps this makes testing easier, too? Just curious.

dskuza (approved):
just a few comments, nothing blocking

PocketKit/Sources/SaveToPocketKit/SavedItem/SavedItemViewModel.swift
```diff
@@ -35,21 +40,55 @@ class SavedItemViewModel {
...
+            let result = saveService.save(url: url)
+            switch result {
+            case .existingItem(let savedItem):
+                self.savedItem = savedItem
                 infoViewModel = .existingItem
-            case .newItem:
+            case .newItem(let savedItem):
+                self.savedItem = savedItem
                 infoViewModel = .newItem
+            case .taggedItem(_):
```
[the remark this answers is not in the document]
cyndichin: yes you are correct - will omit

PocketKit/Sources/SaveToPocketKit/MainViewController.swift
```diff
@@ -28,7 +28,7 @@ class MainViewController: UIViewController {
                 viewModel: SavedItemViewModel(
                     appSession: appSession,
                     saveService: services.saveService,
-                    dismissTimer: Timer.TimerPublisher(interval: 2.1, runLoop: .main, mode: .default)
+                    dismissTimer: Timer.TimerPublisher(interval: 3.0, runLoop: .main, mode: .default)
```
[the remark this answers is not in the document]
cyndichin: also went over with Laurie, so got her 👍 as well :)

PocketKit/Sources/PocketKit/Tags/PocketAddTagsViewModel.swift
```diff
@@ -0,0 +1,30 @@
...
+
+    @Published
+    var tags: [String] = []
+
+    init(item: SavedItem, source: Source, saveAction: @escaping () -> Void) {
+        self.item = item
+        self.source = source
+        self.saveAction = saveAction
+
+        tags = item.tags?.compactMap { $0 as? Tag }.map { $0.name ?? "" } ?? []
```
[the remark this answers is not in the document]
cyndichin: great catch! I actually just used one `compactMap` instead 
do you have a preference? 

`tags = item.tags?.compactMap { ($0 as? Tag)?.name } ?? []`

note: updated for `SaveToAddTagsViewModel` as well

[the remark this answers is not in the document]
dskuza: The single `compactMap` looks fine to me - I think the mapping is simple enough where this is fine.

PocketKit/Sources/SaveToPocketKit/SaveToAddTagsViewModel.swift
cyndichin: These functions ultimately calls methods in the `PocketSaveService` which the `SavedItemViewModel` already depends on and so I felt that it was simpler to utilize the save service in that view model instead of injecting it in the `SaveToAddTagsViewModel`. I personally think it makes testing easier too for `SaveToAddTagsViewModel` since we don't need to consider the save service + stub the methods, though we do have a mock for it.

I didn't do this same type of implementation with `source` for PocketAddTagsViewModel since I didn't think of this way until this PR but I can if you recommend to be consistent. Is there a way / implementation you prefer? Thanks for asking this - love the curious questions 😄

dskuza: It seems akin to the factory pattern (though we're not injecting a dependency), which is something I don't have a lot of "experience" with. @jacobsimeon is more well-versed in it than me, so I'd be interested what his $0.02 here are. But, nothing blocking, just a curiosity.

cyndichin: okay! I do have a tech debt ticket to refactor parts of the tag work, so we can revisit the discussion then! thanks @dskuza

jacobsimeon: The pattern @cyndichin is using seems like an alternative to a) having a delegate that can respond to certain actions or b) injecting in another dependency like `source`. I see Cyndi's point that it makes testing easier since you don't have to inject a mock source. That said, the actual method being called on source is a critical piece of behavior in the case of saving tags.
So, I'd actually prefer to see `source` getting injected here and then have a test on `SaveToAddTagsViewModel` that asserts we're calling the correct methods. Though, it does occur to me that it would get complicated since we're in the context of the save extension.

Not saying anything needs to change here, but I do prefer passing in the object(s) you need (`source` or `saveService` in this case) to do work instead of passing in some closures that happen to be easier to construct when initializing the view model.

That said, it would be tricky to have this view model take in a single dependency since the main app wants to use `Source` and the extension wants to use `SaveService`. So you'd end up having to create another protocol that can be conformed to by both `Source` and `SaveService` and injecting that.

Again, not saying anything needs to change (this is a perfectly fine solution), but you asked for my $0.02. I guess I gave $200 though 😆

PR #282, opened 2022-10-14
fix(curatedInfo): pulling in old curatedInfo object for recomendations

bassrock:
## Summary
* Ensure that we follow how other clients choose to display item recommendation data, until this logic is fully upserted into the Pocket Graph. 

## References 
* https://getpocket.atlassian.net/browse/IN-748

## Implementation Details
* I followed the derivers implemented by the Web team [here](https://github.com/Pocket/web-client/blob/75f606321564cd85da82cc8ace0a8e14c799a48c/src/common/api/derivers/item.js#L268)
* I moved queries and fragments out of list.graphql into their own files. Once we have Apollo 1.0 setup and merged, this organization will get even more cleaner: https://github.com/Pocket/pocket-ios/pull/277

### Non-Goal
* Switching to Unified Home + Corpus Items

## Test Steps
* View Home
    * When home is viewed, and we have curated metadata present it is displayed.
    * You can verify correct metadata is shown by navigating to our curation tools and searching each recommendation in the corpus. When you pull up the item in the corpus, and the item has run on Firefox New Tab the metadata should match in the iOS app.

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/Sync/RemoteMapping/SlateLineup+remoteMapping.swift
+++ b/PocketKit/Sources/Sync/RemoteMapping/SlateLineup+remoteMapping.swift
@@ -44,6 +44,9 @@ extension Recommendation {
 
     func update(from remote: RemoteRecommendation, in space: Space) {
         remoteID = remote.id
+        title = remote.curatedInfo?.title
+        excerpt = remote.curatedInfo?.excerpt
+        imageURL = remote.curatedInfo?.imageSrc.flatMap(URL.init)
 
         let recommendationItem = try? space.fetchOrCreateItem(byRemoteID: remote.item.remoteId)
         recommendationItem?.update(from: remote.item.fragments.itemSummary)
--- a/PocketKit/Sources/Sync/Slates/SlateService.swift
+++ b/PocketKit/Sources/Sync/Slates/SlateService.swift
@@ -22,11 +22,15 @@ class APISlateService: SlateService {
     func fetchSlateLineup(_ identifier: String) async throws {
         let query = GetSlateLineupQuery(lineupID: identifier, maxRecommendations: 5)
 
-        guard let remote = try await apollo.fetch(query: query).data?.getSlateLineup else {
-            return
+        do {
+            guard let remote =  try await apollo.fetch(query: query).data?.getSlateLineup else { return }
+            try await handle(remote: remote)
+
+        } catch {
+            print("error \(error)")
+            throw error
         }
 
-        try await handle(remote: remote)
     }
 
     func fetchSlate(_ slateID: String) async throws {
--- a/PocketKit/Sources/Sync/Slates/SlateService.swift
+++ b/PocketKit/Sources/Sync/Slates/SlateService.swift
@@ -22,15 +22,11 @@ class APISlateService: SlateService {
     func fetchSlateLineup(_ identifier: String) async throws {
         let query = GetSlateLineupQuery(lineupID: identifier, maxRecommendations: 5)
 
-        do {
-            guard let remote =  try await apollo.fetch(query: query).data?.getSlateLineup else { return }
-            try await handle(remote: remote)
-
-        } catch {
-            print("error \(error)")
-            throw error
+        guard let remote = try await apollo.fetch(query: query).data?.getSlateLineup else {
+            return
         }
 
+        try await handle(remote: remote)
     }
 
     func fetchSlate(_ slateID: String) async throws {
--- a/PocketKit/Tests/PocketKitTests/Recomendation/RecomendationExtensionTests.swift
+++ b/PocketKit/Tests/PocketKitTests/Recomendation/RecomendationExtensionTests.swift
@@ -1,0 +1,121 @@
+import XCTest
+import Sync
+
+@testable import Sync
+@testable import PocketKit
+
+class RecomendationExtensionTests: XCTestCase {
+    private var source: MockSource!
+    private var space: Space!
+
+    override func setUp() {
+        source = MockSource()
+        space = .testSpace()
+
+        continueAfterFailure = false
+    }
+
+    override func tearDownWithError() throws {
+        try space.clear()
+    }
+
+// MARK: - Builder Helpers
+
+    func buildItem(syndicatedArticle: SyndicatedArticle? = nil) -> Item {
+        return space.buildItem(
+            topImageURL: URL(string: "https://top-image-url.jpeg"),
+            excerpt: "Some item excerpt",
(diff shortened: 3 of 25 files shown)


PocketKit/Sources/Sync/Slates/SlateService.swift
```diff
@@ -22,11 +22,15 @@ class APISlateService: SlateService {
...
 
-        guard let remote = try await apollo.fetch(query: query).data?.getSlateLineup else {
-            return
+        do {
+            guard let remote =  try await apollo.fetch(query: query).data?.getSlateLineup else { return }
+            try await handle(remote: remote)
+
+        } catch {
+            print("error \(error)")
+            throw error
```
dskuza, line 31: This is some debug code left over from our rubber ducky session. Let's try and figure out if this error is bubbling up correctly, and also determine how to best handle this error (revert to rethrowing, or capturing here. Either way, remove the print).

PocketKit/Tests/PocketKitTests/Recomendation/RecomendationExtensionTests.swift
```diff
@@ -0,0 +1,121 @@
...
+
+    func subjectWithCuratedInfo() -> Recommendation {
+        let recommendation = space.buildRecommendation(
+            item: buildItem(),
+            imageURL: URL(string: "https://curated-info-image.jpeg"),
+            title: "A Curated Title",
+            excerpt: "Some curated excerpt"
+        )
+        return recommendation
+    }
```
dskuza, line 45: I feel like there's an opportunity for some cleanup, here. At bare minimum, `space.buildRecommendation(item: buildItem())` is reused. I'm wondering if we can consolidate this into a single API. Not required, just a potential for cleanup.

PocketKit/Tests/PocketKitTests/Recomendation/RecomendationExtensionTests.swift
```diff
@@ -0,0 +1,121 @@
...
+     */
+    func assertCachedImageURLContainsOriginalImageURL(imageURL: URL, originalUrl: URL) {
+        XCTAssertNotNil(imageURL.absoluteString.range(of: originalUrl.absoluteString))
+    }
+
+    func curatedInfoAssertions(subject: Recommendation) {
+        assertCachedImageURLContainsOriginalImageURL(imageURL: subject.bestImageURL!, originalUrl: URL(string: "https://curated-info-image.jpeg")!)
+        XCTAssertEqual(subject.bestTitle, "A Curated Title")
+        XCTAssertEqual(subject.bestExcerpt, "Some curated excerpt")
+    }
```
dskuza, line 92: I'd like to see these "helper" functions extracted into an extension (within the same file, just separated from the tests themselves).

PocketKit/Tests/PocketKitTests/Recomendation/RecomendationExtensionTests.swift
```diff
@@ -0,0 +1,121 @@
...
+
+    /**
+     Helper function to assert that a cached image url contains the original image.
+     This is because the recomendation helpers will encode the underlying image to a image cache url
+     */
+    func assertCachedImageURLContainsOriginalImageURL(imageURL: URL, originalUrl: URL) {
+        XCTAssertNotNil(imageURL.absoluteString.range(of: originalUrl.absoluteString))
+    }
+
+    func curatedInfoAssertions(subject: Recommendation) {
```
dskuza, line 88:
```suggestion
    func assertCuratedInfo(_ subject: Recommendation) {
```

Naming; this function asserts on information, rather than returning assertions.

dskuza (changes requested):
@bassrock One definitive change requested re: the debug code from our rubber ducky session. Otherwise, this first pass is mainly opportunities for cleanup. I'd _like_ to see them made, then I'll do a second pass.

PocketKit/Sources/Sync/Slates/SlateService.swift
bassrock: Done.

PocketKit/Tests/PocketKitTests/Recomendation/RecomendationExtensionTests.swift
bassrock: Done.

PocketKit/Tests/PocketKitTests/Recomendation/RecomendationExtensionTests.swift
bassrock: Done

bassrock:
@dskuza all changes made

PocketKit/Sources/Sync/RemoteMapping/SlateLineup+remoteMapping.swift
```diff
@@ -44,6 +44,9 @@ extension Recommendation {
 
     func update(from remote: RemoteRecommendation, in space: Space) {
         remoteID = remote.id
+        title = remote.curatedInfo?.title
+        excerpt = remote.curatedInfo?.excerpt
+        imageURL = remote.curatedInfo?.imageSrc.flatMap(URL.init)
```
dskuza, line 49: For my own knowledge: I just want to make sure I'm interpreting this right - a Recommendation's title, excerpt, and imageURL will be based on its curated info, otherwise its syndicated article? Since we don't store curated info in Core Data, the `curatedInfo` that's obtained via query is the middleman to these values.

dskuza (commented):
@bassrock Do these changes _only_ affect home, or is there potential for these changes to also require being added to the slate detail views (i.e when something like "Editors' Picks" -> "See All" is performed)?

edit: I meant to leave this as a comment, and not another "Changes requested".

PocketKit/Sources/Sync/RemoteMapping/SlateLineup+remoteMapping.swift
bassrock: Right now it will. And semi correct, if there's no curatedInfo, then we see if if its syndicated and if so we use that, then if its not syndicated we use the Item data. 

This is also a bit of foresight, because when we move to new home next year, all of this data will be on the Recommendation with no need for fall backs.

bassrock:
@bassrock this change should be for any Reccomendation

dskuza:
@bassrock ah, I see. just for clarification - have you tested that this also affects the slate detail(s) the same way? looking at which view models are affected, my assumption is yes, but I just want to clarify the expectation.

dskuza (approved):

PR #284, opened 2022-10-18
build(danger): fix up danger to always be posting information

bassrock:
## Summary
* Updated Danger to always lint all files
   *  Ensured that danger is commenting on lint failures
   * Make danger post when there are no violations!
* Ensure Danger is posting coverage 
* https://getpocket.atlassian.net/browse/IN-901

## Implementation Decisions

For coverage to work 2 things needed to happen:
* Danger needs to be run and setup in the root folder
* All the pocketkit libraries needed to be added to the main build scheme

> Danger needs to be run and setup in the root folder

While not done in this pr, I wonder if we could or should put everything under PocketKit at the root level. While looking at other app or library setups in GitHub this seems to be the norm and what a lot of swift/xcode plugin libraries are designed to work with.

#### Main comment
Screen Shot 2022-10-18 at 9 04 17 PM

#### Swiftlint comment at top level
Screen Shot 2022-10-18 at 4 08 57 PM

#### Inline Comments
Screen Shot 2022-10-18 at 8 57 01 PM


dskuza:
@bassrock What exactly are the project changes, here? Just the scheme? git doesn't to a good job at visualizing that (since it's just a lovely ol' xml file). Is it this? 
image

bassrock:
@dskuza all of the library targets were not included before. In order for test coverage to be produced they need to be included in the main scheme.

dskuza:
@bassrock another thing (perhaps I've forgotten since the first introduction of this) - what's `Fake.swift` for?

bassrock:
@dskuza Linking is not the issue, nor the multi-package dependency. Purely the folder structure being under one level, cause most build tools do not expect this setup.. I can do an example of what I am thinking.

bassrock:
@dskuza https://github.com/Pocket/pocket-ios/blob/b1775e62520a6532dbfc6341e391b0a158e47677/Sources/DangerDependencies/Fake.swift#L1 😄

README.md
```diff
@@ -92,3 +92,11 @@ git checkout my-pr-branch
 git rebase develop
 git push origin my-pr-branch --force[-with-lease]
 ```
+
+### Running Danger Locally
+
+You can run danger locally to see what it would output by running the following in the root of the repository.
+
+```bash
+swift run danger-swift pr [some-pr-url]
```
cyndichin, line 101: 🔥 would it be possible to run it locally without a PR URL

dskuza (approved):
:+1: I see nothing wrong with this.

I don't want my questions to come off as a "I don't want to change our project structure at all", more-so just a hesitation since things are working fine as they are right now. But, if there's room for improvement overall, cool, and if there are improvements that will help us with regard to build tools, that's cool, too!

README.md
bassrock: @cyndichin yea. `swift run danger-swift local` should work, but haven't used it!

bassrock: Actually i lied, it fails because it tries to do git things.

cyndichin (approved):
Looks great to me 😄 - thanks for this addition! If we come to a commit standard cross org, maybe we can also use danger to warn against PRs that violate that standard.

bassrock:
@dskuza i'll demo what im thinking, its really a small adjustment so we don't need to `cd` or have other issues.

@cyndichin Yes! Check out https://github.com/danger/awesome-danger#swift-danger-swift lots of options.

README.md
cyndichin: ah bummer :(

PR #289, opened 2022-10-19
refactor(failing test): typeText not 100% correct [IN-908]

timc-mozilla:
## Summary
* TypeText was missing the "g" in "tag" every so often causing the test to fail

## References 
* [IN-908](https://getpocket.atlassian.net/jira/software/projects/IN/boards/80?selectedIssue=IN-908)

## Implementation Details
* added each character individually based on David's expertise

## Test Steps
* run `test_userAddTags_showsConfirmationView`

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/Tests iOS/SaveToPocketTests.swift
+++ b/Tests iOS/SaveToPocketTests.swift
@@ -62,23 +62,18 @@ class SaveToPocketTests: XCTestCase {
         safari.buttons["add-tags-button"].wait().tap()
 
         let addTagsView = AddTagsViewElement(safari.otherElements["add-tags"])
-        var tagString = "tag 1"
 
+        // typeText is flakey and cannot type "Tag 1" 100% of the time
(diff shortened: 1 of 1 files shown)


dskuza:
Just popping in to say: this is a very silly fix, but … it works. My personal $0.02 are that if this fixes a flaky test (because the simulator for some reason doesn't want to handle keyboard taps into a text field on time), then this is a valid fix. Our time can be better spent working on other issues aside from "why doesn't the simulator type text correctly?"

cyndichin (commented):
We are using `typeText` in other places which seem to be okay, I'm curious on why there is an issue only in this case?

bassrock:
I am bewildered. Thank you, XCode.

bassrock (approved):

PR #294, opened 2022-10-24
Add "AllTests" test plan and make it the default

jacobsimeon:
This will be the default test plan for use locally. CI runs individual test plans.

### DIFF
--- a/PocketKit/Sources/Sync/PocketSaveService.swift
+++ b/PocketKit/Sources/Sync/PocketSaveService.swift
@@ -219,9 +219,11 @@ class SaveOperation<Mutation: GraphQLMutation>: AsyncOperation {
     }
 
     private func storeUnresolvedSavedItem() {
-        let unresolved: UnresolvedSavedItem = space.new()
-        unresolved.savedItem = savedItem
-        try? space.save()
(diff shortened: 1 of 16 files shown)


dskuza (approved):

PocketKit/Sources/Sync/PocketSaveService.swift
```diff
@@ -219,9 +219,11 @@ class SaveOperation<Mutation: GraphQLMutation>: AsyncOperation {
     }
 
     private func storeUnresolvedSavedItem() {
-        let unresolved: UnresolvedSavedItem = space.new()
-        unresolved.savedItem = savedItem
-        try? space.save()
```
bassrock, line 224: Should this method always save on the main thread under the hood?

bassrock (approved):
🤞

PocketKit/Sources/Sync/PocketSaveService.swift
jacobsimeon: Yes, but so do all of the other methods on `space` (e.g. `new()`). hence wrapping them in `performAndWait` here instead of encapsulating it in the actual `save` method.

PR #303, opened 2022-10-26
fix(tags): Download All Tags [IN-866]

cyndichin:
## Summary
Fix issue where tags were not all downloading after a user adds a tag on a different device and then signs in and view tags on another device.

## References 
IN-866

## Implementation Details
Made changes to how we are passing in the input values in our GraphQL query for PaginationInput. We may want to handle the case in which we don't pass any input values and throw an error or return default via backend, but making this fix for now.

See slack convo here: https://pocket.slack.com/archives/C01VDFM30J2/p1666800998893259

## Test Steps
1. Sign out of pocket
2. Add new tag to oldest Archive item on Web
3. Sign into Pocket
4. Navigate to Saves on iOS 
5. Tap on Add Tags + Filter Tags
6. Verify that new tag appears

## PR Checklist:
- N/A Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/Sync/Operations/FetchList.swift
+++ b/PocketKit/Sources/Sync/Operations/FetchList.swift
@@ -77,7 +77,7 @@ class FetchList: SyncOperation {
 
     private func fetchTags() async throws {
         var shouldFetchNextPage = true
-        var pagination = PaginationInput()
+        var pagination = PaginationInput(first: .null)
 
         while shouldFetchNextPage {
             let query = TagsQuery(pagination: .init(pagination))


PocketKit/Sources/Sync/Operations/FetchList.swift
```diff
@@ -77,7 +77,7 @@ class FetchList: SyncOperation {
 
     private func fetchTags() async throws {
         var shouldFetchNextPage = true
-        var pagination = PaginationInput()
+        var pagination = PaginationInput(first: .null)
```
timc-mozilla, line 80: someone correct me if im wrong, but is `first: .null` just setting the first edge of the page?

timc-mozilla: nvm, sorry. this seems fine. this returns the `first _n_ elements`. 
in that case, do we need this paginated if we are just grabbing every single tag?

cyndichin: By default, it returns 30 for each page (set by backend). I think pagination is still useful to save on network response time and payload size. If 30 is the best number for retrieving tags, not sure, but might be a discussion that we will want to have with backend.

bassrock: @timc-mozilla 
> in that case, do we need this paginated if we are just grabbing every single tag?

Yes, there are some users with thousands of tags we'd need to look at analytics to determine our max use case.

timc-mozilla (approved):

PR #311, opened 2022-11-03
feat(refresh): Add Home Refresh Rule [IN-744]

cyndichin:
## Summary
Add Home Refresh Rule so that when user opens app, it does not automatically refresh unless last refresh is >= 12 hours

## References 
IN-744

## Implementation Details
Added `HomeRefreshCoordinator` that uses user defaults to capture the timestamp of a refresh. Modify `refresh` in `HomeViewModel` to use this coordinator as well. 

## Test Steps
- [ ] WHEN I open the app
AND 12+ hours have elapsed since Home was last refreshed
THEN Home should automatically refresh
- [ ] WHEN I open the app
AND < 12 hours have elapsed since Home was last refreshed
THEN Home should not automatically refresh
- [ ] IF I am on the Home screen
AND I pull to refresh
AND 12+ hours have elapsed since Home was last refreshed
THEN Home should refresh
- [ ] IF I am on the Home screen
AND I pull to refresh
AND < 12 hours have elapsed since Home was last refreshed
THEN Home should refresh

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/PocketKit/Refresh/HomeRefreshCoordinator.swift
+++ b/PocketKit/Sources/PocketKit/Refresh/HomeRefreshCoordinator.swift
@@ -1,0 +1,56 @@
+import Foundation
+import UIKit
+import Combine
+import Sync
+
+protocol HomeRefreshCoordinatorProtocol {
+    func refresh(isForced: Bool, _ completion: @escaping () -> Void)
+}
+
+class HomeRefreshCoordinator: HomeRefreshCoordinatorProtocol {
+    static let dateLastRefreshKey = "HomeRefreshCoordinator.dateLastRefreshKey"
+    private let notificationCenter: NotificationCenter
+    private let userDefaults: UserDefaults
+    private let source: Source
+    private let minimumRefreshInterval: TimeInterval
+    private var subscriptions: [AnyCancellable] = []
+    private var isRefreshing: Bool = false
+
+    init(notificationCenter: NotificationCenter, userDefaults: UserDefaults, source: Source, minimumRefreshInterval: TimeInterval = 12 * 60 * 60) {
+        self.userDefaults = userDefaults
+        self.notificationCenter = notificationCenter
+        self.minimumRefreshInterval = minimumRefreshInterval
+        self.source = source
+
+        self.notificationCenter.publisher(for: UIScene.willEnterForegroundNotification, object: nil).sink { [weak self] _ in
+            self?.refresh { }
+        }.store(in: &subscriptions)
+    }
+
+    func refresh(isForced: Bool = false, _ completion: @escaping () -> Void) {
+        if shouldRefresh(isForced: isForced), !isRefreshing {
+            Task {
+                do {
+                    isRefreshing = true
+                    try await source.fetchSlateLineup(HomeViewModel.lineupIdentifier)
+                    userDefaults.setValue(Date(), forKey: Self.dateLastRefreshKey)
+                    Crashlogger.breadcrumb(category: "refresh", level: .info, message: "Home Refresh Occur")
+                } catch {
+                    Crashlogger.capture(error: error)
+                }
+                completion()
+                isRefreshing = false
+            }
+        }
+    }
+
+    private func shouldRefresh(isForced: Bool = false) -> Bool {
+        guard let lastActiveTimestamp = userDefaults.object(forKey: Self.dateLastRefreshKey) as? Date else {
+            return true
+        }
+
+        let timeSinceLastRefresh = Date().timeIntervalSince(lastActiveTimestamp)
+
+        return timeSinceLastRefresh >= minimumRefreshInterval || isForced
+    }
+}
--- a/PocketKit/Tests/PocketKitTests/Refresh/HomeRefreshCoordinatorTests.swift
+++ b/PocketKit/Tests/PocketKitTests/Refresh/HomeRefreshCoordinatorTests.swift
@@ -1,0 +1,136 @@
+import XCTest
+import Sync
+import Combine
+
+@testable import PocketKit
+
+class HomeRefreshCoordinatorTests: XCTestCase {
+    private var notificationCenter: NotificationCenter!
+    private var userDefaults: UserDefaults!
+    private var source: MockSource!
+    private var subscriptions: Set<AnyCancellable>!
+
+    override func setUpWithError() throws {
+        notificationCenter = NotificationCenter()
+        userDefaults = UserDefaults()
+        source = MockSource()
+        subscriptions = []
+    }
+
+    override func tearDownWithError() throws {
+        userDefaults.removeObject(forKey: HomeRefreshCoordinator.dateLastRefreshKey)
+    }
+
+    func subject(
+        notificationCenter: NotificationCenter? = nil,
+        userDefaults: UserDefaults? = nil,
+        source: Source? = nil,
+        minimumRefreshInterval: TimeInterval = 12 * 60 * 60
+    ) -> HomeRefreshCoordinator {
+        HomeRefreshCoordinator(
+            notificationCenter: notificationCenter ?? self.notificationCenter,
+            userDefaults: userDefaults ?? self.userDefaults,
+            source: source ?? self.source,
+            minimumRefreshInterval: minimumRefreshInterval
+        )
+    }
+
+    func test_firstRefresh_setsUserDefaults() {
+        XCTAssertNil(userDefaults.object(forKey: HomeRefreshCoordinator.dateLastRefreshKey))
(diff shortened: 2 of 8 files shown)


PocketKit/Tests/PocketKitTests/Refresh/HomeRefreshCoordinatorTests.swift
```diff
@@ -0,0 +1,135 @@
...
+        XCTAssertEqual(userDefaults.object(forKey: HomeRefreshCoordinator.dateLastRefreshKey) as? Date, date)
+    }
+
+    func test_coordinator_whenEnterForeground_whenDataIsStale_refreshesHome() {
+        let fetchExpectation = expectation(description: "expected to fetch slate lineup")
+        source.stubFetchSlateLineup { _ in fetchExpectation.fulfill() }
+        let date = Calendar.current.date(byAdding: .hour, value: -2, to: Date())
+        userDefaults.setValue(date, forKey: HomeRefreshCoordinator.dateLastRefreshKey)
+
+        let coordinator = subject()
```
cyndichin, line 127: this gives me a warning message because `coordinator` is not being used, but using `_ = subject()` breaks the test and does not trigger the post, anyone encounter something similar?

timc-mozilla (approved):

PocketKit/Sources/PocketKit/Refresh/HomeRefreshCoordinator.swift
```diff
@@ -0,0 +1,56 @@
...
+class HomeRefreshCoordinator: HomeRefreshCoordinatorProtocol {
+    static let dateLastRefreshKey = "HomeRefreshCoordinator.dateLastRefreshKey"
+    private let notificationCenter: NotificationCenter
+    private let userDefaults: UserDefaults
+    private let source: Source
+    private let refreshRuleTime: Int
+    private var subscriptions: [AnyCancellable] = []
+    private var isRefreshing: Bool = false
+
+    init(notificationCenter: NotificationCenter, userDefaults: UserDefaults, source: Source, refreshRuleTime: Int = 12) {
```
dskuza, line 19: As an alternative to `refreshRuleTime` being an `Int` representing hours, we can pass in a minimum refresh interval (of type `TimeInterval`) and use `Date.timeIntervalSince(_:)` for comparison, rather than using calendar components. This would just open us up to more options if we ever change this time in the future. Non-blocker, but a change that might be nice.

PocketKit/Sources/PocketKit/Refresh/HomeRefreshCoordinator.swift
```diff
@@ -0,0 +1,56 @@
...
+
+    func refresh(isForced: Bool = false, _ completion: @escaping () -> Void) {
+        if shouldRefresh(isForced: isForced), !isRefreshing {
+            Task {
+                do {
+                    isRefreshing = true
+                    try await source.fetchSlateLineup(HomeViewModel.lineupIdentifier)
+                    userDefaults.setValue(Date(), forKey: Self.dateLastRefreshKey)
+                } catch {
+                    Crashlogger.capture(error: error)
```
dskuza, line 38: As a complement to capturing the error, would it be worth adding a breadcrumb in the `do` block to mark that a refresh has been made?

PocketKit/Tests/PocketKitTests/Refresh/HomeRefreshCoordinatorTests.swift
```diff
@@ -0,0 +1,134 @@
...
+        coordinator.refresh {
+            XCTFail("Should not have refreshed")
+        }
+        XCTAssertEqual(userDefaults.object(forKey: HomeRefreshCoordinator.dateLastRefreshKey) as? Date, date)
+    }
+
+    func test_coordinator_whenDataIsStale_refreshesHome() {
+        let date = Calendar.current.date(byAdding: .hour, value: -2, to: Date())
```
dskuza, line 67: Should this be -2, or -12?

PocketKit/Tests/PocketKitTests/Refresh/HomeRefreshCoordinatorTests.swift
```diff
@@ -0,0 +1,134 @@
...
+        let expectRefresh = expectation(description: "Refresh home")
+        coordinator.refresh {
+            expectRefresh.fulfill()
+        }
+        wait(for: [expectRefresh], timeout: 1)
+        XCTAssertNotEqual(userDefaults.object(forKey: HomeRefreshCoordinator.dateLastRefreshKey) as? Date, date)
+    }
+
+    func test_coordinator_whenForceRefresh_refreshesHome() {
+        XCTAssertNil(userDefaults.object(forKey: HomeRefreshCoordinator.dateLastRefreshKey))
```
dskuza, line 83: Do we want this to be nil? I feel like the code path here is if that there _is_ no last date, then we're not forcing it, we're just hitting the case that we've never been refreshed. In my mind, if we set this value to `Date()`, _then_ attempt a force-refresh, we should refresh (since we're within the 12 hours, still).

PocketKit/Tests/PocketKitTests/Refresh/HomeRefreshCoordinatorTests.swift
```diff
@@ -0,0 +1,134 @@
...
+        let coordinator = subject()
+
+        coordinator.refresh(isForced: true) { }
+        wait(for: [fetchExpectation], timeout: 2)
+
+        XCTAssertEqual(source.fetchSlateLineupCall(at: 0)?.identifier, "e39bc22a-6b70-4ed2-8247-4b3f1a516bd1")
+    }
+
+    func test_coordinator_whenEnterForeground_whenDataIsNotStale_doesNotRefreshHome() {
+
```
dskuza, line 108: Extra newline 😉

PocketKit/Tests/PocketKitTests/Refresh/HomeRefreshCoordinatorTests.swift
dskuza: I wonder if this is some sort of object lifecycle thing. Having an explicit variable definitely keeps it around for the life of this test, so I'm wondering if `_` removes the object from memory immediately (thus, nothing to receive the notification).

dskuza (commented):
@cyndichin Looks good so far. Just some food for thought. Almost at approval, I just wanted to open up some discussions.

PocketKit/Tests/PocketKitTests/Refresh/HomeRefreshCoordinatorTests.swift
cyndichin: yes, you are right! good catch! I set the date up to be the current date, so that a forced refresh will call the refresh button even if within the 12 hours.

PocketKit/Tests/PocketKitTests/Refresh/HomeRefreshCoordinatorTests.swift
cyndichin: Ah, I set the `refreshRuleTime` in the tests to be 1 instead of 12, but I can see how that can be confusing, so I'll update it to 12.

PocketKit/Sources/PocketKit/Refresh/HomeRefreshCoordinator.swift
cyndichin: Sounds good to me! I think it's valuable to understand how often the refresh occurs. Does this make sense for the breadcrumb details? 

`Crashlogger.breadcrumb(category: "refresh", level: .info, message: "Home Refresh Occur")`

PocketKit/Tests/PocketKitTests/Refresh/HomeRefreshCoordinatorTests.swift
cyndichin: Using the memory debugger, it confirms that it isn't kept around, so you're right. Wish Xcode didn't leave the warning for it though.
image

```
Initialization of immutable value 'coordinator' was never used; consider replacing with assignment to '_' or removing it
```

PocketKit/Sources/PocketKit/Refresh/HomeRefreshCoordinator.swift
cyndichin: updated! I like making it more flexible - thanks for the alternative 💯

PocketKit/Sources/PocketKit/Refresh/HomeRefreshCoordinator.swift
dskuza: That works for me!

dskuza (approved):

PR #312, opened 2022-11-03
feat(migration): migrate legacy user sessions to new session type

dskuza:
## Summary

As required, performs a migration of a legacy user session (guid, access token, and user identifier) from the legacy session store to the new keychain-based store.

## References 

IN-845

## Implementation Details

There are two main portions to this implementation:
1. Verify that the application _needs_ to run the migration
    - If the user's last known application version was < 8.x.x, and the migration had not yet been run
2. Perform the migration, either in the main app or the save extension.

For obtaining the previous version _as well as_ the legacy session information, we have to dig into an encrypted JSON file, stored in a shared directory (so that both the extension and application can read from it). We use `RNCryptor` to decrypt the data (from the JSON file), using a password stored in `UserDefaults`. To simplify deserialization, we create a `LegacyStore` type that represents the data we want to decrypt (specifically for that migration). 

We have to attempt to run the migration in two spots:
1. The main app
2. The save extension

The reason for also attempting to run within the save extension is that there is a possibility that the user updated the app (either explicitly or via automatic update), and opened the save extension prior to opening the main app. This way, one migration is synchronized across both targets.

## Test Steps

???

## PR Checklist:

- [x] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/PocketKit/PocketAppDelegate.swift
+++ b/PocketKit/Sources/PocketKit/PocketAppDelegate.swift
@@ -81,6 +81,23 @@ public class PocketAppDelegate: UIResponder, UIApplicationDelegate {
         }
         Textiles.initialize()
 
+        do {
+            let migration = LegacyUserMigration(
+                userDefaults: userDefaults,
+                encryptedStore: PocketEncryptedStore(),
+                appSession: appSession
+            )
+
+            let attempted = try migration.perform()
+            if attempted {
+                Crashlogger.breadcrumb(category: "launch", level: .info, message: "Legacy user migration required; running.")
+            } else {
+                Crashlogger.breadcrumb(category: "launch", level: .info, message: "Legacy user migration not required; skipped.")
+            }
+        } catch {
+            Crashlogger.capture(error: error)
+        }
+
         return true
     }
 
--- a/PocketKit/Sources/PocketKit/PocketAppDelegate.swift
+++ b/PocketKit/Sources/PocketKit/PocketAppDelegate.swift
@@ -81,20 +81,27 @@ public class PocketAppDelegate: UIResponder, UIApplicationDelegate {
         }
         Textiles.initialize()
 
-        do {
-            let migration = LegacyUserMigration(
-                userDefaults: userDefaults,
-                encryptedStore: PocketEncryptedStore(),
-                appSession: appSession
-            )
+        let legacyUserMigration = LegacyUserMigration(
+            userDefaults: userDefaults,
+            encryptedStore: PocketEncryptedStore(),
+            appSession: appSession
+        )
 
-            let attempted = try migration.perform()
+        do {
+            let attempted = try legacyUserMigration.perform()
             if attempted {
                 Crashlogger.breadcrumb(category: "launch", level: .info, message: "Legacy user migration required; running.")
             } else {
                 Crashlogger.breadcrumb(category: "launch", level: .info, message: "Legacy user migration not required; skipped.")
             }
+        } catch LegacyUserMigrationError.missingStore {
+            Crashlogger.breadcrumb(category: "launch", level: .info, message: "No previous store for user migration; skipped.")
+            // Since we don't have a store, we can skip any further attempts at running this migration.
+            legacyUserMigration.forceSkip()
         } catch {
+            // All errors are something we can't resolve client-side, so we don't want to re-attempt
+            // on further launches.
+            legacyUserMigration.forceSkip()
             Crashlogger.capture(error: error)
         }
 
--- a/PocketKit/Sources/PocketKit/PocketAppDelegate.swift
+++ b/PocketKit/Sources/PocketKit/PocketAppDelegate.swift
@@ -84,25 +84,26 @@ public class PocketAppDelegate: UIResponder, UIApplicationDelegate {
         let legacyUserMigration = LegacyUserMigration(
             userDefaults: userDefaults,
             encryptedStore: PocketEncryptedStore(),
-            appSession: appSession
+            appSession: appSession,
+            groupID: Keys.shared.groupID
         )
 
         do {
             let attempted = try legacyUserMigration.perform()
             if attempted {
-                Crashlogger.breadcrumb(category: "launch", level: .info, message: "Legacy user migration required; running.")
+                Log.breadcrumb(category: "launch", level: .info, message: "Legacy user migration required; running.")
             } else {
-                Crashlogger.breadcrumb(category: "launch", level: .info, message: "Legacy user migration not required; skipped.")
+                Log.breadcrumb(category: "launch", level: .info, message: "Legacy user migration not required; skipped.")
             }
         } catch LegacyUserMigrationError.missingStore {
-            Crashlogger.breadcrumb(category: "launch", level: .info, message: "No previous store for user migration; skipped.")
+            Log.breadcrumb(category: "launch", level: .info, message: "No previous store for user migration; skipped.")
             // Since we don't have a store, we can skip any further attempts at running this migration.
             legacyUserMigration.forceSkip()
         } catch {
             // All errors are something we can't resolve client-side, so we don't want to re-attempt
             // on further launches.
             legacyUserMigration.forceSkip()
-            Crashlogger.capture(error: error)
+            Log.capture(error: error)
         }
 
         return true
--- a/PocketKit/Sources/SharedPocketKit/Migrations/LegacyUserMigration.swift
+++ b/PocketKit/Sources/SharedPocketKit/Migrations/LegacyUserMigration.swift
@@ -1,0 +1,121 @@
+import Foundation
+
+public enum LegacyUserMigrationError: Error, LocalizedError {
+    case missingKey
+    case missingStore
+    case failedDecryption(Error)
+    case missingData
+    case failedDeserialization(Error)
+
+    public var errorDescription: String? {
+        switch self {
+        case .missingKey: return "Missing decryption key for legacy user migration"
+        case .missingStore: return "Missing file for legacy store"
+        case .failedDecryption(let error): return "Failed to decrypt store: \(error)"
+        case .missingData: return "Data is missing after decrypting store"
+        case .failedDeserialization(let error): return "Failed to deserialize data: \(error)"
+        }
+    }
+}
+
+public class LegacyUserMigration {
+    static let migrationKey = "com.mozilla.pocket.next.migration.legacyUser"
+    static let versionKey = "version"
+    static let decryptionKey = "kPKTCryptoKey"
+
+    private let userDefaults: UserDefaults
+    private let encryptedStore: LegacyEncryptedStore
+    private let appSession: AppSession
+
+    public init(
+        userDefaults: UserDefaults,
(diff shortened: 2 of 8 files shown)


PocketKit/Sources/SharedPocketKit/Migrations/LegacyUserMigration.swift
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    public init(userDefaults: UserDefaults, encryptedStore: LegacyEncryptedStore, appSession: AppSession) {
+        self.userDefaults = userDefaults
+        self.encryptedStore = encryptedStore
+        self.appSession = appSession
+    }
+
+    public func perform() throws {
+        guard isRequired == true else {
```
bassrock, line 41: @dskuza we should put analytics in this code path, that way we can tell when its ok to remove all this.

PocketKit/Sources/PocketKit/PocketAppDelegate.swift
```diff
@@ -80,6 +80,23 @@ public class PocketAppDelegate: UIResponder, UIApplicationDelegate {
...
+
+            let attempted = try migration.perform()
+            if attempted {
+                Crashlogger.breadcrumb(category: "launch", level: .info, message: "Legacy user migration required; running.")
+            } else {
+                Crashlogger.breadcrumb(category: "launch", level: .info, message: "Legacy user migration not required; skipped.")
+            }
+        } catch {
+            Crashlogger.capture(error: error)
+        }
```
cyndichin, line 98: should we consider putting this piece in `SharedPocketKit` since we are using it for both the app and extension?

cyndichin, line 98: can we bypass performing the migration early on by checking if `version was < 8.x.x and the migration has not been run here`?

cyndichin (commented):
Looks good and super clean!! How do I test this or is this to be tested at a later date? I have some comments to better understand what's going on - thanks! Also, love that the tests help me understand the flow better 💯

nzeltzer:
@dskuza, if I'm reading this right, in the legacy app, all of the data necessary to continue a user session were stored in the encrypted json blob on disk, and not in the keychain.

Is that right? 🧐

PocketKit/Sources/SharedPocketKit/Migrations/LegacyUserMigration.swift
```diff
@@ -0,0 +1,121 @@
...
+        self.userDefaults = userDefaults
+        self.encryptedStore = encryptedStore
+        self.appSession = appSession
+    }
+
+    @discardableResult
+    public func perform() throws -> Bool {
+        guard let password = userDefaults.string(forKey: Self.decryptionKey) else {
+            throw LegacyUserMigrationError.missingKey
+        }
```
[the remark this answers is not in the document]
dskuza: @cyndichin in the legacy app :)

PocketKit/Sources/PocketKit/PocketAppDelegate.swift
dskuza: The (current) problem is `version` is read at decryption-time, which is when `perform` is called. I think this can be refactored, though.

PocketKit/Sources/SharedPocketKit/Migrations/LegacyUserMigration.swift
```diff
@@ -0,0 +1,121 @@
...
+
+        let legacyStore: LegacyStore
+        do {
+            legacyStore = try JSONDecoder().decode(LegacyStore.self, from: decryptedData)
+        } catch {
+            throw LegacyUserMigrationError.failedDeserialization(error)
+        }
+
+        guard isRequired(version: legacyStore.version) else {
+            updateUserDefaults()
```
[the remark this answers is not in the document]
dskuza: If we're not required to run, then we update user defaults to mark that it has been "complete", i.e further attempts at running this migration will be skipped.

PocketKit/Tests/SharedPocketKitTests/LegacyUserMigrationTests.swift
```diff
@@ -0,0 +1,220 @@
...
+        }
+
+        XCTAssertNoThrow(try migration.perform())
+    }
+
+    func test_perform_withMissingKey_throwsError() {
+        let migration = subject()
+
+        do {
+            try migration.perform()
```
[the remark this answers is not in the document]
dskuza: Probably!

PocketKit/Tests/SharedPocketKitTests/LegacyUserMigrationTests.swift
```diff
@@ -0,0 +1,220 @@
...
+        migration = subject()
+        XCTAssertFalse(migration.isRequired(version: "9.0.0"))
+    }
+
+    func test_isRequired_withPreviousVersionGreaterThanOrEqualTo8_andRun_returnsFalse() {
+        userDefaults.set(true, forKey: LegacyUserMigration.migrationKey)
+
+        var migration = subject()
+        XCTAssertFalse(migration.isRequired(version: "8.0.0"))
+    }
```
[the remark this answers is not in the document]
dskuza: Correct, but for the sake of testing, I added this test.

PocketKit/Tests/SharedPocketKitTests/LegacyUserMigrationTests.swift
```diff
@@ -0,0 +1,220 @@
...
+extension LegacyUserMigrationTests {
+    func test_isRequired_withPreviousVersionLessThan8_andNotRun_returnsTrue() {
+        let migration = subject()
+        XCTAssertTrue(migration.isRequired(version: "7.0.0"))
+    }
+
+    func test_isRequired_withPreviousVersionLessThan8_andRun_returnsFalse() {
+        userDefaults.set(true, forKey: LegacyUserMigration.migrationKey)
+        let migration = subject()
+        XCTAssertFalse(migration.isRequired(version: "7.0.0"))
```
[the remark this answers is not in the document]
dskuza: Correct!

PocketKit/Tests/SharedPocketKitTests/LegacyUserMigrationTests.swift
```diff
@@ -0,0 +1,220 @@
...
+    ) -> LegacyUserMigration {
+        return LegacyUserMigration(
+            userDefaults: userDefaults ?? self.userDefaults,
+            encryptedStore: encryptedStore ?? self.encryptedStore,
+            appSession: appSession ?? self.appSession
+        )
+    }
+
+    override func setUp() {
+        userDefaults = UserDefaults(suiteName: "LegacyUserMigrationTests")
```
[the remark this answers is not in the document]
dskuza: It's a good way of ensuring we encapsulate user defaults into a single suite, and can reset the entire set of defaults in one pass!

dskuza:
> @dskuza, if I'm reading this right, in the legacy app, all of the data necessary to continue a user session were stored in the encrypted json blob on disk, and not in the keychain.

Is that right? 🧐

@nzeltzer correct!

cyndichin (approved):
looking great from code review standpoint! resolved most of the comments, except the two minor comments about test (non blockers)!

PocketKit/Sources/PocketKit/PocketAppDelegate.swift
```diff
@@ -80,6 +80,30 @@ public class PocketAppDelegate: UIResponder, UIApplicationDelegate {
...
+        } catch LegacyUserMigrationError.missingStore {
+            Crashlogger.breadcrumb(category: "launch", level: .info, message: "No previous store for user migration; skipped.")
+            // Since we don't have a store, we can skip any further attempts at running this migration.
+            legacyUserMigration.forceSkip()
+        } catch {
+            // All errors are something we can't resolve client-side, so we don't want to re-attempt
+            // on further launches.
+            legacyUserMigration.forceSkip()
+            Crashlogger.capture(error: error)
+        }
```
dskuza, line 105: @bassrock This chunk of code is the same for the main app and the save extension, but I want to make sure of something. This is the flow based on this implementation as-is.

1. Attempt to perform the migration. If it succeeds, continue as normal.
2. Attempt to perform the migration. If there is no store, mark the migration as complete, and continue as normal (the user will be logged out). This is the flow new users or fresh installs will see.
3. Attempt to perform the migration. If there is no store, mark the migration as complete (since we can't resolve this client-site (failed to decrypt, for example)), and continue with the user logged out.

Situation 3 is the one I'm not happiest about, but I believe that it has to be done this way.

Any thoughts?

PocketKit/Sources/SharedPocketKit/Migrations/LegacyUserMigration.swift
```diff
@@ -0,0 +1,176 @@
...
+
+private class LegacyPasswordRetriever {
+    let password: String?
+
+    init(keychain: Keychain) {
+        let query: CFDictionary = [
+            kSecClass: kSecClassGenericPassword,
+            kSecAttrService: "Pocket",
+            kSecAttrAccount: "encryption",
+            kSecAttrAccessGroup: "group.com.ideashower.ReadItLaterProAlphaNeue",
```
dskuza, line 160: Is there a better way to do this, or do we create a ticket as a reminder to update this once we're in the true app space?

nzeltzer: A better way to query the keychain, or a better way to do something else?

dskuza: @nzeltzer thanks for requesting clarification. the "access group" will be different for the final release vs app store. it's basically a reflection of the bundle identifier, prefixed with `group.`

dskuza: @nzeltzer i'd have to test to see if we can retrieve the bundle id since we're technically in a swift package that isn't the app itself. let me try that out.

dskuza: it's also tricky since we potentially run the migration from the extension, which has a different bundle ID than the app, and we are required to use specific access groups.

dskuza: It looks like `Bundle.main` when called in the migration returns the correct bundle ID, so we might be able to generate this dynamically.

Gio2018 (changes requested):
Great groundwork for user migration @dskuza !!
I did test it against the production app (by installing that one and then this one on top of it) and I was signed out after upgrading.
It seems that when trying to retrieve the password from the keychain, an `errSecMissingEntitlement` (`-34018`) is returned.
Here's a video that described what I tried (don't mind the time gap between installation and login screen, it was a leftover breakpoint 😄 )

<p align=center>


</p>

dskuza:
@Gio2018 I think I resolved the issue - looks like I was just fetching the decryption key from the wrong spot. Can you test again? I tested against the app store version by both launching the app to kickstart the migration, or running the save extension first to kickstart the migration.

Gio2018 (approved):
Works great @dskuza !

PR #316, opened 2022-11-04
feat(appbadge): App Badge

timc-mozilla:
## Summary
* Add app badge for number of saved articles

## References 
* [IN-981](https://getpocket.atlassian.net/jira/software/projects/IN/boards/80?selectedIssue=IN-981)

## Implementation Details
* Added app badges

## Test Steps
* Go to settings
* turn on app badge
* go to home of iphone and check to see if app badges are showing
* turn off app badges in settings

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/PocketKit/AppBadgeTracker.swift
+++ b/PocketKit/Sources/PocketKit/AppBadgeTracker.swift
@@ -1,0 +1,71 @@
+import Foundation
+import Sync
+import UIKit
+import Combine
+
+protocol BadgeProvider: AnyObject {
+    var applicationIconBadgeNumber: Int { get set }
+}
+
+extension UIApplication: BadgeProvider { }
+
+class AppBadgeSetup {
+
+    private let source: Source
+    private let notificationCenter: NotificationCenter
+    private var subscriptions: Set<AnyCancellable> = []
+    private var userDefaults: UserDefaults
+    private let badgeProvider: BadgeProvider
+    /// This completion block is called once the badge value is updated. This completion block is not currently used in the app code, but is utilized in tests, since setting the badge needs to occur on the main thread (when using UIApplication as the provider), which is called asynchronously. Thus, this is added as async test support.
+    private let completion: (() -> Void)?
+
+    init(
+        source: Source,
+        userDefaults: UserDefaults,
+        notificationCenter: NotificationCenter = .default,
+        badgeProvider: BadgeProvider,
+        completion: (() -> Void)? = nil
+    ) {
+        self.source = source
+        self.notificationCenter = notificationCenter
+        self.userDefaults = userDefaults
+        self.badgeProvider = badgeProvider
+        self.completion = completion
+
+        setupNotificationSubscription()
+    }
+
+    private func setupNotificationSubscription() {
+        self.notificationCenter
+            .publisher(for: .listUpdated)
+            .sink { [weak self] _ in
+                self?.manualCheckForSavedCount()
+            }
+            .store(in: &subscriptions)
+    }
+
+    func manualCheckForSavedCount() {
+        let numberOfSavesRequest = Requests.fetchSavedItems()
+        var numberOfSaves: Int
+        let currentValue = userDefaults.bool(forKey: AccountViewModel.ToggleAppBadgeKey)
+        if currentValue == false {
+            numberOfSaves = 0
+        } else {
+            do {
+                numberOfSaves = try source.mainContext.fetch(numberOfSavesRequest).count
+                print(numberOfSaves)
+            } catch {
+                numberOfSaves = 0
+            }
+        }
+
+        updateBadgeValue(numberOfSaves: numberOfSaves)
+    }
+
+    private func updateBadgeValue(numberOfSaves: Int) {
+        DispatchQueue.main.async { [weak self] in
+            self?.badgeProvider.applicationIconBadgeNumber = numberOfSaves
+            self?.completion?()
+        }
+    }
+}
--- a/PocketKit/Sources/PocketKit/MyList/SavedItemsList/SavedItemsListViewModel.swift
+++ b/PocketKit/Sources/PocketKit/MyList/SavedItemsList/SavedItemsListViewModel.swift
@@ -59,14 +59,16 @@ class SavedItemsListViewModel: NSObject, ItemsListViewModel {
 
     private var selectedFilters: Set<ItemsListFilter>
     private let availableFilters: [ItemsListFilter]
+    private let notificationCenter: NotificationCenter
 
-    init(source: Source, tracker: Tracker, listOptions: ListOptions) {
+    init(source: Source, tracker: Tracker, listOptions: ListOptions, notificationCenter: NotificationCenter) {
         self.source = source
         self.tracker = tracker
         self.selectedFilters = [.all]
         self.availableFilters = ItemsListFilter.allCases
         self.itemsController = source.makeItemsController()
         self.listOptions = listOptions
+        self.notificationCenter = notificationCenter
 
         super.init()
 
@@ -531,7 +533,6 @@ extension SavedItemsListViewModel: SavedItemsControllerDelegate {
         guard .update == type else {
             return
         }
-
         var snapshot = buildSnapshot()
         snapshot.reloadItems([ItemsListCell<ItemIdentifier>.item(savedItem.objectID)])
         _snapshot = snapshot
@@ -539,6 +540,7 @@ extension SavedItemsListViewModel: SavedItemsControllerDelegate {
 
     func controllerDidChangeContent(_ controller: SavedItemsController) {
         itemsLoaded()
+        notificationCenter.post(name: .listUpdated, object: nil)
     }
 }
 
--- a/PocketKit/Sources/PocketKit/Settings/AccountViewModel.swift
+++ b/PocketKit/Sources/PocketKit/Settings/AccountViewModel.swift
@@ -6,11 +6,19 @@ import SharedPocketKit
 import SwiftUI
 
 class AccountViewModel: ObservableObject {
+    static let ToggleAppBadgeKey = "AccountViewModel.ToggleAppBadge"
     private let appSession: AppSession
     private let user: User
+    private let userDefaults: UserDefaults
+    private let notificationCenter: NotificationCenter
 
-    init(appSession: AppSession, user: User) {
+    @AppStorage("Settings.ToggleAppBadge")
+    public var appBadgeToggle: Bool = false
+
+    init(appSession: AppSession, user: User, userDefaults: UserDefaults, notificationCenter: NotificationCenter) {
         self.appSession = appSession
+        self.userDefaults = userDefaults
+        self.notificationCenter = notificationCenter
         self.user = user
     }
 
@@ -19,6 +27,23 @@ class AccountViewModel: ObservableObject {
         NotificationCenter.default.post(name: .userLoggedOut, object: appSession.currentSession)
         user.clear()
         appSession.currentSession = nil
+    }
+
+    func toggleAppBadge() {
+        UNUserNotificationCenter.current().requestAuthorization(options: .badge) {
+            (granted, error) in
+            guard error == nil && granted == true else {
+                self.userDefaults.set(false, forKey: AccountViewModel.ToggleAppBadgeKey)
+                DispatchQueue.main.async { [weak self] in
+                    self?.appBadgeToggle = false
+                }
+                return
+            }
+
+            let currentValue = self.userDefaults.bool(forKey: AccountViewModel.ToggleAppBadgeKey)
+            self.userDefaults.setValue(!currentValue, forKey: AccountViewModel.ToggleAppBadgeKey)
+            self.notificationCenter.post(name: .listUpdated, object: nil)
+        }
     }
 
     @Published var isPresentingHelp = false
--- a/PocketKit/Sources/PocketKit/Settings/SettingsView.swift
+++ b/PocketKit/Sources/PocketKit/Settings/SettingsView.swift
@@ -64,6 +64,12 @@ struct SettingsForm: View {
                 })
                 .textCase(nil)
 
+                Section(header: Text("App Customization").style(.settings.header)) {
+                    SettingsRowToggle(title: "Show App Badge Count", model: model) {
+                        model.toggleAppBadge()
+                    }
+                }.textCase(nil)
+
                 Section(header: Text("About & Support").style(.settings.header)) {
                     SettingsRowButton(title: "Help", icon: SFIconModel("questionmark.circle")) { model.isPresentingHelp.toggle() }
                         .sheet(isPresented: $model.isPresentingHelp) {
@@ -80,7 +86,6 @@ struct SettingsForm: View {
                             SFSafariView(url: URL(string: "https://getpocket.com/en/privacy/")!)
                                 .edgesIgnoringSafeArea(.bottom)
                         }
-
                 }.textCase(nil)
             }
             .listRowBackground(Color(.ui.grey7))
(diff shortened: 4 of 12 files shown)


timc-mozilla:
Flaky test. passing locally.

PocketKit/Sources/PocketKit/Settings/AccountViewModel.swift
```diff
@@ -6,10 +6,15 @@ import SharedPocketKit
 import SwiftUI
 
 class AccountViewModel: ObservableObject {
+    static let ToggleAppBadgeKey = "AccountViewModel.ToggleAppBadge"
```
cyndichin, line 9: style nitpick: lowercase variable name? `toggleAppBadgeKey`

PocketKit/Tests/PocketKitTests/AppBadgeTrackerTests.swift
```diff
@@ -0,0 +1,71 @@
...
+
+    private var source: MockSource!
+    private var userDefaults: UserDefaults!
+    private var accountViewModel: AccountViewModel!
+    private var appBadgeTracker: MockAppBadgeTracker!
+
+    override func setUp() {
+        source = MockSource()
+        userDefaults = UserDefaults()
+        appBadgeTracker = MockAppBadgeTracker(source: source, userDefaults: userDefaults)
```
cyndichin, line 51: curious - if we make changes to the `AppBadgeTracker` class, then would these tests still pass since we are testing the mock vs the actual class?

PocketKit/Sources/PocketKit/Settings/SettingsView.swift
```diff
@@ -64,6 +64,13 @@ struct SettingsForm: View {
                 })
                 .textCase(nil)
 
+                Section(header: Text("App Customization").style(.settings.header)) {
+                    SettingsRowToggle(title: "Show App Badge Count") {
+                        model.isToggleAppBadge.toggle()
```
cyndichin, line 69: do we still need this line?

PocketKit/Sources/PocketKit/AppBadgeTracker.swift
```diff
@@ -0,0 +1,54 @@
+import Foundation
+import Sync
+import UIKit
+import Combine
+
+protocol AppBadgeTracker {
+    func manualCheckForSavedCount()
+}
+
+class AppBadgeSetup : AppBadgeTracker {
```
cyndichin, line 10: style nitpick: looks like we get a colon spacing violation

PocketKit/Sources/PocketKit/Settings/AccountViewModel.swift
```diff
@@ -18,8 +23,22 @@ class AccountViewModel: ObservableObject {
...
+    func toggleAppBadge() {
+        UNUserNotificationCenter.current().requestAuthorization(options: .badge) {
+            (granted, error) in
+            if error == nil && granted == true {
+                let currentValue = self.userDefaults.bool(forKey: AccountViewModel.ToggleAppBadgeKey)
+                self.userDefaults.setValue(!currentValue, forKey: AccountViewModel.ToggleAppBadgeKey)
+                self.notificationCenter.post(name: .listUpdated, object: nil)
+            } else {
+                self.userDefaults.set(false, forKey: "SettingsRowToggle")
+            }
```
cyndichin, line 35: style preference, but using `guard` makes it easier for me to read

```
 guard error == nil && granted == true else {
                self.userDefaults.set(false, forKey: "SettingsRowToggle")
                return
            }

            let currentValue = self.userDefaults.bool(forKey: AccountViewModel.ToggleAppBadgeKey)
            self.userDefaults.setValue(!currentValue, forKey: AccountViewModel.ToggleAppBadgeKey)
            self.notificationCenter.post(name: .listUpdated, object: nil)
 ```
 
Now that I think about it, maybe the post here doesn't make as much sense as we originally thought since the list hasn't been updated, so maybe we can change the name or observe that the userDefaults changed its value. Just a thought to consider not a blocker for me

**Review by cyndichin — COMMENTED** [review]
Looks great and functions well on iPhone / iPad!! I actually really like this feature, can finally see how many saves I have lol. A couple of comments for conversation, but nothing too blocking.

File: PocketKit/Sources/PocketKit/Settings/AccountViewModel.swift (hunk 1 above)
Comment (timc-mozilla) [reply PocketKit/Sources/PocketKit/Settings/AccountViewModel.swift:9]: got it. I often do static global variables be UpperCaseCamel.  I'll change it though

File: PocketKit/Tests/PocketKitTests/AppBadgeTrackerTests.swift (hunk 2 above)
Comment (timc-mozilla) [reply PocketKit/Tests/PocketKitTests/AppBadgeTrackerTests.swift:51]: yup, if AppBadgeTracker changes, we will have to update this mock class.  I was looking at the other `Mock` classes that we have to try and copy the behavior, and they all seem to be similar that they are based around a protocol.  If thats not the case and I misread the code, please let me know and I will change it (but will need help with solution)

File: PocketKit/Sources/PocketKit/MyList/SavedItemsList/SavedItemsListViewModel.swift [hunk 6]
```diff
@@ -529,17 +531,14 @@ extension SavedItemsListViewModel: SavedItemsControllerDelegate {
         for type: NSFetchedResultsChangeType,
         newIndexPath: IndexPath?
     ) {
-        guard .update == type else {
-            return
-        }
-
```
cyndichin, line 535: Do we still want to keep this?

PocketKit/Sources/PocketKit/MyList/SavedItemsList/SavedItemsListViewModel.swift
```diff
@@ -529,17 +531,14 @@ extension SavedItemsListViewModel: SavedItemsControllerDelegate {
...
         var snapshot = buildSnapshot()
         snapshot.reloadItems([ItemsListCell<ItemIdentifier>.item(savedItem.objectID)])
         _snapshot = snapshot
     }
 
     func controllerDidChangeContent(_ controller: SavedItemsController) {
         itemsLoaded()
+        notificationCenter.post(name: .listUpdated, object: nil)
```
cyndichin, line 541: just curious, why is this preferred over the other `didChange` function?

PocketKit/Sources/PocketKit/Settings/AccountViewModel.swift
cyndichin: no worries, im actually not sure what the standard is on this team... so could be just my preference. Not sure if we should standardize this or not

PocketKit/Sources/PocketKit/Settings/SettingsView.swift
cyndichin: Hmm from my understanding, the other values are based on presenting another sheet, but in this case we're just toggling, so might not need it if we already have `appBadgeToggle`.

PocketKit/Sources/PocketKit/MyList/SavedItemsList/SavedItemsListViewModel.swift
dskuza: @cyndichin `didChangeContent` is called one time once all changes (which call the other function) are performed / saved. so, instead of sending n number of notifications (for example if 10 items were added before a save), we only send 1. just a performance thing.

PocketKit/Tests/PocketKitTests/AppBadgeTrackerTests.swift
```diff
@@ -0,0 +1,63 @@
+import XCTest
+@testable import PocketKit
+import Combine
+import SharedPocketKit
+@testable import Sync
```
cyndichin, line 5: 🔥 woo hoo tests!~ Not a blocker, but would like to see the `delete` case as well

PocketKit/Sources/PocketKit/Settings/SettingsView.swift
timc-mozilla: okee dokee artichoke!

PocketKit/Sources/PocketKit/MyList/SavedItemsList/SavedItemsListViewModel.swift
cyndichin: makes sense, thanks for the explanation!

PocketKit/Sources/PocketKit/AppBadgeTracker.swift
```diff
@@ -0,0 +1,70 @@
...
+extension UIApplication: BadgeProvider { }
+
+class AppBadgeSetup {
+
+    private let source: Source
+    private let notificationCenter: NotificationCenter
+    private var subscriptions: Set<AnyCancellable> = []
+    private var userDefaults: UserDefaults
+    private let badgeProvider: BadgeProvider
+    private let completion: (() -> Void)?
```
dskuza, line 19: @timc-mozilla Can you leave a quick comment in code about why we added this? Here's an idea:

```
This completion block is called once the badge value is updated. This completion block is not currently used in the app code, but is utilized in tests, since setting the badge needs to occur on the main thread (when using UIApplication as the provider), which is called asynchronously. Thus, this is added as async test support.
```

PocketKit/Sources/PocketKit/Settings/AccountViewModel.swift
```diff
@@ -6,10 +6,18 @@ import SharedPocketKit
 import SwiftUI
 
 class AccountViewModel: ObservableObject {
+    static let ToggleAppBadgeKey = "AccountViewModel.ToggleAppBadge"
     private let appSession: AppSession
+    private let userDefaults: UserDefaults
+    private let notificationCenter: NotificationCenter
 
-    init(appSession: AppSession) {
+    @AppStorage("settingsRowToggleAppBadge")
```
dskuza, line 14: I'm not sure we have to include "settingsRowToggle" in the key here, as that's just an implementation detail. We should decide on settings key naming, so that we can form a basic pattern. 

As an example, the reader settings currently use `readerFontSizeAdjustment` and `readerFontFamily`, so I'm thinking maybe something along the lines of `settingsAppBadge` or `Settings.appBadge`.

Pinging @cyndichin for additional thoughts.

PocketKit/Sources/PocketKit/Settings/AccountViewModel.swift
```diff
@@ -18,6 +26,23 @@ class AccountViewModel: ObservableObject {
         appSession.currentSession = nil
     }
 
+    func toggleAppBadge() {
+        UNUserNotificationCenter.current().requestAuthorization(options: .badge) {
+            (granted, error) in
+            guard error == nil && granted == true else {
+                self.userDefaults.set(false, forKey: "SettingsRowToggle")
```
dskuza, line 33: Is the key here supposed to be different than the one used by `AppStorage`?

EDIT: I think I know what's going on, but I'll need to verify…

PocketKit/Sources/PocketKit/Settings/AccountViewModel.swift
```diff
@@ -18,6 +26,23 @@ class AccountViewModel: ObservableObject {
...
+                DispatchQueue.main.async { [weak self] in
+                    self?.appBadgeToggle = false
+                }
+                return
+            }
+
+            let currentValue = self.userDefaults.bool(forKey: AccountViewModel.ToggleAppBadgeKey)
+            self.userDefaults.setValue(!currentValue, forKey: AccountViewModel.ToggleAppBadgeKey)
+            self.notificationCenter.post(name: .listUpdated, object: nil)
```
dskuza, line 42: Just a thought: `listUpdated` doesn't seem to fit within the context of exactly what's happening here, but I get why it's used. We _could_ add another notification for this to be more explicit, but… I don't think we have to right now.

dskuza (commented):
@timc-mozilla Looking good so far. I'd still like to see (one day) moving to an `AppSettings` type similar to `ReaderSettings`, but we can chalk that up as tech debt to resolve later, or as we add more settings.

One thing: can you explain the usage of `AppStorage` vs the separate key used to (seemingly) store the actual value of the toggle?

PocketKit/Sources/PocketKit/Settings/AccountViewModel.swift
cyndichin: I think we should keep it consistent with our `UserDefaults` convention, so I think `Settings.appBadge`. I also don't mind `Settings.appBadgeToggle` since it gives insight that it's a true or false value too.

PocketKit/Sources/PocketKit/Settings/AccountViewModel.swift
cyndichin: I had a comment here regarding this too, just curious what are some thoughts about adding a user defaults observer? 
https://github.com/Pocket/pocket-ios/pull/316#discussion_r1015591813

PocketKit/Sources/PocketKit/Settings/AccountViewModel.swift
timc-mozilla: nope your right. oversight on my part.

PocketKit/Sources/PocketKit/Settings/AccountViewModel.swift
timc-mozilla: ill udpate to `Settings.appBadgeToggle`

cyndichin:
> @timc-mozilla Looking good so far. I'd still like to see (one day) moving to an `AppSettings` type similar to `ReaderSettings`, but we can chalk that up as tech debt to resolve later, or as we add more settings.

^ With regards to this, this means that we'll have two ObservableObjects that are being used in `SettingsView` if we're following the `ReaderSettingsView` and `ReaderSettings`. Is this ideal or should it only know `AccountViewModel` and maybe we can add `AppSettings` as a property of `AccountViewModel` as opposed to making it an ObservableObject on `SettingsView`.

dskuza:
@cyndichin My reasoning for having two is that I assumed they'd be distinct - reader settings are things like font family, font size, and (later) margins n' such, which are not strictly app related. I don't imagine passing them into the Settings view.

PocketKit/Sources/PocketKit/MyList/SavedItemsList/SavedItemsListViewModel.swift
timc-mozilla: yup, youre right. oversight on me for not adding that back in

PocketKit/Sources/PocketKit/AppBadgeTracker.swift
timc-mozilla: okee dokee

PocketKit/Tests/PocketKitTests/AppBadgeTrackerTests.swift
timc-mozilla: ok

dskuza (approved):
I think this is a good first pass. I think the next time we introduce a setting, we can see how building upon this feels, and make any further changes as necessary. Since we want to get this in sooner-than-later, I think we can move forward with this.

PocketKit/Tests/PocketKitTests/AppBadgeTrackerTests.swift
```diff
@@ -0,0 +1,88 @@
...
+
+        NotificationCenter.default.post(name: .listUpdated, object: nil)
+
+        XCTAssertEqual(badgeProvider.applicationIconBadgeNumber, 0)
+        source.mainContext.delete(savedItem)
+    }
+
+    func test_on_savedItemsUpdated_subscribersCalledAddingElement() {
+        let badgeExpectation = expectation(description: "expected badge count to be updated")
+        let subject = subject() {
```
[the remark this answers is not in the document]
bassrock: @timc-mozilla fyi swiftlint wants the `()` removed

PR #319, opened 2022-11-09
feat(premium): fetch user status [IN-979]

cyndichin:
## Summary
Fetch user premium status on sign-in by using `isPremium` field from GraphQL 

## References 
IN-979

## Implementation Details
* Added `isPremium` field to `userByToken` query that gets called when we are Fetching Saves
* Created new class `User` which stores the users status via @AppStorage
* Pass in the `user` to our `FetchList` in order to `setPremiumStatus`
* Add tests and create mocks for user class

## Test Steps
- [ ] WHEN I sign into the app,
THEN request my premium status
THEN store the premium status

- [ ] WHEN the application is launched 
AND a sync request is made (including for user’s premium status)
THEN store the premium status

- [ ] WHEN the user quits the application after launching
THEN the premium status should persist without another request

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/PocketKit/Home/HomeViewModel.swift
+++ b/PocketKit/Sources/PocketKit/Home/HomeViewModel.swift
@@ -131,7 +131,7 @@ class HomeViewModel {
 
         networkPathMonitor.updateHandler = { [weak self] path in
             if path.status == .satisfied {
-                self?.refresh { }
+                self?.refresh(isForced: true) { }
             }
         }
     }
--- a/PocketKit/Sources/SharedPocketKit/User.swift
+++ b/PocketKit/Sources/SharedPocketKit/User.swift
@@ -1,0 +1,32 @@
+import Foundation
+import SwiftUI
+
+public enum Status: String {
+    case premium = "premium"
+    case free = "free"
+    case unknown = "unknown"
+}
+
+public protocol User {
+    func setPremiumStatus(_ isPremium: Bool)
+    func clear()
+}
+
+public class PocketUser: User {
+    static let userStatusKey = "User.statusKey"
+
+    @AppStorage
+    public var status: Status?
+
+    public init(status: Status = .unknown, userDefaults: UserDefaults) {
+        _status = AppStorage(Self.userStatusKey, store: userDefaults)
+    }
+
+    public func setPremiumStatus(_ isPremium: Bool) {
+        status = isPremium ? .premium : .free
+    }
+
+    public func clear() {
+        status = .unknown
+    }
+}
--- a/PocketKit/Sources/Sync/Operations/FetchList.swift
+++ b/PocketKit/Sources/Sync/Operations/FetchList.swift
@@ -2,8 +2,10 @@ import Foundation
 import Apollo
 import Combine
 import PocketGraph
+import SharedPocketKit
 
 class FetchList: SyncOperation {
+    private let user: User
     private let token: String
     private let apollo: ApolloClientProtocol
     private let space: Space
@@ -13,6 +15,7 @@ class FetchList: SyncOperation {
     private let lastRefresh: LastRefresh
 
     init(
+        user: User,
         token: String,
         apollo: ApolloClientProtocol,
         space: Space,
@@ -21,6 +24,7 @@ class FetchList: SyncOperation {
         maxItems: Int,
         lastRefresh: LastRefresh
     ) {
+        self.user = user
         self.token = token
         self.apollo = apollo
         self.space = space
@@ -61,6 +65,9 @@ class FetchList: SyncOperation {
 
         repeat {
             let result = try await fetchPage(pagination)
+            if let isPremium = result.data?.userByToken?.isPremium as? Bool {
+                user.setPremiumStatus(isPremium)
+             }
 
             if case .started = initialDownloadState.value,
                let totalCount = result.data?.userByToken?.savedItems?.totalCount,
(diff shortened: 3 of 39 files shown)


PocketKit/Sources/PocketKit/Home/HomeViewModel.swift
```diff
@@ -131,7 +131,7 @@ class HomeViewModel {
 
         networkPathMonitor.updateHandler = { [weak self] path in
             if path.status == .satisfied {
-                self?.refresh { }
+                self?.refresh(isForced: true) { }
```
cyndichin, line 134: I missed this during the home refresh work - I believe this should be a force refresh since it should only occur when network connectivity change

PocketKit/Sources/Sync/Operations/FetchList.swift
```diff
@@ -95,7 +100,7 @@ class FetchList: SyncOperation {
 
     private func fetchPage(_ pagination: PaginationSpec) async throws -> GraphQLResult<FetchSavesQuery.Data> {
         let query = FetchSavesQuery(
-            token: token,
+            token: appSession?.session?.accessToken ?? "",
```
cyndichin, line 103: at this point the token should have a value but because we're passing in the appSession and not token, have to do this check here - let me know if there's a better way!

PocketKit/Sources/SharedPocketKit/Session/AppSession.swift
```diff
@@ -1,10 +1,21 @@
...
 
+public enum UserStatus {
+    case premium
+    case free
+}
+
 public class AppSession {
     @KeychainStorage
     public var currentSession: Session?
+    public var currentStatus: UserStatus?
```
dskuza, line 11: Can we just default to `free` and update as necessary?

dskuza (commented):
@cyndichin Quick question - it doesn't look like we're persisting the user status. Will we lose it if the app is quit until next time a fetch is made? I want to give it another pass, but this stuck out in my mind immediately.

PocketKit/Sources/SharedPocketKit/Session/AppSession.swift
cyndichin: I wanted to `nil` it out when the user is logged out, but I can have it default to `free`

cyndichin:
> @cyndichin Quick question - it doesn't look like we're persisting the user status. Will we lose it if the app is quit until next time a fetch is made? I want to give it another pass, but this stuck out in my mind immediately.

Oh yes, you're right. Were you thinking persistence through user defaults? Also, using the app session is a temporary solution for now right, do you foresee using a User object with regards to the Premium work?

PocketKit/Sources/SharedPocketKit/Session/AppSession.swift
dskuza: @cyndichin makes sense; similar thought process to `currentSession` then. an alternative is adding another `loggedOut` case, but I'll leave that decision up to you.

dskuza:
> Oh yes, you're right. Were you thinking persistence through user defaults? Also, using the app session is a temporary solution for now right, do you foresee using a User object with regards to the Premium work?

Well, we _could_ use `@AppStorage`, or alternatively, we can also add it to `session`, and have the concept of a "free" session or a "premium" session, but that does kind of take away from the concept of a User.

We could also just introduce the concept of a User now. 

Either way, I think the next step before having this pull request fully ready for review is adding persistence to the user status!

PocketKit/Sources/SharedPocketKit/Session/AppSession.swift
cyndichin: I decided to make an `unknown` case just since the user may be logged in but the Fetch Saves query was not completed.

cyndichin:
> > Oh yes, you're right. Were you thinking persistence through user defaults? Also, using the app session is a temporary solution for now right, do you foresee using a User object with regards to the Premium work?
> 
> Well, we _could_ use `@AppStorage`, or alternatively, we can also add it to `session`, and have the concept of a "free" session or a "premium" session, but that does kind of take away from the concept of a User.
> 
> We could also just introduce the concept of a User now.
> 
> Either way, I think the next step before having this pull request fully ready for review is adding persistence to the user status!

Thanks! I do like the concept of a User and that it's separate from the session. I updated the PR and squashed commits since its pretty different than what I had originally. Will put my PR in ready for review after I rebase the GraphQL issue PR.

cyndichin (commented):
Curious to know why this `plist` file got generated? 
`PocketKit/Tests/SharedPocketKitTests/UserTests.swift.plist`. I removed it, but was just wondering why it appeared and if anyone faced this when creating a new class.

image

dskuza (approved):
😎

cyndichin:
> Curious to know why this `plist` file got generated? `PocketKit/Tests/SharedPocketKitTests/UserTests.swift.plist`. I removed it, but was just wondering why it appeared and if anyone faced this when creating a new class.
> 
> image

This is because I was using #file instead of a string.

PocketKit/Sources/SharedPocketKit/User.swift
```diff
@@ -0,0 +1,32 @@
+import Foundation
+import SwiftUI
+
+public enum Status: String {
+    case premium = "premium"
+    case free = "free"
+    case unknown = "unknown"
+}
+
+public protocol UserProtocol {
```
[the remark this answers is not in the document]
dskuza: Basically, and just more readable (to me). It kind of feels like the equivalent of doing something like 

`class ViewModelClass { }`

which just doesn't seem right. 

We know it's a protocol because it's defined as such, so there's no need to add the suffix. That's my thinking, at least.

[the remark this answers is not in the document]
cyndichin: makes sense, thanks! will do this going forward

PR #320, opened 2022-11-09
build(graphql): generate updated schema

dskuza:
## Summary

Update the GraphQL schema and auto-generated code to include changes to the `getSlateLineup` and `getSlate` queries. This change made the return type of these queries non-optional, which we then had to reflect in our usage of these queries in `Sync`.

## Implementation Details
 
The commands (in our README) to redownload and regenerate code based on our GraphQL schema were called.


cyndichin (approved):
Just for more clarity, was Bitrise failing because it downloads and generates its on API.swift files as opposed to using the local ones?

dskuza:
@cyndichin yep - which i'm not opposed to. we should strive to use the latest graph (unless we see a reason why we shouldn't)

dskuza:
@cyndichin well, @bassrock is that correct? CI trumps the local files?

bassrock:
@dskuza yea that's correct. Eventually, a developer would have had the same issue, doing it via CI is to catch something like this sooner than later.

timc-mozilla (approved):

PR #322, opened 2022-11-16
ci(swiftlint): add custom rule

cyndichin:
## Summary
Draft PR to demonstrate custom rule. Included an example. Issue with this is that it triggers even in cases when we want it to be explicit, but we can always disable this rule for this cases. Open for team discussion!

Custom Rule is related to this:
https://github.com/Pocket/pocket-ios/wiki/Pocket-Swift-Style-Guide#internal-by-default

## References 
N / A

## Test Steps
Pull down this branch and confirm that the warning appears when explicitly adding internal to a function. 

## Screenshots
image

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -7,7 +7,6 @@ disabled_rules: # rule identifiers to exclude from running
   # Disabled all default rules
   - block_based_kvo
   - closure_parameter_position
-  - custom_rules
   - cyclomatic_complexity
   - deployment_target
   - discouraged_direct_init
@@ -90,6 +89,13 @@ opt_in_rules: # some rules are only opt-in
   - return_arrow_whitespace
   - unused_import
   - yoda_condition
+
+custom_rules:
+  redundant_internal_acl:
+    name: "Redundant Internal Access Level"
+    regex: "internal func"
+    message: "Internal by default; no need to explicitly mark it as such"
+    severity: warning
 
 excluded: # paths to ignore during linting. Takes precedence over `included`.
   - build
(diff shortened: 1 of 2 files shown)


cyndichin (commented):
Bitrise is failing because it does not pass Danger - this is intentional as wanted to demonstrate the example of the custom rule

cyndichin:
Attempted to use regex where we can detect if the violation is within a class vs public class, but this causes long build times + couldn't validate if this actually works in our project. Therefore, I think to avoid complexity, we should prob keep the custom rule simple and add the disable SwiftLint comment or keep ourselves honest. Open to suggestions! 
image

dskuza:
> Update: an alternative is to question why we want this rule when it goes against the explicit act rule here

I think it's just bloat; we already let the compiler infer things: `let foo = "foo"` vs `let foo: String = "foo"`.

cyndichin:
> > Update: an alternative is to question why we want this rule when it goes against the explicit act rule here
> 
> I think it's just bloat; we already let the compiler infer things: `let foo = "foo"` vs `let foo: String = "foo"`.

Yeah, I agree and I added it to our wiki in terms of style guide, but I don't think we should be adding a custom rule for this currently as we'll need to consider cases where we want it explicitly. We can focus on integrated rules that have been proposed on the Mozilla's slack channel and once we see more of a need in the future, we can revisit.

cyndichin:
> > > Update: an alternative is to question why we want this rule when it goes against the explicit act rule here
> > 
> > 
> > I think it's just bloat; we already let the compiler infer things: `let foo = "foo"` vs `let foo: String = "foo"`.
> 
> Yeah, I agree and I added it to our wiki in terms of style guide, but I don't think we should be adding a custom rule for this currently as we'll need to consider cases where we want it explicitly. We can focus on integrated rules that have been proposed on the Mozilla's slack channel and once we see more of a need in the future, we can revisit.

Closing ticket per slack comment here: https://pocket.slack.com/archives/CBKLKAUTC/p1671033379144289
Can re-open as needed.

PR #323, opened 2022-11-16
ci(tests): a few simple sign-out tests

audie1:
Initial Audie commit.

I had to write these in a slightly sloppy fashion, in time I'd like to clean them up when I get a better handle on how to interact with elements in the Account Menu - once the menu has been updated.

Update: Cyndi fixed the element in the menu so I can click on it now with an a.i. much cleaner! Thanks Cyndi :)


Tests iOS/SignOutTests.swift
```diff
@@ -44,4 +44,35 @@ class SignOutTests: XCTestCase {
...
         app.terminate()
     }
+    
+    func test_tappingSignOutshowsLogin() {
+        app.tabBar.settingsButton.wait().tap()
+        tap_SignOut()
+        
+        let account = XCUIApplication()
+        account.alerts["Are you sure?"].scrollViews.otherElements.buttons["Sign Out"].wait().tap()
+        app.accountView.loginText.exists
```
cyndichin, line 54: This can resolve the `Bool is unused` warning sign.
```suggestion
        XCTAssertTrue(app.accountView.loginText.exists)
```

Tests iOS/SignOutTests.swift
```diff
@@ -44,4 +44,35 @@ class SignOutTests: XCTestCase {
...
+    
+    func tap_SignOut() {
+        let account = XCUIApplication()
+        let cellCount = account.cells.count
+        XCTAssertTrue(cellCount > 0)
+
+        let firstCell = account.cells.element(boundBy: 1)
+        XCTAssertTrue(firstCell.exists)
+        firstCell.tap()
+    }
```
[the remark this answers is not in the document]
cyndichin: Oh it seems we had one before, `app.accountView.signOutButton.tap()` but it no longer works. May be due to writing our Settings screen in SwiftUI

Tests iOS/Support/Elements/TabBarElement.swift
```diff
@@ -18,4 +18,8 @@ struct TabBarElement: PocketUIElement {
     var accountButton: XCUIElement {
         element.buttons["Account"]
     }
+    
+    var settingsButton: XCUIElement {
+        element/*@START_MENU_TOKEN@*/.buttons["account-tab-bar-button"]/*[[".buttons[\"Settings\"]",".buttons[\"account-tab-bar-button\"]"],[[[-1,1],[-1,0]]],[0]]@END_MENU_TOKEN@*/
```
cyndichin, line 23: ohh interesting didn't realize these comments appear - was this generated by the recorder?

Tests iOS/SignOutTests.swift
```diff
@@ -44,4 +44,35 @@ class SignOutTests: XCTestCase {
...
+    
+    func test_tappingSignOutCancelshowsAccountsMenu() {
+        app.tabBar.settingsButton.wait().tap()
+        tap_SignOut()
+        
+        let account = XCUIApplication()
+        account.alerts["Are you sure?"].scrollViews.otherElements.buttons["Cancel"].wait().tap()
+        let cellCount = account.cells.count
+        XCTAssertTrue(cellCount > 0)
```
cyndichin, line 64: we can also use `XCTAssertGreaterThan(cellCount, 0)` but either works!

Tests iOS/SignOutTests.swift
```diff
@@ -44,4 +44,35 @@ class SignOutTests: XCTestCase {
...
+        account.alerts["Are you sure?"].scrollViews.otherElements.buttons["Sign Out"].wait().tap()
+        app.accountView.loginText.exists
+    }
+    
+    func test_tappingSignOutCancelshowsAccountsMenu() {
+        app.tabBar.settingsButton.wait().tap()
+        tap_SignOut()
+        
+        let account = XCUIApplication()
+        account.alerts["Are you sure?"].scrollViews.otherElements.buttons["Cancel"].wait().tap()
```
cyndichin, line 62: I like what you did for the `settingsButton` and `loginText`, should we do the same here?

cyndichin (commented):
looks great for the initial draft! thanks for helping us with these tests :)

Tests iOS/SignOutTests.swift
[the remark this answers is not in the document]
cyndichin: image

in `SettingsView.swift`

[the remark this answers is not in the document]
audie1: Oh I was wondering if there would be some custom accessibilityIdentifier names set for this menu, yes that would be a much much better solution for this.

Tests iOS/SignOutTests.swift
audie1: was wondering what that Bool message was about :)

Tests iOS/SignOutTests.swift
audie1: changed it, I like that suggestion thanks

Tests iOS/Support/Elements/TabBarElement.swift
audie1: Recorder generated that yep, I'll remove those - didn't know that would be pushed up as text like that

Tests iOS/SignOutTests.swift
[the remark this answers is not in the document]
audie1: async is great, I'm curious exactly how I can call that element now in the function that it's got an accessibilityIdentifier

Tests iOS/SignOutTests.swift
audie1: I actually had trouble getting this to work any other way, but would be open to any suggestions for improvement...

When I tried to add element.alerts["Are you sure?"].scrollViews.otherElements.buttons["Cancel"].wait().tap()
into AccountViewElement, I got an error

Tests iOS/SignOutTests.swift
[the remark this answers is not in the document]
cyndichin: We add these identifiers manually sometimes and there's a difference in adding accessibilityIdentifiers for views made from `UIKit` vs `SwiftUI`. The Settings View is done in SwiftUI, so we'll need to append the modifier `.accessibilityIdentifier(<name>)` shown above. I don't think we have a naming convention yet, but maybe we should. 

I realize that our `accountview` identifier is incorrect as it points to our old VC vs the new SwiftUI one. Should I make these changes for your branch?

Tests iOS/SignOutTests.swift
cyndichin: I also realize our `AccountViewElement` is pointing to the wrong place so that could be the reason why, I can make the change on this branch if you would like or we can do this async. Let me know how you prefer to proceed :) Our old implementation was using `AccountViewController` but now we are using `SettingsViewController`

cyndichin (commented):
also just to let you know, we have recently adapted to using conventional commits as a team, so our commits are in the format before we merge in
`type(scope): message description [JIRA-1234]`
https://getpocket.atlassian.net/wiki/spaces/PE/pages/2827092110/Engineering+Standards#Git-Commit-Message-Standards

also a quick way to fix the SwiftLint warnings, you can use `swiftlint --fix` in terminal. I was doing it manually before Daniel told me about it 😆

Tests iOS/SignOutTests.swift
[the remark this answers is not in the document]
audie1: Thanks for looking into this @cyndichin! Sure I support any changes you deem appropriate to address this, I just realized it was really difficult to find an easy way to click on the menu options so any solution you think is best I'll work with that.

Tests iOS/SignOutTests.swift
audie1: Oh interesting, yes please make any changes you would like to the branch yes. 🙌

cyndichin (approved):

PR #331, opened 2022-11-23
feat(featureflag): adding in feature flag baseline

bassrock:
## Summary

Add in the ability to use feature flags in the app.

Usage:

```swift

let assigned = Services.shared.featureFlagService.isAssigned(flag: "temp-ios.test", variant: "variant b")

//then track the user felt it
Services.shared.featureFlagService.trackFeatureFlagFelt(flag: "temp-ios.test", variant: "variant b")

```

## Test Steps
* See feature flags are loaded via the console.
* A debug menu for iOS to come with this capability :) 

## PR Checklist:
- [ ] Added Unit / UI tests
- [ ] Basic Self QA


PocketKit/Sources/Sync/PocketSource.swift
```diff
@@ -72,13 +73,13 @@ public class PocketSource: Source {
...
         user: User,
-        sessionProvider: SessionProvider,
+        appSession: AppSession,
         consumerKey: String,
         defaults: UserDefaults,
         backgroundTaskManager: BackgroundTaskManager
     ) {
         let apollo = ApolloClient.createDefault(
-            sessionProvider: sessionProvider,
+            sessionProvider: appSession as! SessionProvider,
```
bassrock, line 82: This does not need to be force cast in Services, but for some reason it does here. And it works. @dskuza  may have insights.

PocketKit/Sources/Sync/Space/Space.swift
```diff
@@ -37,9 +37,7 @@ public class Space {
     func delete(_ object: NSManagedObject, in context: NSManagedObjectContext? = nil) {
         let context = context ?? backgroundContext
         context.performAndWait {
-            guard let object = backgroundObject(with: object.objectID) else {
-                return
-            }
+            let object = context.object(with: object.objectID)
```
bassrock, line 40: @Gio2018 opened a pr with this separately https://github.com/Pocket/pocket-ios/pull/575

PocketKit/Sources/PocketKit/Refresh/FeatureFlagsRefreshCoordinator.swift
```diff
@@ -0,0 +1,78 @@
...
+import Sync
+import SharedPocketKit
+import Combine
+
+/// Refresh coordinator to handle the refreshing of feature flags
+class FeatureFlagsRefreshCoordinator: RefreshCoordinator {
+
+    let taskID: String = "com.mozilla.pocket.refresh.featureFlags"
+
+    let refreshInterval: TimeInterval? = 60 * 60 * 48 // once every 48 hours.
```
Gio2018, line 15: why setting this as an optional here and then force unwrap it below? Can it just be a constant?

PocketKit/Sources/PocketKit/Refresh/UserRefreshCoordinator.swift
```diff
@@ -34,6 +34,7 @@ class UserRefreshCoordinator: RefreshCoordinator {
     func refresh(isForced: Bool = false, _ completion: @escaping () -> Void) {
         Task {
```
Gio2018, line 35: capturing self explicitly might not be strictly required if we are sure that the task completes (although a weak capture is a failsafe that always ensures that self is not retained more than it should), but curious if there's a specific reason why it's not captured here as opposed to other places where it's captured?

PocketKit/Sources/Sync/PocketSource.swift
Gio2018: it feels like we have protocols, types and extensions defined in different modules that also sometimes depend on each other, that might be the cause of this?

Gio2018 (approved):
Looks good @bassrock ! Just a few non blocking questions!

PocketKit/Sources/PocketKit/FeatureFlagService.swift
```diff
@@ -0,0 +1,38 @@
+//
+//  File.swift
+//  
+//
+//  Created by Daniel Brooks on 11/22/22.
+//
```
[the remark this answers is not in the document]
bassrock: ugh. i wish xcode auto added this.

PocketKit/Sources/PocketKit/Refresh/UserRefreshCoordinator.swift
bassrock: i dumb.

PR #339, opened 2022-12-06
feat(search): Local search [IN-963][IN-889]

timc-mozilla:
## Summary
* Added search from core data

## References 
* [IN-963](https://getpocket.atlassian.net/jira/software/projects/IN/boards/80?selectedIssue=IN-963)
* [IN-889](https://getpocket.atlassian.net/jira/software/projects/IN/boards/80?selectedIssue=IN-889)

## Implementation Details
* added two calls for core data search

## Test Steps
* Go to search. run a search and check to see if the number of elements show as expected

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchViewModel.swift
+++ b/PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchViewModel.swift
@@ -74,6 +74,9 @@ class SearchViewModel: ObservableObject {
         submitSearch(with: term)
         showRecentSearches = false
 
+        _ = searchItems(with: searchTerm)
+        // TODO: pass results to update table
+
         guard isPremium, !term.isEmpty else { return }
         recentSearches = updateRecentSearches(with: term)
     }
@@ -106,6 +109,14 @@ class SearchViewModel: ObservableObject {
 
         searches.removeFirst()
         return searches
+    }
+
+    public func searchItems(with searchTerm: String) -> [SavedItem] {
+        let term = searchTerm.trimmingCharacters(in: .whitespaces).lowercased()
+        guard let list = source.searchSaves(search: term), !list.isEmpty else {
+            return []
+        }
+        return list
     }
 
     private func searchResultState() -> EmptyStateViewModel {
--- a/PocketKit/Sources/Sync/PocketSource.swift
+++ b/PocketKit/Sources/Sync/PocketSource.swift
@@ -646,3 +646,10 @@ extension PocketSource {
         }
     }
 }
+
+// MARK: - Search term
+extension PocketSource {
+    public func searchSaves(search: String) -> [SavedItem]? {
+        try? space.fetchSavedItems(bySearchTerm: search, userPremium: user.status == .premium)
+    }
+}
--- a/PocketKit/Sources/Sync/Requests.swift
+++ b/PocketKit/Sources/Sync/Requests.swift
@@ -58,6 +58,19 @@ public enum Requests {
         return request
     }
 
+    public static func fetchSavedItems(bySearchTerm searchTerm: String, userPremium isPremium: Bool) -> NSFetchRequest<SavedItem> {
+        let request = SavedItem.fetchRequest()
+        let urlPredicate = NSPredicate(format: "url CONTAINS %@", searchTerm)
+        let titlePredicate = NSPredicate(format: "item.title CONTAINS %@", searchTerm)
+        var allPredicate = NSCompoundPredicate(type: .or, subpredicates: [urlPredicate, titlePredicate])
+        if isPremium {
+            let tagsPredicate = NSPredicate(format: "%@ IN tags.name", searchTerm)
+            allPredicate = NSCompoundPredicate(type: .or, subpredicates: [urlPredicate, titlePredicate, tagsPredicate])
+        }
+        request.predicate = allPredicate
+        return request
+    }
+
     public static func fetchPersistentSyncTasks() -> NSFetchRequest<PersistentSyncTask> {
         let request = PersistentSyncTask.fetchRequest()
         request.sortDescriptors = [NSSortDescriptor(keyPath: \PersistentSyncTask.createdAt, ascending: true)]
--- a/PocketKit/Sources/Sync/Source.swift
+++ b/PocketKit/Sources/Sync/Source.swift
@@ -79,6 +79,8 @@ public protocol Source {
     func save(url: URL)
 
     func fetchItem(_ url: URL) -> Item?
+
+    func searchSaves(search: String) -> [SavedItem]?
 }
 
 public extension Source {
--- a/PocketKit/Tests/PocketKitTests/SearchViewModelTests.swift
+++ b/PocketKit/Tests/PocketKitTests/SearchViewModelTests.swift
@@ -1,8 +1,8 @@
 import XCTest
-import Sync
 import SharedPocketKit
 import PocketGraph
 
+@testable import Sync
 @testable import PocketKit
 
 class SearchViewModelTests: XCTestCase {
@@ -10,6 +10,7 @@ class SearchViewModelTests: XCTestCase {
     private var user: MockUser!
     private var userDefaults: UserDefaults!
     private var source: MockSource!
+    private var space: Space!
     private var searchService: MockSearchService!
 
     override func setUpWithError() throws {
@@ -17,8 +18,8 @@ class SearchViewModelTests: XCTestCase {
         user = MockUser()
         source = MockSource()
         userDefaults = UserDefaults(suiteName: "SearchViewModelTests")
+        space = .testSpace()
         searchService = MockSearchService()
-
         source.stubMakeSearchService { self.searchService }
     }
 
@@ -27,6 +28,7 @@ class SearchViewModelTests: XCTestCase {
         source = nil
         user = nil
         networkPathMonitor = nil
+        try space.clear()
     }
 
     func subject(
@@ -45,18 +47,24 @@ class SearchViewModelTests: XCTestCase {
 
     // MARK: - Update Scope
     func test_updateScope_forFreeUser_withOnlineSaves_showsSearchEmptyState() {
+        source.stubSearchItems { _ in return [] }
+
         let viewModel = subject()
         viewModel.updateScope(with: .saves)
         XCTAssertTrue(viewModel.emptyState is SearchEmptyState)
     }
 
     func test_updateScope_forFreeUser_withOnlineArchive_showsSearchEmptyState() {
+        source.stubSearchItems { _ in return [] }
+
         let viewModel = subject()
         viewModel.updateScope(with: .archive)
         XCTAssertTrue(viewModel.emptyState is SearchEmptyState)
     }
 
     func test_updateScope_forFreeUser_withAll_showsGetPremiumEmptyState() {
+        source.stubSearchItems { _ in return [] }
+
         let viewModel = subject()
         viewModel.updateScope(with: .all)
         XCTAssertTrue(viewModel.emptyState is GetPremiumEmptyState)
@@ -64,6 +72,8 @@ class SearchViewModelTests: XCTestCase {
 
     func test_updateScope_forPremiumUser_withSaves_showsRecentSearchEmptyState() {
         user.status = .premium
+
+        source.stubSearchItems { _ in return [] }
 
         let viewModel = subject()
         viewModel.updateScope(with: .saves)
@@ -73,6 +83,8 @@ class SearchViewModelTests: XCTestCase {
     func test_updateScope_forPremiumUser_withArchive_showsRecentSearchEmptyState() {
         user.status = .premium
 
+        source.stubSearchItems { _ in return [] }
+
         let viewModel = subject()
         viewModel.updateScope(with: .archive)
         XCTAssertTrue(viewModel.emptyState is RecentSearchEmptyState)
@@ -81,6 +93,8 @@ class SearchViewModelTests: XCTestCase {
     func test_updateScope_forPremiumUser_withAll_showsRecentSearchEmptyState() {
         user.status = .premium
 
+        source.stubSearchItems { _ in return [] }
+
         let viewModel = subject()
         viewModel.updateScope(with: .archive)
         XCTAssertTrue(viewModel.emptyState is RecentSearchEmptyState)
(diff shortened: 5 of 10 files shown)


PocketKit/Sources/Sync/Source.swift
```diff
@@ -77,6 +77,10 @@ public protocol Source {
     func save(url: URL)
 
     func fetchItem(_ url: URL) -> Item?
+
+    func searchPremiumSaves(search: String) -> [SavedItem]?
+
+    func searchSaves(search: String) -> [SavedItem]?
```
cyndichin, line 83: what do you think about consolidating this to one search function and use a boolean or `user.status` to determine what should be fetched?

timc-mozilla: ok

PocketKit/Sources/Sync/PocketSource.swift
```diff
@@ -642,3 +642,10 @@ extension PocketSource {
         }
     }
 }
+
+// MARK: - Search term
+extension PocketSource {
+    public func searchSaves(search: String, isPremium: Bool) -> [SavedItem]? {
+        try? space.fetchSavedItems(bySearchTerm: search, userPremium: isPremium)
```
cyndichin, line 649: what do you think about using `user.status == .premium` as opposed to passing `isPremium` since sync should have a user object?

This way we can simplify the call in the `SearchViewModel` and trust that source will return the right data. Similar to how we rely on the server to provide us the right items via Apollo.

```
let list = self.source.searchSaves(search: term)
```

PocketKit/Sources/Sync/Requests.swift
```diff
@@ -58,6 +58,19 @@ public enum Requests {
         return request
     }
 
+    public static func fetchSavedItems(bySearchTerm searchTerm: String, userPremium isPremium: Bool) -> NSFetchRequest<SavedItem> {
+        let request = SavedItem.fetchRequest()
+        let urlPredicate = NSPredicate(format: "url CONTAINS %@", searchTerm)
+        let titlePredicate = NSPredicate(format: "item.title CONTAINS %@", searchTerm)
+        let tagsPredicate = NSPredicate(format: "%@ IN tags.name", searchTerm)
```
cyndichin, line 65: nitpick - maybe add `tagsPredicate` in the conditional check, so we don't create this object unless necessary

PocketKit/Tests/PocketKitTests/SearchViewModelTests.swift
```diff
@@ -1,52 +1,65 @@
 import XCTest
-import Sync
 import SharedPocketKit
 
+@testable import Sync
```
cyndichin, line 4: curious, why @testable import vs regular import

PocketKit/Tests/PocketKitTests/SearchViewModelTests.swift
```diff
@@ -198,4 +243,118 @@ class SearchViewModelTests: XCTestCase {
...
+                createdAt: Date(timeIntervalSince1970: TimeInterval($0)),
+                tags: ["test"]
+            )
+        }
+        try space.save()
+
+        source.stubSearchItems { _ in return [] }
+
+        let viewModel = subject()
+        var results = viewModel.searchItems(with: "saved")
```
cyndichin, line 313: should this be "test" if we're trying to search on tags? also maybe rename test to shouldNotShowResults

cyndichin (commented):
overall, looks good! core data is not easy so very excited to have local search! have a few comments, some curious + some nitpick

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchViewModel.swift
```diff
@@ -86,6 +91,23 @@ class SearchViewModel: ObservableObject {
         return searches
     }
 
+    public func searchItems(with searchTerm: String) -> [SavedItem] {
+        let term = searchTerm.trimmingCharacters(in: .whitespaces).lowercased()
+        var listOfItems = [SavedItem]()
+        guard isPremium else {
+            let list = self.source.searchSaves(search: term, isPremium: false)
+            if !(list?.isEmpty ?? false) {
```
dskuza, line 99: This could maybe be cleaned up with a `guard let list = ...`

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchViewModel.swift
```diff
@@ -86,6 +91,23 @@ class SearchViewModel: ObservableObject {
...
+                listOfItems = list ?? []
+            }
+            return listOfItems
+        }
+        let list = self.source.searchSaves(search: term, isPremium: isPremium)
+        if !(list?.isEmpty ?? false) {
+            listOfItems = list ?? []
+        }
```
dskuza, line 107: This could maybe be cleaned up with a `guard let list = ...`

PocketKit/Tests/PocketKitTests/SearchViewModelTests.swift
```diff
@@ -198,4 +243,118 @@ class SearchViewModelTests: XCTestCase {
...
+
+        source.stubSearchItems { _ in return [] }
+
+        results = viewModel.searchItems(with: "none")
+        XCTAssertTrue(results.isEmpty)
+
+        try space.clear()
+    }
+
+    func test_savesSearches_withFreeUser_showSearchResults_searchTag() throws {
```
dskuza, line 298:
```suggestion
    func test_savesSearches_withFreeUser_showSearchResults_doesNotSearchTag() throws {
```

dskuza (changes requested):
Few minor changes in line with what Cyndi had suggested.

PocketKit/Tests/PocketKitTests/SearchViewModelTests.swift
timc-mozilla: `@testable` provides unit level access, meaning that I would have access to all internal modules and methods that are also marked testable without having to individually import them.

PocketKit/Sources/Sync/Requests.swift
timc-mozilla: ok

PocketKit/Tests/PocketKitTests/SearchViewModelTests.swift
cyndichin: This makes sense. Usually, we are testing internal functions for a specific class so we import the framework that it belongs in. However, it seems that we are using the internal methods of space (which is a bit odd to me) in order to help set up and tear down our tests related to core data items.

timc-mozilla: solved in a discussion hon slack. https://pocket.slack.com/archives/G01A1760T71/p1671037671453809

PocketKit/Tests/PocketKitTests/SearchViewModelTests.swift
timc-mozilla: I think that testing Title and URL are the same.  They are both string comparisons that take place.  I don't see a benefit in testing both if one of them works as intended.
Tags are a bit different in that Premium searches search Tag whereas non-premium doesn't search tags.
So I'm testing the behavior of Core Data returning based on user status.

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchViewModel.swift
```diff
@@ -86,6 +91,25 @@ class SearchViewModel: ObservableObject {
...
+    public func searchItems(with searchTerm: String) -> [SavedItem] {
+        let term = searchTerm.trimmingCharacters(in: .whitespaces).lowercased()
+        var listOfItems = [SavedItem]()
+        guard isPremium else {
+            guard let list = source.searchSaves(search: term), list.isEmpty else {
+                listOfItems = []
+                return listOfItems
+            }
+            listOfItems = list
+            return listOfItems
```
dskuza, line 103: This codeblock seems to be a duplicate of the immediately following `guard let list = ...`. Do we still need to guard against `isPremium`, since that's now in `Source`? I feel like there's an opportunity to remove some duplicate code.

PocketKit/Tests/PocketKitTests/SearchViewModelTests.swift
cyndichin: thanks for the insight! that sounds good to me too, only suggestion is to modify `saved` to `test` and add the assertion

PocketKit/Sources/Sync/PocketSource.swift
cyndichin: didn't see any modifications related to this, did I miss the conclusion we decided to take?

PocketKit/Tests/PocketKitTests/SearchViewModelTests.swift
timc-mozilla: > Title and URL aren't guaranteed to have the same strings in them.
i'm not saying that Title and URL would have the same strings, but if one string comparison works, the other should also work.
But understood, I'll add another test for URLs

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchViewModel.swift
timc-mozilla: yup yup. cleaned it up.

PocketKit/Tests/PocketKitTests/SearchViewModelTests.swift
```diff
@@ -134,8 +156,12 @@ class SearchViewModelTests: XCTestCase {
 
     func test_updateSearchResults_forPremiumUser_withNoItems_showsNoResultsEmptyState() {
         searchService.stubSearch { _, _ in }
+        source.stubSearchItems { _ in return [] }
 
         user.status = .premium
+
+        source.stubSearchItems { _ in return [] }
+
```
cyndichin, line 164: remove?

PocketKit/Tests/PocketKitTests/SearchViewModelTests.swift
cyndichin: thanks for adding the url test, but seems like we didn't make the update from `saved` to `test` unless I'm missing something. also I like adding the assertion that @dskuza mentioned above 

> I think we can have an additional assertion (within this test) for testing against the tag "test" to verify that we are returning 0 items.

PocketKit/Tests/SyncTests/PocketSourceTests.swift
```diff
@@ -625,3 +625,62 @@ extension PocketSourceTests {
...
+        let results = source.searchSaves(search: "saved")
+        XCTAssertTrue(results?.count == 2)
+
+        let tagResults = source.searchSaves(search: "tag")
+        XCTAssertTrue((tagResults?.isEmpty) != nil)
+
+        try space.clear()
+    }
+
+    func test_searchItems_searchPremiumStandard() throws {
```
cyndichin, line 659: how are we determining the premium standard here?

PocketKit/Tests/PocketKitTests/SearchViewModelTests.swift
timc-mozilla: yup, sorry. i think it got lost when i was fixing a bunch of merge conflicts.

PocketKit/Sources/Analytics/Context/UIContext.swift
```diff
@@ -102,12 +102,6 @@ extension UIContext {
         case tagsSaveChanges = "tagsSaveChanges"
         case externalApp = "external_app"
         case saveExtension = "save_extension"
-        case sortFilterSheet = "sort_filter"
-        case sortByNewest = "sortByNewest"
-        case sortByOldest = "sortByOldest"
-        case sortByLongest = "sortByLongest"
-        case sortByShortest = "sortByShortest"
-        case navigationDrawer = "navigationDrawer"
```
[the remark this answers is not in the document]
timc-mozilla: okee dokee

PocketKit/Sources/Analytics/Context/UIContext.swift
```diff
@@ -162,7 +156,6 @@ extension UIContext {
         public let saves = UIContext(type: .list, hierarchy: 0, identifier: .saves)
         public let archive = UIContext(type: .list, hierarchy: 0, identifier: .archive)
         public let favorites = UIContext(type: .list, hierarchy: 0, identifier: .favorites)
-        public let sortFilterSheet = UIContext(type: .screen, identifier: .sortFilterSheet)
```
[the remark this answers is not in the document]
timc-mozilla: same as above.

PocketKit/Tests/PocketKitTests/SearchViewModelTests.swift
timc-mozilla: I still dont quite understand the reason for this.  Aren't I stubbing the searchItems to return an empty array?  Unless I'm just not understanding whats being asked.

PocketKit/Tests/SyncTests/PocketSourceTests.swift
timc-mozilla: no reason. no one said anything so I just left it.  I'll remove it.

PocketKit/Tests/PocketKitTests/SearchViewModelTests.swift
cyndichin: Oh I see the updated changes now! Looks good The reason is that we're checking specifically for tags, so would want to make sure we are checking `viewModel.searchItems(with: "test")`

PocketKit/Tests/SyncTests/PocketSourceTests.swift
cyndichin: yes pls remove if we don't need this test anymore. just noticed it now

PocketKit/Tests/SyncTests/PocketSourceTests.swift
```diff
@@ -625,3 +625,62 @@ extension PocketSourceTests {
...
+        )
+        _ = createItemsWithTags(2)
+
+        try space.save()
+
+        let results = source.searchSaves(search: "saved")
+        XCTAssertTrue(results?.count == 2)
+
+        let tagResults = source.searchSaves(search: "tag")
+        XCTAssertTrue((tagResults?.isEmpty) != nil)
```
cyndichin, line 654: if we are testing a free user here, shouldn't `tagResults` be empty?

timc-mozilla: yup.

cyndichin (approved):
Thanks for addressing all the comments! Great work :)

PR #341, opened 2022-12-08
feat(search): online search

cyndichin:
## Summary
Add online search for free archive and premium. Currently allows online search for free user until we integrate with local search.

## References 
IN-890, IN-891, IN-892, IN-967

## Implementation Details
Added new GraphQL query `SearchSavedItems` and created a new class called `SearchService` to execute the network request via apollo.  Retrieve results from API and sent the data back as an array of titles to the `SearchViewModel` via publisher. Create a new `SearchItem` class and initialize an item object by passing in properties from `SearchSavedItemParts` and `SavedItem`. We can create an array of `[SearchItem]` which would be used to populate the `SearchView`.

## Test Steps
- [ ] WHEN a free user submits a search
AND “Archive” is the selected scope
AND the device is online
THEN submit the search term by API

- [ ] WHEN I am a premium user
AND the search experience is active
AND the device is online
AND the selected scope is “Saves”
AND I submit a search term
THEN we submit the search term to the API

- [ ] WHEN I am a premium user
AND my device is online
AND the search experience is active
AND the selected scope is “Archives”
AND I submit a search term
THEN we submit the search term by API

- [ ] WHEN search is active
AND the user has a premium subscription
AND the device is online
AND the “All Items” scope is selected
AND the user submits a query
THEN display matched search results for both saved and archived articles

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

## Screenshots
|Online Search (Saves)| Online Search (Archive)
|------|------|
|Online Search Saves|Online Search Archive|


PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchItem.swift
```diff
@@ -0,0 +1,28 @@
...
+            item.bestURL?.absoluteString
+        ]
+            .compactMap { $0 }
+            .first { !$0.isEmpty } ?? ""
+    }
+
+    var detail: String? {
+        guard let tagNames = item.tagNames?.joined(separator: " "), !tagNames.isEmpty else { return nil }
+        return "Tags: \(tagNames)"
+    }
```
cyndichin, line 27: this is temporary for demo purposes until we flesh this out

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchViewModel.swift
```diff
@@ -64,14 +69,26 @@ class SearchViewModel: ObservableObject {
...
+        self.searchResults = []
+    }
+
+    private func submitSearch(with term: String) {
+        clear()
+        source.search(for: term, scope: selectedScope)
+        source.events.receive(on: DispatchQueue.main).sink { [weak self] event in
+            guard case let .searchListFetched(items) = event, let self else { return }
+            self.searchResults = items.compactMap { SearchItem(item: $0) }
+        }.store(in: &subscriptions)
```
dskuza, line 89: Maybe I'm overlooking something - is this subscription ever getting cancelled? E.g when the search completed, else we have a lingering cancellable every time a search is performed.

PocketKit/Sources/Sync/Operations/SearchList.swift
```diff
@@ -0,0 +1,72 @@
+import Apollo
+import PocketGraph
+import SharedPocketKit
+
+class SearchList: SyncOperation {
```
dskuza, line 5: We also have `AsyncOperation` and `RetriableOperation`. Out of curiosity, would it make sense to leverage one of those? But, this implementation seems fine as well.

PocketKit/Sources/Sync/PocketSource.swift
```diff
@@ -435,6 +435,19 @@ extension PocketSource {
...
+// MARK: - Search
+extension PocketSource {
+    public func search(for term: String, scope: SearchScope) {
+        let operation = operations.searchList(
+            apollo: apollo,
+            scope: scope,
+            term: term,
+            events: _events
+        )
+        enqueue(operation: operation, task: .searchList(term: term, scope: scope))
```
dskuza, line 447: I'm wondering if it makes sense to enqueue these operations as we do other actions, such as favoriting. Search almost feels like it could be its own service, where its search can run in parallel to this queue. My thoughts are that if a user's offline and stacks a _bunch_ of item actions (archive, favorites, etc), opens search, it's enqueued, then goes online, then the search technically runs, no? Based on this implementation.

dskuza (commented):
@cyndichin I don't want to dive too deep into test-land, since I have a major question regarding our use of the Sync queue for search. Mind taking a peek at my question and continuing the conversation?

PocketKit/Sources/PocketKit/MyList/SavedItemsList/SavedItem+ItemsListItem.swift
```diff
@@ -2,6 +2,10 @@ import Sync
 import Foundation
 
 extension SavedItem: ItemsListItem {
+    var id: String? {
+        item?.remoteID
+    }
```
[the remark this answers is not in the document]
cyndichin: exactly! I needed an identifier to iterate over the array of items, so I decided to use the remoteID here, which defines SearchItem's `id` property

PocketKit/Sources/Sync/Operations/SearchList.swift
cyndichin: Good point! I think Search fits better as `AsyncOperation` vs `SyncOperation` since now I think about it we would probably want to just rely on server data at that point and not need to wait for our other operations to complete even if the are pushing changes up to the server. I'll make this update. 

I didn't see any specification on the retry front so don't think it's in the scope of search. Do you see a ticket / requirement that I missed, otherwise I'll clarify during standup. We did have a Try Again button previously if the user was offline, but it was removed and the user would just search again if anything.

PocketKit/Sources/Sync/PocketSource.swift
cyndichin: Yeah, I saw a delay in search too while waiting for the fetch list to load, so considering the conversation regarding `Async`, I'll remove this. In terms of the user being offline and enqueuing, I'll make those considerations when working on this ticket `https://getpocket.atlassian.net/browse/IN-980`.

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchViewModel.swift
cyndichin: Nope I overlooked this! You're correct, I'll clear out the subscription in `clear()`

PocketKit/Sources/Sync/Operations/SearchList.swift
```diff
@@ -24,19 +24,32 @@ class SearchList: SyncOperation {
...
-    func execute() async -> SyncOperationResult {
+    override func start() {
+        guard !isCancelled else {
+            finishOperation()
+            return
+        }
+
+        Task {
+            isFinished = false
+            isExecuting = true
```
cyndichin, line 35: kind of mimicked `FetchArchivePagesOperation`, but after reading this article `https://www.avanderlee.com/swift/asynchronous-operations/`, I think we should update `isFinished` and `isExecuting` here

PocketKit/Tests/SyncTests/Support/SearchListTests.swift
```diff
@@ -0,0 +1,104 @@
...
+            term: term ?? self.term,
+            events: events ?? self.events
+        )
+    }
+
+    func test_search_forSaves_fetchesSearchSavedItemsQueryWithTerm() {
+        apollo.setupSearchListResponse()
+        let service = subject()
+        let searchExpectation = expectation(description: "searchExpectation")
+
```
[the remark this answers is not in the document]
cyndichin: oopsie daisies

dskuza:
@cyndichin I still have my concerns regarding keeping search within `PocketSource`, and not a separate service. Would you like to discuss live?

PocketKit/Sources/Sync/PocketSource.swift
cyndichin: Discussed over zoom and decided to create a separate service without using OperationQueues to avoid complexity

PocketKit/Sources/Sync/Operations/SearchList.swift
cyndichin: Discussed over zoom and decided to create a separate service without using OperationQueues to avoid complexity

cyndichin:
> @cyndichin I still have my concerns regarding keeping search within `PocketSource`, and not a separate service. Would you like to discuss live?

We discuss this live, updated the changes in this [commit](https://github.com/Pocket/pocket-ios/pull/341/commits/648f36842bb837de33f0cfe137967c6b0c13486c)

dskuza (approved):
Awesome, I'm really happy with the move to a search service. Just a few comments, but nothing blocking this pull request from being merged.

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchViewModel.swift
```diff
@@ -64,14 +69,27 @@ class SearchViewModel: ObservableObject {
...
 
+    func clear() {
+        searchResults = []
+        subscriptions = []
+    }
+
+    private func submitSearch(with term: String) {
+        clear()
+        source.search(for: term, scope: selectedScope)
+        source.events.receive(on: DispatchQueue.main).sink { [weak self] event in
```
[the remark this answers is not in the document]
cyndichin: oh this code has been outdated, but yes you're right! the updated changes should have it subscribed first. This is the new code using `searchService` instead of `source`

```
searchService.results.receive(on: DispatchQueue.main).sink { [weak self] items in
             guard let self else { return }
             self.searchResults = items.compactMap { SearchItem(item: $0) }
         }.store(in: &subscriptions)
         Task {
             await searchService.search(for: term, scope: selectedScope)
         }
```

Tests iOS/Support/Elements/SearchViewElement.swift
```diff
@@ -22,4 +22,16 @@ struct SearchViewElement: PocketUIElement {
...
     }
+
+    var searchResultsView: XCUIElement {
+        let query: XCUIElementQuery
+
+        if #available(iOS 16, *) {
+            query = element.collectionViews
+        } else {
+            query = element.tables
+        }
```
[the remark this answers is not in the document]
cyndichin: yes! its not based on table views anymore

PocketKit/Sources/Sync/Operations/SearchService.swift
```diff
@@ -1,48 +1,46 @@
 import Apollo
+import Foundation
 import PocketGraph
 import SharedPocketKit
+import Combine
 
-class SearchList: SyncOperation {
+public protocol SearchService: AnyObject {
+    var results: Published<[SearchSavedItemParts]>.Publisher { get }
+    func search(for term: String, scope: SearchScope) async
```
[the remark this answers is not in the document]
cyndichin: I was considering this as well, but the pagination makes this a bit difficult as we don't want to wait until all items are loaded before we return. Instead we use the published property to send out 30 items at a time.

PR #345, opened 2022-12-15
ci(tests): flakey test fixes

audie1:
## Summary
Fix attempts for flakey UI tests

## References 
https://getpocket.atlassian.net/browse/IN-950

## Implementation Details
Skeleton view test failure is the first attempt

## Test Steps
If Bitrise stops failing the tests, we will know the fixes are working :)

### DIFF
--- a/Tests iOS/MyList/SavesTests.swift
+++ b/Tests iOS/MyList/SavesTests.swift
@@ -311,11 +311,17 @@ class SavesTests: XCTestCase {
         XCTAssertEqual(listView.itemCount, 0)
         XCTAssertEqual(listView.skeletonCellCount, 4)
         promises[0].completeWith(.success(Response.saves("saves-loading-page-1")))
-
+        //            These conditionals try to fix the flakey test failures by reloading saves
+        if listView.itemCount != 2 {
+            promises[0].completeWith(.success(Response.saves("saves-loading-page-1")))
+        }
         XCTAssertEqual(listView.itemCount, 2)
         XCTAssertEqual(listView.skeletonCellCount, 1)
-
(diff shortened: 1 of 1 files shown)


Tests iOS/MyList/SavesTests.swift
```diff
@@ -312,10 +312,17 @@ class SavesTests: XCTestCase {
         XCTAssertEqual(listView.skeletonCellCount, 4)
         promises[0].completeWith(.success(Response.saves("saves-loading-page-1")))
 
+        if listView.itemCount != 2 {
+//            This conditional tries fix the flakey test failures by reloading saves
```
cyndichin, line 316: seems like we're doing a check to see if the itemCount equals, otherwise do another reload. What if it fails again? Do we know why it fails the first time?

audie1: I can't say for sure, but I was able to replicate a failure every time in Xcode by setting a breakpoint after the item load, and then removing the breakpoint and re-running - it would fail after the removal. This fix solved the problem, but I have no way of knowing for sure if Bitrise fails for the same reason so if this doesn't work I'll have to try something else.

cyndichin (approved):

Tests iOS/MyList/SavesTests.swift
cyndichin: thanks for looking into this! approving as this will help avoid the test from failing

audie1:
This is a first attempt at a fix for the failing UI test `test_list_showsSkeletonCellsDuringInitialFetch` which has been failing a lot.

PR #347, opened 2022-12-19
ci(tests): laggy reader test fix

audie1:
## Summary
test_tappingItem_displaysNativeReaderView in SavesTests.swift has this strange behavior where it's spending a lot of extra time scrolling the Reader UI. The test isn't currently breaking but it's using a lot of extra time so I'll add this as something to look into and fix as part of the Flakey tests work.

## References 
Jira ticket https://getpocket.atlassian.net/browse/IN-950

## Implementation Details
I'm going to change how the elements are found and scrolled to in order to fix this.

## Test Steps
The test should take much less time to complete after the changes.

## Screenshots
Gif in Slack:
https://pocket.slack.com/archives/C01UH57EJKD/p1670535231375889

### DIFF
--- a/Tests iOS/MyList/SavesTests.swift
+++ b/Tests iOS/MyList/SavesTests.swift
@@ -226,12 +226,13 @@ class SavesTests: XCTestCase {
         ]
 
         for expectedString in expectedContent {
-            let cell = app
-                .readerView
-                .cell(containing: expectedString)
-                .wait()
-
-            app.readerView.scrollCellToTop(cell)
+            guard app.readerView.cell(containing: expectedString).isHittable else {
+                app.readerView.element.swipeUp()
+                let cell = app.readerView.cell(containing: expectedString)
+                app.readerView.scrollCellToTop(cell)
+                XCTAssertTrue(cell.exists)
+                return
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


Tests iOS/MyList/SavesTests.swift
```diff
@@ -226,12 +226,15 @@ class SavesTests: XCTestCase {
...
-            app.readerView.scrollCellToTop(cell)
+            if app.readerView.cell(containing: expectedString).isHittable {
+                print("The string was found:")
+                print(expectedString)
+            } else {
+                app.readerView.element.swipeUp()
+                let cell = app.readerView.cell(containing: expectedString)
+                app.readerView.scrollCellToTop(cell)
+                XCTAssertTrue(cell.exists)
+            }
```
cyndichin, line 237: does this check if its hittable first before scrolling? Do we want to an some assertions here instead of prints?

audie1: After discussion, we decided to go with a guard conditional, I'll implement that now.

cyndichin (approved):

audie1:
Merging now...

Previous run time: 61.432 seconds
New run time: 17.573 seconds

PR #348, opened 2022-12-20
feat(search): search return type

timc-mozilla:
## Summary
* changed return key type

## References 
* [IN-969](https://getpocket.atlassian.net/browse/IN-969)

## Implementation Details
* changed return type

## Test Steps
* Go to search
* if nothing is entered, search should be not allowed
* if text is entered, search should be allowed

## PR Checklist:
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/PocketKit/MyList/SavesContainerViewController.swift
+++ b/PocketKit/Sources/PocketKit/MyList/SavesContainerViewController.swift
@@ -129,6 +129,7 @@ class SavesContainerViewController: UIViewController, UISearchBarDelegate {
 
     func searchBar(_ searchBar: UISearchBar, selectedScopeButtonIndexDidChange selectedScope: Int) {
         guard let titles = searchBar.scopeButtonTitles else { return }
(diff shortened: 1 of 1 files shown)


PocketKit/Sources/PocketKit/MyList/SavesContainerViewController.swift
```diff
@@ -20,7 +20,7 @@ protocol SelectableViewController: UIViewController {
     func didBecomeSelected(by parent: SavesContainerViewController)
 }
 
-class SavesContainerViewController: UIViewController, UISearchBarDelegate {
+class SavesContainerViewController: UIViewController, UISearchBarDelegate, UITextFieldDelegate {
```
cyndichin, line 23: why do we need to add this delegate?

timc-mozilla: needed it for an older solution i had for it. removed now.

cyndichin (approved):
🔥

PR #351, opened 2022-12-22
feat(search): search with recent search

timc-mozilla:
## Summary
* search with recent search

## References 
* [IN-960](https://getpocket.atlassian.net/browse/IN-960)

## Implementation Details
* modified keyboard and navigation

## Test Steps
* search for two different items
* click the either one again
* search should update, keyboard should hide, and the list should update with most recent search order

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

## Screenshots
N/A

### DIFF
--- a/PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchView.swift
+++ b/PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchView.swift
@@ -10,7 +10,7 @@ struct SearchView: View {
         if let results = viewModel.searchResults, !results.isEmpty {
             ResultsView(results: results)
         } else if viewModel.showRecentSearches == true, !viewModel.recentSearches.isEmpty {
-            RecentSearchView(recentSearches: viewModel.recentSearches)
+            RecentSearchView(viewModel: viewModel, recentSearches: viewModel.recentSearches)
         } else if let emptyState = viewModel.emptyState {
             SearchEmptyView(viewModel: emptyState)
         }
@@ -46,6 +46,8 @@ struct SearchEmptyView: View {
 
 // MARK: - Recent Searches Component
 struct RecentSearchView: View {
+    @ObservedObject
+    var viewModel: SearchViewModel
     var recentSearches: [String]
 
     var body: some View {
@@ -59,7 +61,7 @@ struct RecentSearchView: View {
                     }
                     .contentShape(Rectangle())
                     .onTapGesture {
-                        // TODO: Handle recent search tap
+                        viewModel.searchText = recentSearch
                     }
                 }
             }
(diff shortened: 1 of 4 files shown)


Tests iOS/MyList/SearchTests.swift
```diff
@@ -178,4 +178,28 @@ class SearchTests: XCTestCase {
...
+
+        searchField.tap()
+        searchField.typeText("test\n")
+        searchView = app.saves.searchView.searchResultsView.wait()
+
+        app.navigationBar.buttons["Cancel"].tap()
+
+        let app = XCUIApplication()
+        searchField.tap()
+        app.collectionViews["recent-searches"].staticTexts["item"].tap()
```
timc-mozilla, line 202: unfamiliar with XCTest and how to get the test to tap the list item `item`.   If there is a better solution, please let me know.

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchView.swift
```diff
@@ -59,7 +61,7 @@ struct RecentSearchView: View {
                     }
                     .contentShape(Rectangle())
                     .onTapGesture {
-                        // TODO: Handle recent search tap
+                        viewModel.updateSearchFromRecentSearch(with: recentSearch)
```
cyndichin, line 64: what do you think about doing `viewModel.searchText = recentSearch` here instead of creating a function?

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchViewModel.swift
```diff
@@ -54,6 +54,8 @@ class SearchViewModel: ObservableObject {
         }
     }
 
+    @Published var searchText = ""
```
cyndichin, line 57: we can add a `didSet` here so that we know that whenever this is updated, then we want to call `updateSearchResults `

could we rename this to something similar to `recentSearchTerm` so we have context that this is for recent search specifically?

Tests iOS/MyList/SearchTests.swift
cyndichin: this seems fine to me, can we use `app.saves.searchView.recentSearchesView` instead of `collectionViews["recent-searches"]`. If you want to get fancy, you can also make a element (similar to when we use `app.saves.itemView(matching: "Item 1")`) so that you can use 
`app.saves.searchView.recentSearchesView(matching: "item")`

dskuza (approved):
Cyndi raises some good ideas. I'll approve with the understanding that those might be addressed.

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchView.swift
timc-mozilla: we'd still need to call the search function, so im not sure if we're solving anything by adding the change you suggested.

timc-mozilla: nvm, i see the changes you requested for `SearchViewModel`

PR #352, opened 2023-01-04
feat(search): select search scopes

cyndichin:
## Summary
Switching scopes during an active search performs a search and results are cached if already performed a search for the term. _Note: Loading state will replace current empty state in future ticket._

## References 
IN-973

## Implementation Details
Add new parameter for `updateScope` to take in search term if use switch scopes. This triggers a new search. Created new property `cachedSearchResults` to store cached results in a session and clear when user exits search experience.

## Test Steps
- [ ] WHEN I have an active search
AND I switch scopes
AND we have not yet performed a search for this search term and scope during this search session
THEN we perform the search (if possible)

- [ ] WHEN I have an active search
AND I switch scopes
AND we have performed a search for this search term and scope during this search session
THEN we display the previously displayed search results
 
Note: show offline view if no connection

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

## Screenshots
|iPad|
|------|
|iPad|

[IN-973]: https://getpocket.atlassian.net/browse/IN-973?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchViewModel.swift
+++ b/PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchViewModel.swift
@@ -20,11 +20,14 @@ class SearchViewModel: ObservableObject {
         return user.status == .premium
     }
 
-    private var selectedScope: SearchScope = .saves {
-        didSet {
-            self.emptyState = isPremium ? premiumEmptyState(for: selectedScope) : freeEmptyState(for: selectedScope)
-        }
-    }
+    private var selectedScope: SearchScope = .saves
+    private var currentSearchTerm: String = ""
+
+    private var cachedSearchResults: [SearchScope: [String: [SearchItem]]] = [
+        .saves: [:],
+        .archive: [:],
+        .all: [:]
+    ]
 
     @Published
     var emptyState: EmptyStateViewModel?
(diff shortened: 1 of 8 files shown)


PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchViewModel.swift
```diff
@@ -103,6 +134,7 @@ class SearchViewModel: ObservableObject {
     }
 
     private func updateRecentSearches(with searchTerm: String) -> [String] {
+        guard isPremium else { return [] }
```
cyndichin, line 137: had to move this check here for my tests to pass for free users, not sure why but it works. If anyone has insight on why moving this allows `test_updateScope_forFreeUser_withArchiveAndTerm_showsResults` to pass, I'm open ears!

cyndichin: Got on a call with @timc-mozilla and @OrlaM during our weekly search meetings. It seems that this may be due to the asynchronous search call via our search service. However, we are using a completion block to mock that call, so something funky seems to be happening. We discussed that our future strategy as a team should either be to consistently use the async / await pattern or completion block pattern. Will open this discussion in slack.

timc-mozilla (approved):
Live code reviewed.  Cyndi, as long as your comfortable with the change you made for the test, its good with me.

cyndichin:
> Live code reviewed. Cyndi, as long as your comfortable with the change you made for the test, its good with me.
It may change as we work on future tickets and I do want to revisit our tests towards the end, but good for now! Thanks!

PR #354, opened 2023-01-06
feat(search): support pagination, copy updates, cancel search

cyndichin:
## Summary
This PR consist of work related to 3 tickets:
* Support pagination for online search results
* Copy updates for Search Flow
* User can cancel searching

## References 
IN-993, IN-1089, IN-968

## Implementation Details
Small changes in terms of updating string values for the various empty states, modify pagesize constant to use `30` or `50` depending on device size and clear search field when user cancels search.

## Test Steps
- [ ] We should request the same number of search results per page, as we do list items. 
[ ] For small devices, like phone, this should be 30 items per page. 
[ ] For large devices, like iPads, this should be 50 items per page. 
- [ ] Verify copy updates for search flow
- [ ] WHEN search is active
AND the user taps the cancel button
THEN the search experience is closed
AND the user is returned to where they were before activating search (e.g., saved or archived)
AND any active searches are cancelled

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/PocketKit/MyList/SavesContainerViewController.swift
+++ b/PocketKit/Sources/PocketKit/MyList/SavesContainerViewController.swift
@@ -128,7 +128,7 @@ class SavesContainerViewController: UIViewController, UISearchBarDelegate {
         }
         navigationItem.searchController?.showsSearchResultsController = true
 
-        searchViewModel.$searchText.sink { searchText in
+        searchViewModel.$searchText.dropFirst().sink { searchText in
             self.updateSearchBar(searchText: searchText)
         }.store(in: &subscriptions)
     }
@@ -158,6 +158,7 @@ class SavesContainerViewController: UIViewController, UISearchBarDelegate {
     func searchBarCancelButtonClicked(_ searchBar: UISearchBar) {
         searchViewModel.clear()
         searchViewModel.clearCache()
+        searchBar.text = nil
     }
 
     func updateSearchScope() {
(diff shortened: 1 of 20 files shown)


PocketKit/Sources/PocketKit/MyList/SavesContainerViewController.swift
```diff
@@ -128,7 +128,7 @@ class SavesContainerViewController: UIViewController, UISearchBarDelegate {
         }
         navigationItem.searchController?.showsSearchResultsController = true
 
-        searchViewModel.$searchText.sink { searchText in
+        searchViewModel.$searchText.dropFirst().sink { searchText in
```
dskuza, line 131: Clarification, please! Is this because `$searchText` is getting published the first time it's subscribed to, which is kind of unnecessary (since it's an empty string)?

dskuza (commented):
@cyndichin specifically for this mention in the PR description:

> Support pagination for online search results

Is this simply an update to a constant value? There's nothing code-wise that really adds support for pagination, if I'm reading correctly.

PocketKit/Sources/PocketKit/MyList/SavesContainerViewController.swift
cyndichin: when 

> @cyndichin specifically for this mention in the PR description:
> 
> > Support pagination for online search results
> 
> Is this simply an update to a constant value? There's nothing code-wise that really adds support for pagination, if I'm reading correctly.

yes, the constant modifies the pagination input, which was already implemented in previous online search work


```
var shouldFetchNextPage = true
var items: [SearchSavedItemParts] = []
var pagination = PaginationInput(first: GraphQLNullable<Int>(integerLiteral: Constants.pageSize))

 while shouldFetchNextPage {
            let filter = getSearchFilter(with: scope)
            let query = SearchSavedItemsQuery(term: term, pagination: .init(pagination), filter: .some(filter))
            let result = try await apollo.fetch(query: query)
            result.data?.user?.searchSavedItems?.edges.forEach { edge in
                items.append(edge.node.savedItem.fragments.searchSavedItemParts)
            }
            if let pageInfo = result.data?.user?.searchSavedItems?.pageInfo {
                pagination.after = pageInfo.endCursor ?? .none
                shouldFetchNextPage = pageInfo.hasNextPage
            } else {
                shouldFetchNextPage = false
            }
            _results = items
        }
```

PR #355, opened 2023-01-09
ci(tests): fix for saving tags test

audie1:
## Summary
I'd like to rewrite some of the test_addTagsToItemFromSaves_savesNewTags() test in AddTagsItemTests

Currently, it doesn't quite function correctly, because the tag name is typically not entered correctly by the simulator keyboard, so only the stored value is checked, which doesn't account for the new tag name being added successfully
I'm going to figure out a solution and check the new tag name on the Saves screen, after the add tags modal is dismissed to sharpen up this test a bit

## Implementation Details
Basically I'm deleting the extra character that the simulator keyboard adds after the field is tapped, then choosing a random Int to use as a temp test tag name, which is checked after it's entered as a solution.

## PR Checklist:
- [ ] Added Unit / UI tests
- [ ] Basic Self QA

## Screenshots
Pocket — AddTagsItemTests swift 2023-01-09 09-55-11

### DIFF
--- a/Tests iOS/AddTagsItemTests.swift
+++ b/Tests iOS/AddTagsItemTests.swift
@@ -55,10 +55,13 @@ class AddTagsItemTests: XCTestCase {
         let addTagsView = app.addTagsView.wait()
         addTagsView.wait()
         addTagsView.newTagTextField.tap()
-        addTagsView.newTagTextField.typeText("Tag 1")
+        addTagsView.newTagTextField.typeText(XCUIKeyboardKey.delete.rawValue)
+        let randomInt = Int.random(in: 1..<155)
+        let tagInt = String(randomInt)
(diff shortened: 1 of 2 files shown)


Tests iOS/AddTagsItemTests.swift
```diff
@@ -55,10 +55,13 @@ class AddTagsItemTests: XCTestCase {
         let addTagsView = app.addTagsView.wait()
         addTagsView.wait()
         addTagsView.newTagTextField.tap()
-        addTagsView.newTagTextField.typeText("Tag 1")
+        addTagsView.newTagTextField.typeText(XCUIKeyboardKey.delete.rawValue)
```
cyndichin, line 58: FYI - test still pass after deleting this line for me, so not sure if we need to have, but I'm okay with keeping it in as well!

PR #358, opened 2023-01-12
fix(search): fix caching issues with switching scopes

cyndichin:
## Summary
Fix issue where switching search scopes too quickly causes wrong results due to caching the results for the wrong scope

## References 
IN-1127

## Implementation Details
Previously, we were using two types of search (online, local) with one cache. We learned that this was a data issue in terms of how we were setting cached results and updating search results. From the meeting, we decided it was best to follow the [SOLID principles](https://www.freecodecamp.org/news/solid-principles-explained-in-plain-english/) and create four different instances that act as containers for the 4 search types (local saves, online saves, online archive, online all items). Each container will have its own cache and this will avoid the issue of having results from one scope appear in another. Overall, this decouples each scope from one another.

## Test Steps
1. Go to Search
2. Switch Between Scopes Quickly
3. Search results for all items are correct and cache properly

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/PocketKit/MyList/SearchItemsList/LocalSavesSearch.swift
+++ b/PocketKit/Sources/PocketKit/MyList/SearchItemsList/LocalSavesSearch.swift
@@ -1,0 +1,27 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import Sync
+
+class LocalSavesSearch {
+    private let source: Source
+    private var cache: [String: [SearchItem]] = [:]
+
+    init(source: Source) {
+        self.source = source
+    }
+
+    func search(with term: String) -> [SearchItem] {
+        guard cache[term] == nil else {
+            return cache[term] ?? []
+        }
+        let items = source.searchSaves(search: term)?.compactMap { SearchItem(item: $0) } ?? []
+        cache[term] = items
+        return cache[term] ?? []
+    }
+
+    func clear() {
+        cache = [:]
+    }
+}
--- a/PocketKit/Sources/PocketKit/MyList/SearchItemsList/LocalSavesSearch.swift
+++ b/PocketKit/Sources/PocketKit/MyList/SearchItemsList/LocalSavesSearch.swift
@@ -20,8 +20,4 @@ class LocalSavesSearch {
         cache[term] = items
         return cache[term] ?? []
     }
-
-    func clear() {
-        cache = [:]
-    }
 }
--- a/PocketKit/Sources/PocketKit/MyList/SearchItemsList/OnlineSearch.swift
+++ b/PocketKit/Sources/PocketKit/MyList/SearchItemsList/OnlineSearch.swift
@@ -14,11 +14,7 @@ class OnlineSearch {
     private let scope: SearchScope
 
     @Published
-    var results: [SearchItem]? {
-        didSet {
-            subscriptions = []
-        }
-    }
+    var results: [SearchItem]?
 
     init(source: Source, scope: SearchScope) {
         self.source = source
@@ -42,10 +38,5 @@ class OnlineSearch {
         Task {
             await searchService.search(for: term, scope: scope)
         }
-    }
-
-    func clear() {
-        cache = [:]
-        subscriptions = []
     }
 }
--- a/PocketKit/Sources/PocketKit/MyList/SearchItemsList/OnlineSearch.swift
+++ b/PocketKit/Sources/PocketKit/MyList/SearchItemsList/OnlineSearch.swift
@@ -1,0 +1,51 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import Sync
+import Combine
+import SharedPocketKit
+
+class OnlineSearch {
+    private let source: Source
+    private let searchService: SearchService
+    private var subscriptions: [AnyCancellable] = []
+    private var cache: [String: [SearchItem]] = [:]
+    private let scope: SearchScope
+
+    @Published
+    var results: [SearchItem]? {
+        didSet {
+            subscriptions = []
+        }
+    }
+
+    init(source: Source, scope: SearchScope) {
+        self.source = source
+        self.searchService = source.makeSearchService()
+        self.scope = scope
+    }
+
+    func search(with term: String) {
+        guard cache[term] == nil else {
+            results = cache[term] ?? []
+            return
+        }
+
+        searchService.results.sink { [weak self] items in
+            guard let self else { return }
+            let searchItems = items.compactMap { SearchItem(item: $0) }
+            self.cache[term] = searchItems
+            self.results = self.cache[term] ?? []
+        }.store(in: &subscriptions)
+
+        Task {
+            await searchService.search(for: term, scope: scope)
+        }
+    }
+
+    func clear() {
+        cache = [:]
+        subscriptions = []
+    }
+}
--- a/PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchViewModel.swift
+++ b/PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchViewModel.swift
@@ -1,3 +1,7 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
 import UIKit
 import Network
 import Sync
@@ -11,7 +15,12 @@ class SearchViewModel: ObservableObject {
     private let user: User
     private let userDefaults: UserDefaults
     private let source: Source
-    private let searchService: SearchService
+
+    private let savesLocalSearch: LocalSavesSearch
+    private let savesOnlineSearch: OnlineSearch
+    private let archiveOnlineSearch: OnlineSearch
+    private let allOnlineSearch: OnlineSearch
+
     private var isOffline: Bool {
         networkPathMonitor.currentNetworkPath.status == .unsatisfied
     }
@@ -21,13 +30,6 @@ class SearchViewModel: ObservableObject {
     }
 
     private var selectedScope: SearchScope = .saves
-    private var currentSearchTerm: String = ""
-
-    private var cachedSearchResults: [SearchScope: [String: [SearchItem]]] = [
-        .saves: [:],
-        .archive: [:],
-        .all: [:]
-    ]
 
     @Published
     var emptyState: EmptyStateViewModel?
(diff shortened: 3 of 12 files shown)


PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchType.swift
```diff
@@ -0,0 +1,140 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import Sync
+import Combine
+import SharedPocketKit
+
+class SavesOnlineSearch {
```
cyndichin, line 9: will separate these classes into their own files before putting PR in ready to review

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchViewModel.swift
```diff
@@ -71,7 +68,11 @@ class SearchViewModel: ObservableObject {
         self.userDefaults = userDefaults
         self.source = source
         networkPathMonitor.start(queue: .global())
-        self.searchService = source.makeSearchService()
+
+        self.savesLocalSearch = SavesLocalSearch(source: source)
+        self.savesOnlineSearch = SavesOnlineSearch(source: source)
+        self.archiveOnlineSearch = ArchiveOnlineSearch(source: source)
+        self.allOnlineSearch = AllOnlineSearch(source: source)
```
cyndichin, line 75: is this preferred? I prefer the following to avoid duplicate code and still allows separate instances of the object so that caching is scope specific. Let me know if I a missing something.
```
self.savesOnlineSearch = OnlineSearch(source: source, scope: .saves)
 self.archiveOnlineSearch = OnlineSearch(source: source, scope: .archive)
 self.allOnlineSearch = OnlineSearch(source: source, scope: .all)
```

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchViewModel.swift
```diff
@@ -110,32 +105,34 @@ class SearchViewModel: ObservableObject {
...
-            self.searchResults = items.compactMap { SearchItem(item: $0) }
-        }.store(in: &subscriptions)
-        Task {
-            await searchService.search(for: term, scope: selectedScope)
+    private func submitSearch(with term: String) {
+        switch selectedScope {
+        case .saves:
+            // TODO: Handle Offline for Premium https://getpocket.atlassian.net/browse/IN-971
+            guard isPremium else {
+                self.searchResults = savesLocalSearch.search(with: term)
```
cyndichin, line 119: By setting all these to `self.searchResults` we still get the same issue where one search may be updating the value for a different scope. However, I am unable to see this visually as we get the correct caching now for the scopes. Do we prefer to separate out `self.searchResults` to `self.searchResultsForSaves`, `self.searchResultsForArchive`, `self.searchResultsForAll`?

This a small case I feel and wanted to double check with the team on the overhead of adding a results property for each scope vs recycling.

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchViewModel.swift
timc-mozilla: I don't quite understand what the difference is.  You're going to end up with `OnlineSearch` and `OfflineSearch`. So it'll end up being the same amount of work?

Also, in this case, why are we using self?  I don't see a need for self to be used.

dskuza: One can get caught up in initializers doing that "unnecessarily". Typically, it's used because an initializer argument label matches that of a property name for the type being initializes, which makes it necessary. In this case, there's nothing matching, so it can be omitted.

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchType.swift
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    func search(with term: String) {
+        guard self.cache[term] == nil else {
+            self.results = self.cache[term] ?? []
+            return
+        }
+
+        searchService.results.sink { [weak self] (items, scope) in
+            guard let self else { return }
```
dskuza, line 30: We can add an additional guard to ignore all scopes that are _not_ just saves.

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchType.swift
```diff
@@ -0,0 +1,140 @@
...
```
dskuza, line 89: We can add an additional guard to ignore all scopes that are not just archived.

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchType.swift
```diff
@@ -0,0 +1,140 @@
...
```
dskuza, line 126: We can add an additional guard to ignore all scopes that are not just all.

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchViewModel.swift
```diff
@@ -110,32 +105,34 @@ class SearchViewModel: ObservableObject {
...
+                return
+            }
+            savesOnlineSearch.search(with: term)
+            savesOnlineSearch.$results.receive(on: DispatchQueue.main).sink { items in
+                self.searchResults = items
```
dskuza, line 124: We still want to be weary of setting this _every_ time we receive results. We want to make sure we're setting this value only if the currently selected scope matches that of the new search helper time.

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchViewModel.swift
cyndichin: Discussed on Zoom:
* Will update to use to using this convention: https://github.com/kodecocodes/swift-style-guide#use-of-self
* Will update to using one class type for the Online searches

PocketKit/Sources/PocketKit/MyList/SearchItemsList/LocalSearch.swift
```diff
@@ -0,0 +1,27 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import Sync
+
+class LocalSearch {
```
dskuza, line 7:
```suggestion
class LocalSavesSearch {
```

Just to clarify that it's a local search _for saves_, and not archived items.

dskuza (approved):
:+1: I think this is in line with what we discussed previously. I'd suggest having @nzeltzer take a quick peek, to see if it aligns with what her suggestions were.

PocketKit/Sources/PocketKit/MyList/SearchItemsList/LocalSearch.swift
cyndichin: good point, updated! I'll request her review as well - thanks for the quick response!

PocketKit/Sources/PocketKit/MyList/SearchItemsList/LocalSavesSearch.swift
```diff
@@ -0,0 +1,27 @@
...
+    func search(with term: String) -> [SearchItem] {
+        guard cache[term] == nil else {
+            return cache[term] ?? []
+        }
+        let items = source.searchSaves(search: term)?.compactMap { SearchItem(item: $0) } ?? []
+        cache[term] = items
+        return cache[term] ?? []
+    }
+
+    func clear() {
```
nzeltzer, line 24: I would recommend dropping this functionality. By making a search re-usable, you encourage accidents. Instead, treat them as disposable objects.

PocketKit/Sources/PocketKit/MyList/SearchItemsList/OnlineSearch.swift
```diff
@@ -0,0 +1,51 @@
...
+            self.cache[term] = searchItems
+            self.results = self.cache[term] ?? []
+        }.store(in: &subscriptions)
+
+        Task {
+            await searchService.search(for: term, scope: scope)
+        }
+    }
+
+    func clear() {
```
nzeltzer, line 47: Same. Let's make searches disposable objects and discourage re-use through by opinionated API.

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchViewModel.swift
```diff
@@ -110,32 +112,38 @@ class SearchViewModel: ObservableObject {
     func clear() {
         searchResults = []
         subscriptions = []
+        savesLocalSearch.clear()
+        savesOnlineSearch.clear()
+        archiveOnlineSearch.clear()
+        allOnlineSearch.clear()
```
nzeltzer, line 118: Instead of clearing, let's instantiate new search instances. This prevents the likelihood of us accidentally perpetuating state from one search into the next.

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchViewModel.swift
```diff
@@ -110,32 +112,38 @@ class SearchViewModel: ObservableObject {
...
-
-    private func submitOnlineSearch(with term: String) {
-        clear()
-        searchService.results.receive(on: DispatchQueue.main).sink { [weak self] items in
-            guard let self else { return }
-            self.searchResults = items.compactMap { SearchItem(item: $0) }
-        }.store(in: &subscriptions)
-        Task {
-            await searchService.search(for: term, scope: selectedScope)
+    private func submitSearch(with term: String) {
```
nzeltzer, line 121: Recommend making state a function parameter, rather than checking an instance variable. By removing a local variable, you can simplify your control logic, decrease the likelihood of errors, and make the function easier to debug and test. You can expand on that logic to make this a pure function if you want. One of the key advantages being that testing pure functions is trivial.

nzeltzer (changes requested):
@cyndichin, great work here. I think we can go one step further and reduce the likelihood of downstream problems by not allowing ourselves to re-use searches, and instead of treating them as disposables. Furthermore, if you want to write code that's truly immune to state-based bugs, you could rewrite your master search function to be a pure function: one where there the only data operated on is data passed in as function parameters. 

Happy to discuss live! Really love how close we are to foreclosing the class of bug we saw before!

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchViewModel.swift
cyndichin: Not too sure if I understand this fully. I updated the PR with new changes, but let me know if I missed something and happy to get on a call. For a pure function, it seems that the same input always return the same result, but we don't return the results due the nature of pagination from the online search.

nzeltzer: You are correct. 

This looks like a meaningful improvement to me.

nzeltzer (approved):

PR #359, opened 2023-01-12
feat(search): saved search offline notification

timc-mozilla:
show banner for offline notification on saves

## Summary
* when the user is offline during a saves search, a banner will show if the user is premium and offline

## References 
*  [IN-1067](https://getpocket.atlassian.net/jira/software/projects/IN/boards/80?selectedIssue=IN-1067)

## Implementation Details
* added new notification banner in swift UI

## Test Steps
* if the user is offline and premium, a banner show show

## PR Checklist:
- [ ] Added Unit / UI tests
- [x] Basic Self QA

## Screenshots

[IN-1067]: https://getpocket.atlassian.net/browse/IN-1067?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchView.swift
+++ b/PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchView.swift
@@ -29,6 +29,8 @@ struct ResultsView: View {
 
     @State private var showingAlert = false
 
+    @State var bannerData: BannerModifier.BannerData = BannerModifier.BannerData(image: .looking, title: "Limited search results", detail: "You can only search titles and URLs while offline. Connect to the internet to use Premium's full-text search.")
+
     var body: some View {
         List(results, id: \.id) { item in
             HStack {
@@ -46,6 +48,7 @@ struct ResultsView: View {
         }
         .listStyle(.plain)
         .accessibilityIdentifier("search-results")
+        .banner(data: $bannerData, show: $viewModel.isPremiumAndOffline)
         .alert(isPresented: $showingAlert) {
             Alert(title: Text("You must have an internet connection to view this item."), dismissButton: .default(Text("OK")))
         }
--- a/PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchViewModel.swift
+++ b/PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchViewModel.swift
@@ -34,6 +34,8 @@ class SearchViewModel: ObservableObject {
         return user.status == .premium
     }
 
+    var isPremiumAndOffline: Bool = false
+
     private var selectedScope: SearchScope = .saves
 
     @Published
@@ -86,6 +88,8 @@ class SearchViewModel: ObservableObject {
         savesOnlineSearch = OnlineSearch(source: source, scope: .saves)
         archiveOnlineSearch = OnlineSearch(source: source, scope: .archive)
         allOnlineSearch = OnlineSearch(source: source, scope: .all)
+
+        isPremiumAndOffline = isPremium && isOffline
     }
 
     func updateScope(with scope: SearchScope, searchTerm: String? = nil) {
--- a/PocketKit/Sources/PocketKit/Banner/BottomNotification.swift
+++ b/PocketKit/Sources/PocketKit/Banner/BottomNotification.swift
@@ -1,0 +1,70 @@
+import SwiftUI
+import Textile
+import UIKit
+
+struct BannerModifier: ViewModifier {
+    struct BannerData {
+        var image: ImageAsset
+        var title: String
+        var detail: String
+    }
+
+    @Binding var data: BannerData
+    @Binding var show: Bool
+
+    func body(content: Content) -> some View {
+        ZStack(alignment: .bottom) {
+            content
+            if show {
+                VStack {
+                    HStack {
+                        Image(asset: data.image)
+                            .resizable()
+                            .frame(width: 83, height: 50, alignment: .leading)
+                            .padding(8)
+                        VStack(alignment: .leading, spacing: 2) {
+                            Text(data.title)
+                                .style(.title)
+                            Text(data.detail)
+                                .style(.subtitle)
+                        }
+                        Spacer()
+                    }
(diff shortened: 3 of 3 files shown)


PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchViewModel.swift
```diff
@@ -17,7 +17,13 @@ class SearchViewModel: ObservableObject {
...
     private var isPremium: Bool {
-        return user.status == .premium
+        user.status == .premium
+    }
+
+    var _isPremiumAndOffline: Bool = false
+    var isPremiumAndOffline: Bool {
+        get { return _isPremiumAndOffline }
+        set { self._isPremiumAndOffline = newValue }
     }
```
cyndichin, line 27: could we do `var isPremiumAndOffline: Bool = false` instead of having two properties?

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchView.swift
```diff
@@ -19,7 +19,12 @@ struct SearchView: View {
 
 // MARK: - Search Results Component
 struct ResultsView: View {
+    @ObservedObject
+    var viewModel: SearchViewModel
     var results: [SearchItem]
+
+    @State var bannerData: BannerModifier.BannerData = BannerModifier.BannerData(image: .looking, title: "Limited search results", detail: "You can only search titles and URLs while offline. Connect to the internet to use Premium's full-text search.")
```
cyndichin, line 26: what do you think about encapsulating this logic in the banner since the results view shouldn't have to know this detail?

cyndichin (commented):
yay for SwiftUI! left a few comments and I'm having trouble seeing the banner on my iPhone 8. What steps should I take? I have my phone offline -> search and tapped on "pocket" in my recent searches. thanks!

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchViewModel.swift
timc-mozilla: I tried doing that, but I had problems with using a var property without setting get/set.

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchView.swift
timc-mozilla: Because we want to make the banner reusable, having a level of separation would be good imo. I also dont think there is any logic involved other than passing in the image, and two strings.

cyndichin: sounds good! Do you think this would be better to add to the `viewModel` so that we don't have the data in the View. A `BannerViewModel` may be overkill for now. Don't have to do this now, but curious to hear your perspective.

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchViewModel.swift
cyndichin: IMG_D06BC6A4FF9B-1

cyndichin (changes requested):
Banner looks good in light mode on iPad! A couple of comments and request change to not remove the Package files

Pocket.xcodeproj/project.xcworkspace/xcshareddata/WorkspaceSettings.xcsettings
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>PreviewsEnabled</key>
+	<false/>
+</dict>
+</plist>
```
cyndichin, line 8: do we need this change?

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchView.swift
timc-mozilla: I dont think it would provide much benefit with having a VM, but I'm open to doing it if needed.  It just seems like too much separation imo.  happy to talk further about when/what we should use a VM for though.

Pocket.xcodeproj/project.xcworkspace/xcshareddata/WorkspaceSettings.xcsettings
timc-mozilla: wont build on my end without it.

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchView.swift
cyndichin: ok! lets keep this as is for now

Pocket.xcodeproj/project.xcworkspace/xcshareddata/WorkspaceSettings.xcsettings
cyndichin: weird, why is that? what if you checkout this change?

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchViewModel.swift
cyndichin: thanks for the update on the boolean, in terms of dark mode, we can handle this after doing design QA

cyndichin (commented):
looks good to me! only outstanding is that I don't think we need to commit: 
- Pocket.xcodeproj/project.xcworkspace/xcshareddata/WorkspaceSettings.xcsettings
- let's not delete `Package.resolved`, should be fixed after rebasing

cyndichin (approved):
looks great! thanks for the hard work 🔥

timc-mozilla:
all tests pass locally

PR #361, opened 2023-01-17
ci(tests): Reader tests and small fix

audie1:
## Summary
Creates a new set of Reader tests, fixes a small tags test.

## References 
Merging:
https://github.com/Pocket/pocket-ios/pull/334
https://github.com/Pocket/pocket-ios/pull/355

## Implementation Details
See PR's above.


cyndichin (approved):
Looks like the same as the other Reader PR - so approving.

PR #362, opened 2023-01-19
project(spm): update resolved files

cyndichin:
## Summary
Update Package.resolved files to match the exact versions

## References 
IN-1126

## Implementation Details
Removed the Package.resolved files and run the command line to resolve packages `swift package resolve`

Ref: https://blog.eidinger.info/swift-package-manager-understand-resolve-reset-and-update

## Test Steps
- [ ] Verify that you no longer see the Package.resolved issue and that the exact versions match the ones specified in `Package.swift`

## PR Checklist:
- N/A Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,61 +1,59 @@
 {
-  "object": {
-    "pins": [
-      {
-        "package": "DangerSwiftCoverage",
-        "repositoryURL": "https://github.com/f-meloni/danger-swift-coverage",
-        "state": {
-          "branch": null,
-          "revision": "6a21a6e595e2b13443428942ce128b3ba6b827ee",
-          "version": "1.2.1"
-        }
-      },
-      {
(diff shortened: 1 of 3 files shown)


timc-mozilla (approved):
worked for me. i think the real test would be merging this branch in and seeing if other branches break.  I wonder if there is a risk for any residual outdated package.resolved files from other branches that could continue this problem.

cyndichin:
> worked for me. i think the real test would be merging this branch in and seeing if other branches break. I wonder if there is a risk for any residual outdated package.resolved files from other branches that could continue this problem.

yeah, we'll need to make sure to rebase the branches with this change, but would be curious to see if any additional changes would be needed. Seems like the Xcode UI is not as reliable so I would defer to the command line. thanks for the review!

PR #366, opened 2023-01-23
feat(search): tapping on search item navigates to article

cyndichin:
## Summary
Tapping on a search result opens the item in the best view

## References 
IN-972

## Implementation Details
When the user taps on an item it calls the `select` function which retrieves or creates a `SavedItem` using a new function in `Source` called `fetchOrCreateSavedItem`. In order to create a `SavedItem` at a later date, we needed a way to get the `SavedItemParts` details which is why we created `SearchSavedItem` and modified `ItemsListItem` to pass this data. Along with this work, I decided to use the existing graphQL fragments even if they contain a deprecated version on `topImageUrl`. Instead of making the updates here, I will create a new ticket to handle that update.

Overall, utilized the same logic from our Saves / Archives lists in presenting the best view for an item and moved it into the `SearchViewModel`. The logic works well for items saved locally, but not for items presented via online search, so I added a workaround in downloading the item as a saved item if a user taps on it. I also investigated creating a new model and presenting the Reader view without having to rely on locally saved items, but this proved to be a large restructuring, so opted for the workaround after time boxing this. 

Going forward, we should discuss how we want to refactor the logic in presenting the Reader view as well as think if we want to create a model where the reader may not have to rely on locally saved items. 

## Test Steps
- [ ] WHEN I tap on a search result
THEN we open the reader
AND display the selected article with the best view

- [ ] WHEN the device is offline
AND NOT available locally
AND the selected article isn’t available offline
THEN we display a system alert message

”You must have an internet connection to view this item”

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

## Screenshots
|Select Search Item|
|------|
|Select Search Item|

Note: my internet connection was super weak, so not everything loads in video capture


cyndichin:
> overall lgtm

thanks for the quick review @timc-mozilla!

PR #372, opened 2023-01-31
feat(search): search locally for saves if online fails

cyndichin:
## Summary
An online search of saves that fails because for network reasons or other errors is resubmitted as an offline search

## References 
IN-971

## Implementation Details
Converted the `search` function to throw an error if it encounters one in `PocketSearchService` . We pass this error on in our OnlineSearch class and we use the `Result` type to return whether the search failed or succeeded. If there is a failure, then run local search for saves only in our `SearchViewModel`. Updated UI banner with some refinements to match Figma and figured out bug with animation in SwiftUI as we needed to specific zIndex(-1).

## Test Steps
- [ ] WHEN the user submits a search query
AND the network connection is active
AND the connection fails
THEN re-submit the search query as a search of local data
AND display the offline search message (banner)

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/PocketKit/Banner/BottomNotification.swift
+++ b/PocketKit/Sources/PocketKit/Banner/BottomNotification.swift
@@ -9,7 +9,7 @@ struct BannerModifier: ViewModifier {
         var detail: String
     }
 
-    @Binding var data: BannerData
+    let data: BannerData
     @Binding var show: Bool
 
     func body(content: Content) -> some View {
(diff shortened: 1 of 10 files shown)


cyndichin:
This work is mainly to handle resubmitting local search if there is an issue with online search. Error handling will be addressed in a future ticket. Thanks for reviewing in advance!

timc-mozilla (approved):

PR #379, opened 2023-02-07
feat(premium): tos and privacy links

timc-mozilla:
## Summary
* Adding Privacy Policy and ToS links

## References 
* [IN-1051](https://getpocket.atlassian.net/browse/IN-1051)

## Implementation Details
* added Safari View to allow users to stay on current stack of screens

## Test Steps
* As a free user, go to settings
* select "Go Premium"
* Click on "Privacy Policy" and/or "Terms of Services"

## PR Checklist:
- [x] Basic Self QA

## Screenshots

[IN-1051]: https://getpocket.atlassian.net/browse/IN-1051?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/PocketKit/Sources/PocketKit/Premium/PremiumUpgradeView.swift
+++ b/PocketKit/Sources/PocketKit/Premium/PremiumUpgradeView.swift
@@ -1,5 +1,6 @@
 import SwiftUI
 import Textile
+import SafariServices
 
 struct PremiumUpgradeView: View {
     @Environment(\.dismiss)
@@ -175,10 +176,28 @@ Refunds are not available for unused portions of a subscription.
 }
 
 private struct PremiumTermsView: View {
+    @State var showSafari = false
+    @State var urlString = "https://getpocket.com/en/privacy/"
+    // TODO: URLs need to be localized
+    let privacyPolicyUrlString = "https://getpocket.com/en/privacy/"
+    let toSUrlString = "https://getpocket.com/en/tos/"
+
     var body: some View {
         HStack(spacing: 16) {
-            Button { } label: { Text("Privacy Policy").style(.terms) }
-            Button { } label: { Text("Terms of Service").style(.terms) }
+            Button(action: {
+                self.urlString = privacyPolicyUrlString
+                self.showSafari = true
+            }, label: { Text("Privacy Policy").style(.terms) })
+                .sheet(isPresented: $showSafari) {
+                    SafariView(url: URL(string: self.urlString)!)
+                }
+            Button(action: {
+                self.urlString = toSUrlString
+                self.showSafari = true
+            }, label: { Text("Terms of Service").style(.terms) })
+                .sheet(isPresented: $showSafari) {
+                    SafariView(url: URL(string: self.urlString)!)
+                }
         }.padding(.bottom)
     }
 }
@@ -202,6 +221,17 @@ private struct PremiumBackgroundView: View {
 
     private var borderWidth: CGFloat {
         CGFloat(UIDevice.current.userInterfaceIdiom == .pad ? 18.0 : 13.0)
+    }
+}
+
+struct SafariView: UIViewControllerRepresentable {
+    let url: URL
+
+    func makeUIViewController(context: UIViewControllerRepresentableContext<SafariView>) -> SFSafariViewController {
+        return SFSafariViewController(url: url)
+    }
+
+    func updateUIViewController(_ uiViewController: SFSafariViewController, context: UIViewControllerRepresentableContext<SafariView>) {
     }
 }
 
--- a/PocketKit/Sources/PocketKit/Premium/PremiumUpgradeView.swift
+++ b/PocketKit/Sources/PocketKit/Premium/PremiumUpgradeView.swift
@@ -177,28 +177,48 @@ Refunds are not available for unused portions of a subscription.
 
 private struct PremiumTermsView: View {
     @State var showSafari = false
-    @State var urlString = "https://getpocket.com/en/privacy/"
-    // TODO: URLs need to be localized
-    let privacyPolicyUrlString = "https://getpocket.com/en/privacy/"
-    let toSUrlString = "https://getpocket.com/en/tos/"
 
     var body: some View {
         HStack(spacing: 16) {
             Button(action: {
-                self.urlString = privacyPolicyUrlString
                 self.showSafari = true
             }, label: { Text("Privacy Policy").style(.terms) })
-                .sheet(isPresented: $showSafari) {
-                    SafariView(url: URL(string: self.urlString)!)
+            .sheet(isPresented: $showSafari) {
+                if let privacyUrl = getUrlFor(typeOf: .PrivacyPolicy) {
+                    SFSafariView(url: privacyUrl)
                 }
+            }
+            .accessibilityIdentifier("privacy-policy")
             Button(action: {
-                self.urlString = toSUrlString
                 self.showSafari = true
             }, label: { Text("Terms of Service").style(.terms) })
-                .sheet(isPresented: $showSafari) {
-                    SafariView(url: URL(string: self.urlString)!)
+            .sheet(isPresented: $showSafari) {
+                if let toSUrl = getUrlFor(typeOf: .TermsOfService) {
+                    SFSafariView(url: toSUrl)
                 }
+            }
+            .accessibilityIdentifier("terms-of-service")
         }.padding(.bottom)
+    }
+
+    enum Link {
+        case PrivacyPolicy
+        case TermsOfService
+    }
+
+    private func getUrlFor(typeOf: Link) -> URL? {
+        switch typeOf {
+        case .PrivacyPolicy:
+            guard let privacyUrl = URL(string: "https://getpocket.com/privacy/") else {
+                return nil
+            }
+            return privacyUrl
+        case .TermsOfService:
+            guard let toSUrl = URL(string: "https://getpocket.com/tos/") else {
+                return nil
+            }
+            return toSUrl
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


PocketKit/Sources/PocketKit/Premium/PremiumUpgradeView.swift
```diff
@@ -175,10 +176,28 @@ Refunds are not available for unused portions of a subscription.
 }
 
 private struct PremiumTermsView: View {
+    @State var showSafari = false
+    @State var urlString = "https://getpocket.com/en/privacy/"
+    // TODO: URLs need to be localized
+    let privacyPolicyUrlString = "https://getpocket.com/en/privacy/"
+    let toSUrlString = "https://getpocket.com/en/tos/"
```
Gio2018, line 183: can we set these as `private`?

Gio2018 (commented):
Hey @timc-mozilla , the terms of service and privacy policy screen show as expected. One thing I noticed is that, if on those safari views I tap on the user Icon, the user here is not logged in and it shows an app banner (see screenshot). I don't think these need to be addressed in this PR, but we might want to track them as they can be confusing for the user, what do you think?
<p align=center>

</p>



~~Also left a code comment~~ nvm, I amended the commend

PocketKit/Sources/PocketKit/Premium/PremiumUpgradeView.swift
```diff
@@ -175,10 +176,28 @@ Refunds are not available for unused portions of a subscription.
...
     var body: some View {
         HStack(spacing: 16) {
-            Button { } label: { Text("Privacy Policy").style(.terms) }
-            Button { } label: { Text("Terms of Service").style(.terms) }
+            Button(action: {
+                self.urlString = privacyPolicyUrlString
+                self.showSafari = true
+            }, label: { Text("Privacy Policy").style(.terms) })
+                .sheet(isPresented: $showSafari) {
+                    SafariView(url: URL(string: self.urlString)!)
```
cyndichin, line 192: force unwrapping can be dangerous, should we check if this value is non optional first?

PocketKit/Sources/PocketKit/Premium/PremiumUpgradeView.swift
```diff
@@ -175,10 +176,28 @@ Refunds are not available for unused portions of a subscription.
...
+                }
+            Button(action: {
+                self.urlString = toSUrlString
+                self.showSafari = true
+            }, label: { Text("Terms of Service").style(.terms) })
+                .sheet(isPresented: $showSafari) {
+                    SafariView(url: URL(string: self.urlString)!)
```
cyndichin, line 199: why not use `SFSafariView(url: webURL)` directly? Similar to how we use it in `EmptyStateView`

PocketKit/Sources/PocketKit/Premium/PremiumUpgradeView.swift
bassrock: @timc-mozilla lets remove the `/en/` portion of the URL. That way if we ever localize these, the user gets that version automatically. Then you can remove the TODO you have.

PocketKit/Sources/PocketKit/Premium/PremiumUpgradeView.swift
timc-mozilla: because I didnt know that `SFSafariView` existed.  thanks for letting me know.

PocketKit/Sources/PocketKit/Premium/PremiumUpgradeView.swift
timc-mozilla: gotcha gotcha. thanks

PocketKit/Sources/PocketKit/Premium/PremiumUpgradeView.swift
```diff
@@ -175,12 +176,51 @@ Refunds are not available for unused portions of a subscription.
...
+                    SFSafariView(url: toSUrl)
+                }
+            }
+            .accessibilityIdentifier("terms-of-service")
         }.padding(.bottom)
     }
+
+    enum Link {
+        case PrivacyPolicy
+        case TermsOfService
```
cyndichin, line 207: nit: enum cases usually lower camel case
https://github.com/kodecocodes/swift-style-guide#naming

PocketKit/Sources/PocketKit/Premium/PremiumUpgradeView.swift
```diff
@@ -175,12 +176,51 @@ Refunds are not available for unused portions of a subscription.
...
+    }
+
+    private func getUrlFor(typeOf: Link) -> URL? {
+        switch typeOf {
+        case .PrivacyPolicy:
+            guard let privacyUrl = URL(string: "https://getpocket.com/privacy/") else {
+                return nil
+            }
+            return privacyUrl
```
cyndichin, line 216: can also do `return URL(string: "https://getpocket.com/privacy/")` instead of using the guard because we're return an optional. In fact, I'm okay with doing away with the enum entirely and just using 

```
if let privacyUrl = URL(string: "https://getpocket.com/privacy/") {
                     SFSafariView(url: privacyUrl)
                 }
                 ```

File: PocketKit/Sources/PocketKit/Premium/PremiumUpgradeView.swift [hunk 6]
```diff
@@ -175,12 +176,51 @@ Refunds are not available for unused portions of a subscription.
...
+            .accessibilityIdentifier("privacy-policy")
+            Button(action: {
+                self.showTermsOfService = true
+            }, label: { Text("Terms of Service").style(.terms) })
+            .sheet(isPresented: $showTermsOfService) {
+                if let toSUrl = getUrlFor(typeOf: .TermsOfService) {
+                    SFSafariView(url: toSUrl)
+                }
+            }
+            .accessibilityIdentifier("terms-of-service")
```
cyndichin, line 201: since we have accessibility identifiers, will we also be writing tests?

cyndichin (commented):
a couple of comments, none blockers. Only blocker is wondering if we should be adding tests for this, especially since we accessibility identifiers for them now.

PocketKit/Sources/PocketKit/Premium/PremiumUpgradeView.swift
```diff
@@ -1,5 +1,6 @@
 import SwiftUI
 import Textile
+import SafariServices
```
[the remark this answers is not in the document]
timc-mozilla: removed.

PocketKit/Sources/PocketKit/Premium/PremiumUpgradeView.swift
bassrock: @cyndichin can we enforce this with Swiftlint in another PR? Seems like we should.

PocketKit/Sources/PocketKit/Premium/PremiumUpgradeView.swift
bassrock: General guard question here. We are defining the URL, I think Swfit's interface here is bad. I'd rather us force unwrap in this case and fail hard and fast if we are defining the input URL. No need to guard something we know will always be true.

bassrock (approved):
I'm comfortable moving this forward and addressing the comments here in a followup ticket. @timc-mozilla can you ticket the followups up?

PocketKit/Sources/PocketKit/Premium/PremiumUpgradeView.swift
Gio2018: But if for some reason that URL is not accessible, we'd crash the app instead of just not showing the privacy, right?

PocketKit/Sources/PocketKit/Premium/PremiumUpgradeView.swift
timc-mozilla: https://getpocket.atlassian.net/jira/software/projects/IN/boards/80/backlog

PocketKit/Sources/PocketKit/Premium/PremiumUpgradeView.swift
timc-mozilla: as discussed prior in other meetings, since settings is mostly all redirected links, there's no need to write tests for settings screen.

PocketKit/Sources/PocketKit/Premium/PremiumUpgradeView.swift
cyndichin: If we want to go through the failing hard and fast route, then I think we should capture the error. I personally don't like having force unwraps and rather have the button do nothing vs crash. However if we are okay using forced unwraps going forward for defined URLs, we can do that and I'll add it to our wiki.

It's very unlikely to crash, but still possible right?

PocketKit/Sources/PocketKit/Premium/PremiumUpgradeView.swift
cyndichin: hmmm I don't see a SwiftLint rule for this, but we're planning on having a discussion relating to this. 

https://pocket.slack.com/archives/C01UH57EJKD/p1676391464059509

This is what I'm used to in Swift convention, but happy to revisit as a team.

```
## Naming

### Enum
When it comes to enum types, we like to use `UpperCamelCase`.
When it comes to enum cases, we like to use `lowerCamelCase`.

```
enum ReadableEvent {
    case archive
    case delete
    case contentUpdated
}
```

PocketKit/Sources/PocketKit/Premium/PremiumUpgradeView.swift
bassrock: @Gio2018 

> But if for some reason that URL is not accessible, we'd crash the app instead of just not showing the privacy, right?

I don't believe that is the case, the URL does not make any network calls. It would crash only if the value passed to URL was not a valid URL pattern. In this case, the URL is defined very strictly and passed in by us. 

 
@cyndichin 
> It's very unlikely to crash, but still possible right? 
The only crash I see happening here would be a developer error. Which we want to fail fast, and sentry will pick up.

PR #385, opened 2023-02-10
feat(analytics): explict declaration

bassrock:
## Summary
Android is working on implementing analytics in a similar manner to the [Web team](https://github.com/Pocket/web-client/blob/main/src/connectors/snowplow/actions/home.js)

In this method of analytics, we become very explicit about the data we pass, and define them in a single entry point. It also allows us to define our own analytic identifiers following a [common pattern.](https://getpocket.atlassian.net/wiki/spaces/PE/pages/2857566283/Pocket+Architecture+-+Analytics+Vision#UI-Identifier-Naming-Convention) 

All analytics should also be defined in the linked spreadsheet on confluence and collaborated with across team as needed.

See reference file: https://github.com/Pocket/Android/pull/2020/files#diff-dca572fb8586a326c9530290055514ef3797aaf8523a00c22ca54a07c5f57e57

More info coming Monday with @kirill-demtchouk 

## References 
* https://pocket.slack.com/archives/C04HWMUKLN5/p1675109088032829

## Implementation Details

## Test Steps

## PR Checklist:
- [ ] Added Unit / UI tests
- [ ] Basic Self QA


cyndichin (commented):
thanks @bassrock ! great work :) let me know how I can assist. Definitely think we should get this work in before search analytics. Shall we go through a thorough analytics test to verify the calls?

bassrock:
Bits of this pr will be broken up.

Closing for  https://github.com/Pocket/pocket-ios/pull/390

PR #403, opened 2023-02-17
ci(tests): hyperlink+tags tests flaky fix

audie1:
## Summary
Fixes for a number of small-fail flaky tests.

I've made two attempts at changing the tests so that they don't flaky fail (time will tell if these changes are effective at preventing the flaky fails):

  **Tags**: I changed this test so it opens the All Tags view instead of the item tag, to check that the save function worked

  **Hyperlink**: I changed this test to long press a different hyperlink in a different item

Tracking the fail rate via Bitise graph tracking.


bassrock (approved):
🤞

PR #405, opened 2023-02-17
feat(search): favorite search results

cyndichin:
## Summary
PR adds functionality to favorite search result items

## References 
IN-1121

## Implementation Details
* Create `PocketItemViewModel` to hold the logic of favoriting items. Similar to existing logic except we have to create a savedItem from remote data. 
* Rename `SearchItem` to `PocketItem` to be more reusuable
* Fixed bug related to new use of entities for analytics. We were using old context from APIUser and User, so created new `persistentEntities` object.

## Test Steps
- [ ] GIVEN search is active
AND a search has been performed
AND there are 1+ results
AND search results are visible
THEN every search result includes a favorite button

- [ ] GIVEN a search result is a favorite
THEN the search result favorite button is filled

- [ ] GIVEN a search result is not a favorite
THEN the search result favorite button is not filled

- [ ] GIVEN that a user taps on the search result favorite button 
AND the search result has already been favorited
THEN the search result will be unfavorited
AND the search result button will not be filled

- [ ] GIVEN that a user taps on the search result favorite button 
AND the search result has not already been favorited
THEN the search result will be favorited
AND the search result button will be filled

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/Analytics/AppEvents/Search.swift
+++ b/PocketKit/Sources/Analytics/AppEvents/Search.swift
@@ -3,7 +3,70 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import Foundation
+import SharedPocketKit
 
 public extension Events {
     struct Search {}
+}
+
+public extension Events.Search {
+    static func favoriteItem(
+        itemUrl: URL,
+        positionInList: Int,
+        scope: SearchScope
+    ) -> Event {
+        var identifier = ""
+        switch scope {
+        case .saves:
+            identifier = "global-nav.search.saves.favorite"
+        case .archive:
+            identifier = "global-nav.search.archive.favorite"
+        case .all:
+            identifier = "global-nav.search.all.favorite"
+        }
+
+        return Engagement(
+            .general,
+            uiEntity: UiEntity(
+                .button,
+                identifier: identifier,
+                index: positionInList
+            ),
+            extraEntities: [
+                ContentEntity(
+                    url: itemUrl
+                )
+            ]
+        )
+    }
+
+    static func unfavoriteItem(
+        itemUrl: URL,
+        positionInList: Int,
+        scope: SearchScope
+    ) -> Event {
+        var identifier = ""
+        switch scope {
+        case .saves:
+            identifier = "global-nav.search.saves.unfavorite"
+        case .archive:
+            identifier = "global-nav.search.archive.unfavorite"
+        case .all:
+            identifier = "global-nav.search.all.unfavorite"
+        }
+
+        return Engagement(
+            .general,
+            uiEntity: UiEntity(
+                .button,
+                identifier: identifier,
+                index: positionInList
+            ),
+            extraEntities: [
+                ContentEntity(
+                    url: itemUrl
+                )
+            ]
+        )
+    }
 }
--- a/PocketKit/Sources/PocketKit/MyList/ItemsList/PocketItemViewModel.swift
+++ b/PocketKit/Sources/PocketKit/MyList/ItemsList/PocketItemViewModel.swift
@@ -1,0 +1,86 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import SwiftUI
+import Analytics
+import Sync
+import SharedPocketKit
+
+class PocketItemViewModel: ObservableObject {
+    private let tracker: Tracker
+    private let source: Source
+    let item: PocketItem
+    let index: Int
+
+    @Published
+    var isFavorite: Bool
+
+    init(item: PocketItem, index: Int, source: Source, tracker: Tracker) {
+        self.item = item
+        self.index = index
+        self.source = source
+        self.tracker = tracker
+        self.isFavorite = item.isFavorite
+    }
+
+    func favoriteAction(index: Int, scope: SearchScope) -> ItemAction {
+        if isFavorite {
+            return .unfavorite { [weak self] _ in
+                self?._unfavorite(index: index, scope: scope)
+            }
+        } else {
+            return .favorite { [weak self] _ in
+                self?._favorite(index: index, scope: scope)
+            }
+        }
+    }
+
+    private func _favorite(index: Int, scope: SearchScope) {
+        guard
+            let id = item.id,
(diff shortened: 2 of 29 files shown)


PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchItem.swift
```diff
@@ -40,8 +50,57 @@ struct SearchItem {
         ItemAction.share { _ in Log.info("Share button tapped!") }
     }
 
-    var favoriteAction: ItemAction {
-        ItemAction.favorite { _ in Log.info("Favorite button tapped!") }
+    func favoriteAction(index: Int, scope: SearchScope) -> ItemAction {
+        if isFavorite {
+            return .unfavorite { [weak self] _ in
+                guard let self else { return }
```
bassrock, line 56: @cyndichin when will this be nil? Can we sentry log it, this feels like a case we shouldn't ever want to hit

cyndichin: This is due to weakifying self and it's a bit of a boiler plate code in Swift when handling this situation. If self is released, we don't want the closure to prevent self from being removed so we make it a weak self. If self is nil, then we don't want to trigger the code. I think it is nil when ARC frees up the memory and deallocates the instance. By doing the `guard` statement we are ensuring that we want a strong reference (instance exists) and so the object is available until execution of closure is complete at which point self can then be released. 

Not opposed to adding a sentry log, but maybe makes more sense as an `info` or `debug` message?

cc: @timc-mozilla @Gio2018 let me know if you have any additional input on this

timc-mozilla (dismissed):

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchItem.swift
cyndichin: I decided to use optional self instead as it's better readability but please let me know if you would prefer the other method. I have no strong opinions on either.

PocketKit/Sources/PocketKit/MyList/ItemsList/PocketItemViewModel.swift
```diff
@@ -0,0 +1,86 @@
...
+            let savedItem = source.fetchOrCreateSavedItem(
+                with: id,
+                and: item.remoteItemParts
+            )
+        else {
+            Log.capture(message: "Saved Item not created")
+            return
+        }
+
+        tracker.track(event: Events.Search.favoriteItem(itemUrl: savedItem.url, positionInList: index, scope: scope))
```
cyndichin, line 68: this is the only part that makes it specific to Search. I am leaving this here for now and we can refactor once we decide to use this for other lists. Open to any concerns though. thanks!

PocketKit/Sources/Analytics/AppEvents/Search.swift
```diff
@@ -3,7 +3,70 @@
...
+
+public extension Events.Search {
+    static func favoriteItem(
+        itemUrl: URL,
+        positionInList: Int,
+        scope: SearchScope
+    ) -> Event {
+        var identifier = ""
+        switch scope {
+        case .saves:
```
bassrock, line 20: yay!

PocketKit/Sources/Sync/PocketSource.swift
```diff
@@ -649,15 +649,17 @@ extension PocketSource {
     }
 
     public func fetchOrCreateSavedItem(with remoteID: String, and remoteParts: SavedItem.RemoteSavedItem?) -> SavedItem? {
+        let savedItem = (try? space.fetchSavedItem(byRemoteID: remoteID))
+
         guard let remoteParts, let url = URL(string: remoteParts.url) else {
-            Log.breadcrumb(category: "sync", level: .warning, message: "Skipping updating of SavedItem because we do not have a valid url or we have no remoteParts")
-            return nil
+            Log.breadcrumb(category: "sync", level: .debug, message: "SavedItem found and don't need to create one")
```
bassrock, line 655: @cyndichin isnt the meaning of the guard here, that we have an invalid URL? not that we don't need to create one.

PocketKit/Sources/Sync/PocketSource.swift
```diff
@@ -649,15 +649,17 @@ extension PocketSource {
...
+            return savedItem
         }
 
-        let savedItem = (try? space.fetchSavedItem(byRemoteID: remoteID)) ?? SavedItem(context: space.context, url: url, remoteID: remoteID)
-        savedItem.update(from: remoteParts, with: space)
+        let remoteSavedItem = SavedItem(context: space.context, url: url, remoteID: remoteID)
```
bassrock, line 659: Don't we still want to update the SavedItem object we pulled from CoreData with the latest data?

PocketKit/Sources/PocketKit/MyList/ItemsList/PocketItemViewModel.swift
bassrock: @cyndichin can you put this comment in the code?

PocketKit/Sources/Sync/PocketSource.swift
cyndichin: This function purpose is to check if there's a saved item already existing, otherwise we have to create it from the remote parts. If we want to check if a url exists, then would checking on the `savedItem.url` be better? A local savedItem would not have any `remoteParts`.

PocketKit/Sources/Sync/PocketSource.swift
cyndichin: This function was meant to create a savedItem based on the user interaction with the card, but let me think about this

cyndichin: I updated the savedItem with remote parts! Rather than enforcing that there is a url, we assume that url is already associated with the savedItem as we now have it as non nil in core data.

bassrock (approved):

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchItem.swift
Gio2018: I agree @cyndichin . If we don't do this, we might end up with a memory leak. Another option if we don't want to do the unwrap, would be to use `unowned`, but only if we are 100% certain that `self` will never be nil when we call that closure, otherwise the app will crash.

PR #416, opened 2023-02-21
fix(reader): makes reader use readable content width

bassrock:
## Summary

Makes the reader user-readable content width so that the reader looks better on devices like the iPad.

## References 
* IN-790

## Implementation Details
Was not sure what tests I could use here.

## Test Steps
* Use the reader on a large tablet screen.

## PR Checklist:
- [ ] Added Unit / UI tests
- [ ] Basic Self QA

## Screenshots

Before:

Simulator Screen Shot - iPad Mini - 2023-02-21 at 15 45 56

After:
Simulator Screen Shot - iPad Mini - 2023-02-21 at 15 44 43

### DIFF
--- a/PocketKit/Sources/PocketKit/Article/ReadableViewController.swift
+++ b/PocketKit/Sources/PocketKit/Article/ReadableViewController.swift
@@ -93,6 +93,12 @@ class ReadableViewController: UIViewController {
         readableViewModel.fetchDetailsIfNeeded()
     }
 
+    override func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
+        super.viewWillTransition(to: size, with: coordinator)
+        // ensure on device rotate and change the view re-draws
+        collectionView.collectionViewLayout.invalidateLayout()
+    }
+
     required init?(coder: NSCoder) {
         fatalError("Unable to instantiate \(Self.self) from xib/storyboard")
     }
@@ -263,15 +269,12 @@ extension ReadableViewController {
 
         switch index {
         case 0:
-            let availableItemWidth = environment.container.effectiveContentSize.width
-            - Constants.metaSectionContentInsets.leading
-            - Constants.metaSectionContentInsets.trailing
-
+            let availableItemWidth = view.readableContentGuide.layoutFrame.width
             let height = metadata?.size(for: availableItemWidth).height ?? 1
             let group = NSCollectionLayoutGroup.vertical(
                 layoutSize: NSCollectionLayoutSize(
                     widthDimension: .fractionalWidth(1),
-                    heightDimension: .absolute(height)
+                    heightDimension: .estimated(height)
                 ),
                 subitems: [
                     NSCollectionLayoutItem(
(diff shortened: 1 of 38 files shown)


PocketKit/Sources/PocketKit/Article/ReadableViewController.swift
```diff
@@ -93,6 +93,12 @@ class ReadableViewController: UIViewController {
         readableViewModel.fetchDetailsIfNeeded()
     }
 
+    override func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
```
bassrock, line 96: Very subtly you can see that images may use more space then they should after a rotation, this attempts to correct that by making the collection view re-layout after a size change.

PocketKit/Sources/PocketKit/Article/ReadableViewController.swift
```diff
@@ -284,20 +287,26 @@ extension ReadableViewController {
             )
 
             let section = NSCollectionLayoutSection(group: group)
-            section.contentInsets = Constants.metaSectionContentInsets
+            // Zero out the default leading/trailing contentInsets, but preserve the default top/bottom values.
```
bassrock, line 290: Used https://developer.apple.com/forums/thread/679863 as my guide.

PocketKit/Sources/PocketKit/Article/ReadableViewController.swift
```diff
@@ -313,7 +322,15 @@ extension ReadableViewController {
...
-            section.contentInsets = Constants.contentSectionContentInsets
+            // Zero out the default leading/trailing contentInsets, but preserve the default top/bottom values.
+            // This ensures each section will be inset horizontally exactly to the readable content width.
+            var contentInsets = section.contentInsets
+            contentInsets.leading = 0
+            contentInsets.trailing = 0
+            contentInsets.top = Constants.contentSectionContentInsets.top
+            contentInsets.bottom = Constants.metaSectionContentInsets.bottom
+            section.contentInsets = contentInsets
+            section.contentInsetsReference = .readableContent
```
cyndichin, line 333: nit: similar as logic above, should we put this in a private func, so we only have to change in one place

PocketKit/Tests/SyncTests/Support/Space+factories.swift
```diff
@@ -110,8 +110,6 @@ extension Space {
     }
 }
 
-
-
```
cyndichin, line 114: just another note, should we enforce SwiftLint rules instead of having them as warnings? I notice at times we merge them in

cyndichin (approved):
thanks for resolving this bug! as discussed with the team in terms of how we review PRs, I didn't pull down, but let me know if you would like me to test

PocketKit/Tests/SyncTests/Support/Space+factories.swift
bassrock: @cyndichin yes we should.

PocketKit/Sources/PocketKit/Article/ReadableViewController.swift
bassrock: im gonna leave that change out for now given it uses different constants,

PR #417, opened 2023-02-22
feat(snowplow): adding in baseline snowplow analytics for ui testing

bassrock:
## Summary

Snowplow Micro is a java version of snowplow that acts as a collector that will validate all the events we send to it based on our Pocket snowplow event schema.  This creates a set of helper classes and sets up the necessary framework for any developer or QA person to start validating the generated analytics within our UI tests.

There are lots of room for optimization and more helper methods, but I would like to implement all that as we go and start to use this. 

## References 
* https://github.com/snowplow-incubator/snowplow-micro

## Implementation Decisions
* This will require snowplow micro running locally to execute UI tests moving forward. I believe that to be an acceptable compromise to ensure our analytics are validated.

## Test Steps
* This mainly adds testing methods for developers and QA

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

## Screenshots

Some logging of micro running in the background in ci:
Screenshot 2023-02-22 at 5 36 31 PM

### DIFF
--- a/AllTests.xctestplan
+++ b/AllTests.xctestplan
@@ -9,12 +9,6 @@
     }
   ],
   "defaultOptions" : {
-    "environmentVariableEntries" : [
-      {
-        "key" : "SNOWPLOW_ENDPOINT",
-        "value" : "http:\/\/localhost:9090"
-      }
-    ],
     "testTimeoutsEnabled" : true
   },
   "testTargets" : [
--- a/AllTests.xctestplan
+++ b/AllTests.xctestplan
@@ -9,6 +9,12 @@
     }
   ],
   "defaultOptions" : {
+    "environmentVariableEntries" : [
+      {
+        "key" : "SNOWPLOW_ENDPOINT",
+        "value" : "http:\/\/localhost:9090"
+      }
+    ],
     "testTimeoutsEnabled" : true
   },
   "testTargets" : [
--- a/Tests iOS/HomeTests.swift
+++ b/Tests iOS/HomeTests.swift
@@ -87,26 +87,44 @@ class HomeTests: XCTestCase {
 
         home.sectionHeader("Slate 2").verify()
         home.recommendationCell("Slate 2, Recommendation 1").verify()
-        
+
         await snowplowMicro.assertBaselineSnowplowExpectation()
 
-        let data = await snowplowMicro.getFirstEvent(with: "home.recent.impression")
+        async let slate1Rec1 = snowplowMicro.getFirstEvent(with: "discover.impression", recommendationId: "slate-1-rec-1")
+        async let slate1Rec2 = snowplowMicro.getFirstEvent(with: "discover.impression", recommendationId: "slate-1-rec-2")
+        async let slate2Rec1 = snowplowMicro.getFirstEvent(with: "discover.impression", recommendationId: "slate-2-rec-1")
+        async let slate2Rec2 = snowplowMicro.getFirstEvent(with: "discover.impression", recommendationId: "slate-1-rec-2")
+
+        let recs = await [slate1Rec1, slate1Rec2, slate2Rec1, slate2Rec2]
+        let loadedSlate1Rec1 = recs[0]!
+        let loadedSlate1Rec2 = recs[1]!
+        let loadedSlate2Rec1 = recs[2]!
+        let loadedSlate2Rec2 = recs[3]!
+
+        snowplowMicro.assertRecommendationImpressionHasNecessaryContexts(event: loadedSlate1Rec1, url: "http://localhost:8080/hello")
+        snowplowMicro.assertRecommendationImpressionHasNecessaryContexts(event: loadedSlate1Rec2, url: "https://example.com/item-2")
+        snowplowMicro.assertRecommendationImpressionHasNecessaryContexts(event: loadedSlate1Rec1, url: "https://example.com/item-1")
+        snowplowMicro.assertRecommendationImpressionHasNecessaryContexts(event: loadedSlate1Rec1, url: "https://example.com/item-2")
     }
 
-    func test_navigatingToHomeTab_showsRecentlySavedItems() {
+    @MainActor
+    func test_navigatingToHomeTab_showsRecentlySavedItems() async {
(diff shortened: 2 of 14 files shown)


AllTests.xctestplan
```diff
@@ -9,6 +9,12 @@
     }
   ],
   "defaultOptions" : {
+    "environmentVariableEntries" : [
+      {
+        "key" : "SNOWPLOW_ENDPOINT",
+        "value" : "http:\/\/localhost:9090"
```
cyndichin, line 15: curious, why not `http://127.0.0.1:9090` as we have in our Environment Variables?

cyndichin: also why do we not need to pass in the identifier anymore?

cyndichin (commented):
Looking great! So excited to have this! can't wait for the demo and be able to test this out via search analytics 🔥

AllTests.xctestplan
bassrock: localhost is more the defacto standard when accessing the local system.

Tests iOS/HomeTests.swift
```diff
@@ -80,22 +87,44 @@ class HomeTests: XCTestCase {
 
         home.sectionHeader("Slate 2").verify()
         home.recommendationCell("Slate 2, Recommendation 1").verify()
+
+        await snowplowMicro.assertBaselineSnowplowExpectation()
```
bassrock, line 91: This is how the new tests are used.

Tests iOS/HomeTests.swift
```diff
@@ -80,22 +87,44 @@ class HomeTests: XCTestCase {
...
-    func test_navigatingToHomeTab_showsRecentlySavedItems() {
+    @MainActor
+    func test_navigatingToHomeTab_showsRecentlySavedItems() async {
         let home = app.launch().homeView.wait()
         home.savedItemCell("Item 1").wait()
         home.savedItemCell("Item 2").wait()
         home.savedItemCell("Item 1").swipeLeft(velocity: .fast)
         home.savedItemCell("Item 3").swipeLeft(velocity: .fast)
         waitForDisappearance(of: home.savedItemCell("Item 3"))
+        await snowplowMicro.assertBaselineSnowplowExpectation()
```
bassrock, line 118: If there a way to add this at the end of every UI test we should. I could not find it.

Tests iOS/Support/AnyCodable.swift
```diff
@@ -0,0 +1,90 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import Foundation
+
+/**
+ Class to help transform Snowplow events without setting up the whole object.
```
bassrock, line 8: see note.

Tests iOS/Support/SnowplowMicro.swift
```diff
@@ -0,0 +1,303 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+import Foundation
```
bassrock, line 4: This class is the snowplow micro magic.

bitrise.yml
```diff
@@ -138,6 +138,25 @@ workflows:
...
+        - script@1:
+           inputs:
+             - content: |
+                #!/usr/bin/env bash
+
+                # fail if any commands fails
+                set -e
+                
+                ## Usually Pocket runs snowplow via our docker image, but Bitrise does not support docker in Mac environments so we use their built jar.
+                ## It uses the values in MICRO_IGLU_REGISTRY_URL and MICRO_IGLU_API_KEY
```
bassrock, line 153: Note this. I think @brahmini also uses the raw jar in her test suite, but calling this out cause this is different then some other teams, but a side of effect of Mac CI 😬

Tests iOS/HomeTests.swift
```diff
@@ -80,22 +87,44 @@ class HomeTests: XCTestCase {
...
     }
 
-    func test_tappingRecentSavesItem_showsReader() {
+    @MainActor
+    func test_tappingRecentSavesItem_showsReader() async {
```
bassrock, line 122: We need to make the tests `async` and `MainActor` to utilize the snowplow micro code.

Tests iOS/Support/SnowplowMicro.swift
```diff
@@ -0,0 +1,322 @@
...
+    func has(identifier: String) -> Bool {
+        return (self.dataDict()["identifier"] as? String) == identifier
+    }
+
+    func has(url: String) -> Bool {
+        return (self.dataDict()["url"] as? String) == url
+    }
+
+    func assertHas(url: String) {
+        return XCTAssertEqual((self.dataDict()["url"] as? String), url)
```
bassrock, line 131: Doing the assert here, so that when it fails we get the inequality printed instead of using `has` and doing an isTrue Assertion

Tests iOS/HomeTests.swift
cyndichin: There is tearDown that occurs at the end of every test but typically used for cleanup.

Tests iOS/HomeTests.swift
cyndichin: thanks! I was just wondering about this

cyndichin (approved):
looks good high level, will be able to do more digging when using this to tests in the future. thanks!

Tests iOS/HomeTests.swift
bassrock: I looked into that, but I couldn't tell if it would fail as expected.

cyndichin: I added `XCTAssertTrue(false)` in the teardown and it fails the test, so I assume so based on that the snowplow one will work the same.

image

bitrise.yml
brahmini: Catching up late here, but, yes we use java jar on the web :)

PR #418, opened 2023-02-22
fix(analytics): few analytics tests that snowplow micro caught

bassrock:
## Summary

Working on integrating Snowplow Micro, and I found some errors in the new analytics. This corrects those.

## Implementation Details

Commented out persistent contexts until @cyndichin 's pr which will fix them


cyndichin (approved):
short and sweet!

PR #421, opened 2023-02-22
Premium Subscriptions: purchase subscription

Gio2018:
## Summary
*This PR adds the capability of purchasing a premium subscription from the App Store

## Implementation Details
* Update `User` to publish `status` changes
* Add transaction listener to `PremiumSubscriptionStore`
* Add support for subscription purchase to `PremiumSubscriptionStore`
* Inject `PemiumSubscriptionStore` from `Services`, so that we can listen to transaction updates at app launch
* Add purchase functionality to `PremiumUpgradeView` and `PremiumUpgradeViewModel`
* Add `Test_Subscriptions.storekit` for `StoreKit` purchase testing.

## Test Steps
WIP

## PR Checklist:
- [ ] ~~Added Unit / UI tests~~ a `StoreKitTest` PR will be open soon
- [x] Basic Self QA

## Screenshots

<p align=center>

</p>


PocketKit/Sources/PocketKit/PocketSceneDelegate.swift
```diff
@@ -45,7 +45,9 @@ public class PocketSceneDelegate: UIResponder, UIWindowSceneDelegate {
                         user: Services.shared.user,
                         userDefaults: Services.shared.userDefaults,
                         notificationCenter: .default
-                    )
+                    ) {
+                        PremiumUpgradeViewModel(store: Services.shared.subscriptionStore)
```
bassrock, line 49: 😍

PocketKit/Sources/PocketKit/Premium/Model/PremiumSubscriptionStore.swift
```diff
@@ -5,15 +5,117 @@
...
+            do {
+                let verifiedTransaction = try verify(transaction)
+                switch verifiedTransaction.productType {
+                case .autoRenewable:
+                    if let subscription = subscriptions.first(where: { $0.product.id == verifiedTransaction.productID }) {
+                        purchasedSubscription = subscription
+                        user.setPremiumStatus(true)
+                    }
+                default:
+                    // We do not have other product types as of now. In the future, who knows..
```
bassrock, line 112: Lets send a Log.capture(message: ) here. That will send something to sentry. If we hit this, something is wrong, and we should know.

PocketKit/Sources/PocketKit/Premium/Model/PremiumSubscriptionStore.swift
```diff
@@ -5,15 +5,117 @@
...
+                    break
+                }
+            } catch {
+                // TODO: use logger here
```
bassrock, line 116: TY!

bassrock (approved):
Straight forward!

PocketKit/Sources/PocketKit/Premium/Model/PremiumSubscriptionStore.swift
```diff
@@ -5,15 +5,117 @@
...
+        Task {
+            do {
+                try await requestSubscriptions()
+            } catch {
+                // TODO: use logger here
+                print(error)
+            }
+        }
     }
-    @MainActor
```
cyndichin, line 15: curious, why remove `@MainActor`

PocketKit/Sources/PocketKit/Settings/AccountViewModel.swift
```diff
@@ -3,13 +3,15 @@ import Analytics
 import Textile
 import SharedPocketKit
 import SwiftUI
+import Combine
```
cyndichin, line 6: what needs combine here?

Config/Test_Subscriptions.storekit
```diff
@@ -0,0 +1,77 @@
+{
```
cyndichin, line 1: was this auto generated?

PocketKit/Sources/SharedPocketKit/User.swift
```diff
@@ -8,26 +8,44 @@ public enum Status: String {
 }
 
 public protocol User {
-    var status: Status? { get }
+    var status: Status { get }
+    var statusPublisher: Published<Status>.Publisher { get }
+    var publishedStatus: Published<Status> { get }
```
cyndichin, line 13: what is the different between `statusPublisher` and `publishedStatus` and how are we expecting to use them?

cyndichin (commented):
looks great so far! love the use of async / await, I have a couple of questions for my own knowledge, no blockers. Assuming testing will come at a later date.

PocketKit/Sources/PocketKit/Premium/Model/PremiumSubscriptionStore.swift
```diff
@@ -5,15 +5,117 @@
...
+                    // TODO: use logger here
+                    print(error)
+                }
+            }
+        }
+    }
+
+    /// Varify the passed transaction
+    /// - Parameter transaction: a new transaction to verify
+    /// - Returns: a verified trnasaction, if verification is successful. Throws an error otherwise.
```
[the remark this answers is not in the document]
bassrock: I think we should! This format seems fine to me.

PocketKit/Sources/PocketKit/Premium/Model/PremiumSubscriptionStore.swift
Gio2018: done!

PocketKit/Sources/PocketKit/Premium/Model/PremiumSubscriptionStore.swift
Gio2018: Initially this call was directly made at `PremiumUpgradeView` instantiation. Now, we instantiate the store at app launch and pre-fetch products. This method is also called by `PremiumUpgradeViewModel` when the corresponding view is instantiated, but being that view model marked as `@Mainactor` will ensure that UI updates will still happen on the main thread.

PocketKit/Sources/PocketKit/Premium/Model/PremiumSubscriptionStore.swift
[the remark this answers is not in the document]
Gio2018: Thanks! This is also easy to obtain: in Xcode, when pointing anywhere in a method definition, if you hit `cmd+option+/', a pre-populated documentation in the above format will be generated, and you just need to replace the placeholders

PocketKit/Sources/PocketKit/Settings/AccountViewModel.swift
Gio2018: Nothing, it must be a leftover from something that I tried and decided not to do in the end 😄 . Thanks for the catch!

Config/Test_Subscriptions.storekit
Gio2018: The StoreKit configuration file is created following the instructions in the StoreKit testing document, shared on Slack; so it's not auto-generated.

PocketKit/Sources/SharedPocketKit/User.swift
Gio2018: This relates to what we were discussing in the Premium Subscription architecture meeting: if we want to pass `User` (which is a protocol) to our view models etc, and, at the same time observe changes in `status`, then, since we cannot declare a property in a protocol with a property wrapper (this makes sense, since protocols are generic types, while property wrappers are concrete), then we need a publisher and a published to link, respectively, the projected and the wrapped value of that property, if we declare it as `@Published` in a concrete implementation of `User`. You can read more on the internet, for example [here](https://danielbernal.co/2020/06/26/combine-and-protocols-in-swift/) or [here](https://medium.com/swlh/how-to-define-a-protocol-with-published-property-wrapper-type-1b6349097064). That said, I'm not a big fan of this implementation and if we find a better way to achieve this result, I'd be up for it!

cyndichin: Can we `sink` on `statusPublisher`, do we also need to expose `publishedStatus`? I'll probably have a better understanding when I see this in used. thanks for the explanation though!

PR #429, opened 2023-02-24
feat(search): matching web default sort order

bassrock:
## Summary

When validating the Search work I noticed that my results were not matching Web no matter what I did. It turns out that after talking to @collectedmind, Web utilizes a sort order which changes your default result set. 

This brings us in line with them.

Long term we should align with @collectedmind and a search product partner to align all clients.


cyndichin (approved):
looks good! didn't realize we had to sort. A test to sort by createdAt would be good here, but not going to make it a blocker.

bassrock:
Tests discussed here https://pocket.slack.com/archives/C01UH57EJKD/p1677255000616299

PR #437, opened 2023-02-25
Premium Subscriptions: Update UI after user purchases a subscription

Gio2018:
## Summary
* This PR adds logic to update the app UI when the user successfully purchases a premium subscription

## Implementation Details
* Inject `PremiumViewModel` factory into `SarchViewModel`, so we can instantiate `PremiumUpgradeView` when needed
* `SearchViewModel` now listens to user status changes, to update the UI accordingly
* `EmptyStateView` now accepts an alternate external view that can be used in place of the standard button, so that we are not limited to a button that calls a web view anymore
* `GetPocketPremiumButton` is a button that presents `PocketPremiumView`, used in `EmptyStateView`, online search, free users

## Test Steps
Pre-requisite:
- configure the `StoreKit` testing environment using the steps described in #421 on this branch
Testing steps (see animation below):
- Once the prerequisites are implemented, build and run this branch
- Login with a free account
- Go to "Saves" then select "Search", and navigate to "All Items"
- Make sure you see the premium upsell, with the "Get Pocket Premium" button
- Make sure the button color is `Coral2`
- Tap the "Get Pocket Premium" button and make sure the premium upgrade screen appears
- Purchase a subscriptions and, after you see the "You're all set" alert, tap "OK"
- Make sure the premium upgrade screen gets dismissed, and you don't see the premium upsell in the empty view
- Go to the "Settings" tab, and make sure that the "Go Premium" row above the "Sign Out" does not appear

> **Note**
At this moment, the premium status is not synced with the backend. If you relaunch, the user status will be reverted to free.

## PR Checklist:
- [ ] ~~Added Unit / UI tests~~
- [x] Self Review (review, clean up, documentation, run tests)
- [x] Basic Self QA

## Animation
<p align=center>

</p>

PR #438, opened 2023-02-25
Premium Subscriptions: Premium Subscription Row

Gio2018:
## Summary
*This PR changes the premium row in "Settings" after a user has successfully purchased a premium subscription. The description will change from Go Premium to Premium Subscription, and the tint color from `black1` to `teal2`

## Implementation Details
* Update `SettingsRowButton` to allow color configuration of title and icons
* Update `AccountViewModel` to listen to `User.status` changes as they happen
* Update `Localizable.strings`, add `settings.premiumSubscriptionRow` string

## Test Steps
Pre-requisite:
- checkout this branch and setup `StoreKit` testing as described in #421 
Testing steps (see animation below):
- build/run
- login with a free account
- go to the "Settings" tab
-  tap "Go Premium"
- Make sure the premium upgrade screen appears
- Purchase a subscription
- Tap "OK" in the "You're all set" alert
- Make sure the premium upgrade screen dismisses
- Make sure the premium row (above "Sign Out") now shows "Premium Subscription" and its color is teal

> **Note**
Same limitations as #421 and #437 apply: if you relaunch, the status will be reverted to free.
Also, at this moment, tapping the "Premium Subscription" does nothing. The related screen will be added in a future PR.

## PR Checklist:
- [ ] ~~Added Unit / UI tests~~
- [x] Basic Self QA

## Animation
<p align=center>

</p>

### DIFF
--- a/PocketKit/Sources/PocketKit/Settings/AccountViewModel.swift
+++ b/PocketKit/Sources/PocketKit/Settings/AccountViewModel.swift
@@ -1,8 +1,9 @@
-import Sync
 import Analytics
-import Textile
+import Combine
 import SharedPocketKit
 import SwiftUI
+import Sync
+import Textile
 
 class AccountViewModel: ObservableObject {
     static let ToggleAppBadgeKey = "AccountViewModel.ToggleAppBadge"
@@ -21,9 +22,9 @@ class AccountViewModel: ObservableObject {
     @AppStorage("Settings.ToggleAppBadge")
     public var appBadgeToggle: Bool = false
 
-    public var isPremium: Bool {
-        user.status == .premium
-    }
+    private var userStatusListener: AnyCancellable?
+
+    @Published var isPremium: Bool
 
     init(appSession: AppSession,
          user: User,
@@ -35,6 +36,14 @@ class AccountViewModel: ObservableObject {
         self.userDefaults = userDefaults
         self.notificationCenter = notificationCenter
         self.premiumUpgradeViewModelFactory = premiumUpgradeViewModelFactory
+        self.isPremium = user.status == .premium
+
+        userStatusListener = user
+            .statusPublisher
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] status in
+                self?.isPremium = status == .premium
+            }
     }
 
     func signOut() {
--- a/PocketKit/Sources/PocketKit/Settings/Components/SettingsRowButton.swift
+++ b/PocketKit/Sources/PocketKit/Settings/Components/SettingsRowButton.swift
@@ -7,7 +7,8 @@ struct SettingsRowButton: View {
     var icon: SFIconModel?
     var leadingImageAsset: ImageAsset?
     var trailingImageAsset: ImageAsset?
-    var tintColor: Color?
+    var leadingTintColor: Color = Color(.ui.black1)
+    var trailingTintColor: Color = Color(.ui.black1)
 
     let action: () -> Void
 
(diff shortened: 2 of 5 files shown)


PocketKit/Sources/PocketKit/Settings/AccountViewModel.swift
```diff
@@ -35,6 +36,14 @@ class AccountViewModel: ObservableObject {
         self.userDefaults = userDefaults
         self.notificationCenter = notificationCenter
         self.premiumUpgradeViewModelFactory = premiumUpgradeViewModelFactory
+        self.isPremium = user.status == .premium
+
+        userStatusListener = user
```
cyndichin, line 41: curious to see why this method vs. `.store()`

PocketKit/Sources/PocketKit/Settings/Components/SettingsRowButton.swift
```diff
@@ -7,7 +7,8 @@ struct SettingsRowButton: View {
     var icon: SFIconModel?
     var leadingImageAsset: ImageAsset?
     var trailingImageAsset: ImageAsset?
-    var tintColor: Color?
+    var leadingTintColor: Color = Color(.ui.black1)
+    var trailingTintColor: Color = Color(.ui.black1)
```
cyndichin, line 11: nit: would these ever be inconsistent? if not, maybe we can have one `tintColor = .ui.black` and so one place to change it

PocketKit/Sources/PocketKit/Settings/AccountViewModel.swift
Gio2018: Hey @cyndichin , I'm not sure I follow here: can you please clarify?

Gio2018:
Thanks for reviewing @cyndichin !

>  ran into a couple of issues when purchasing monthly and then switching to annual.

switching between subscriptions is not currently supported. the description above mentions #421 for testing subscriptions, but should probably have had some more description itself

> the modal doesn't not always auto dismiss
This needs attention, although I agree that's not related to the premium row PR: did this happen between launches, and did you clear the transaction history in the `StoreKit` configuration file, as described in the confluence document?
Thanks!!

PocketKit/Sources/PocketKit/Settings/AccountViewModel.swift
cyndichin: I just noticed in our code base we use the following to store our cancellable objects, so was curious to see why this method was preferred over using `.store`

image

PocketKit/Sources/PocketKit/Settings/Components/SettingsRowButton.swift
cyndichin: oh got it, thanks for the clarification!

cyndichin:
> The video capture was after I relaunched my simulator and logged out and signed back in as a free user. However, I don't think it should be a blocker and we can test end to end once the features for premium subscription are implemented.

PocketKit/Sources/PocketKit/Settings/AccountViewModel.swift
Gio2018: Oh I see now, thanks for clarifying! Basically, `,store(in: )`: the reason I did not use that is because, being this a single subscription, we don't really need a `Set` for it, but I can totally change it if you feel that makes the code more clear or more consistent.

Gio2018:
> The video capture was after I relaunched my simulator and logged out and signed back in as a free user. However, I don't think it should be a blocker and we can test end to end once the features for premium subscription are implemented.

Cool thank you!

PocketKit/Sources/PocketKit/Settings/AccountViewModel.swift
cyndichin: either works, was curious about the difference!

PR #442, opened 2023-02-27
fix(localization): adding in some missed strings

bassrock:
## Summary

I noticed a few strings were missing from localization. So adding them in.


Gio2018 (approved):
:shipit:

PR #454, opened 2023-03-02
feat(tags): filter tags list based on input

cyndichin:
## Summary
Add ability to filter tags list based on user input in text field.

## References 
IN-1136

## Implementation Details
* Add `otherTags` property to be used to hold our all tags and filter tags list
* Added `newTagInput` to the view models, so we can apply the `debounce` method 
* Created `filterTags` function to show filtered tags or all tags if none exist

## Test Steps
- [ ] WHEN I am on the tags creation screen (Add Tags Screen)
AND I have selected the text input field
AND I input text 
THEN the list of existing tags should filter to only show tags that include the inputted text

1. All Tags changes to Tags if we’re filtering (similar to legacy)
2. If there are no filtered tags for the input, then we show All Tags
3. The timer to trigger whether to filter is half a second after the user stops entering

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

## Screenshots
https://user-images.githubusercontent.com/6743397/223155713-99bb2131-ad97-4bc0-bbf3-a372e3ab854b.mp4

### DIFF
--- a/PocketKit/Sources/PocketKit/Tags/PocketAddTagsViewModel.swift
+++ b/PocketKit/Sources/PocketKit/Tags/PocketAddTagsViewModel.swift
@@ -1,14 +1,25 @@
 import Combine
 import Sync
 import Textile
+import Foundation
+import SharedPocketKit
 
 class PocketAddTagsViewModel: AddTagsViewModel {
     private let item: SavedItem
     private let source: Source
     private let saveAction: () -> Void
+    private var userInputListener: AnyCancellable?
+
+    var sectionTitle: TagSectionType = .allTags
 
     @Published
     var tags: [String] = []
+
+    @Published
+    var newTagInput: String = ""
+
+    @Published
+    var otherTags: [String] = []
 
     init(item: SavedItem, source: Source, saveAction: @escaping () -> Void) {
         self.item = item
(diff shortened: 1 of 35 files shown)


PocketKit/Sources/PocketKit/Tags/PocketAddTagsViewModel.swift
```diff
@@ -1,30 +1,66 @@
...
+            })
     }
 
     func addTags() {
         source.addTags(item: item, tags: tags)
         saveAction()
     }
 
-    func allOtherTags() -> [String]? {
+    func allOtherTags() {
```
bassrock, line 44: Can we doc block this?

bassrock (approved):
@cyndichin looks good to me! One little comment, but no big deal.

Also what about analytics for this?

PocketKit/Sources/PocketKit/Tags/PocketAddTagsViewModel.swift
cyndichin: will do! I'll also add it to the preexisting functions too

Tests iOS/AddTagsItemTests.swift
```diff
@@ -132,6 +163,41 @@ class AddTagsItemTests: XCTestCase {
...
+        let addTagsView = app.addTagsView.wait()
+        addTagsView.wait()
+        addTagsView.newTagTextField.tap()
+        addTagsView.newTagTextField.typeText("F")
+
+        addTagsView.allTagsRow(matching: "filter tag 0").wait()
+        addTagsView.allTagsRow(matching: "filter tag 1").wait()
+        app.addTagsView.allTagsView.wait()
+
+//        Bitrise is failing, but this passes locally, commenting out for now
```
cyndichin, line 196: not sure why bitrise is failing for this analytics, but I am able to run this successfully locally

bassrock (approved):

PR #459, opened 2023-03-06
Add manual purchase sync to SubscriptionStore

Gio2018:
## Summary
* This PR adds a direct call to purchase sync to `SubscriptionStore`, for those rare cases where the automatic sync does not happen

## Implementation Details
* Add a `restoreSubscription` call to `SubscriptionStore`, which corresponds to an `Appstore.sync()` call in the concrete implementation `PocketSubscriptionStore`

## Test Steps
* There's not much to test at the moment; here is the [documentation](https://developer.apple.com/documentation/storekit/appstore/3791906-sync) related to that call which, according to Apple, will only be needed in rare cases. When the call happens, the listener in place should receive existing transactions and entitlements.

## PR Checklist:
- [ ] Added Unit / UI tests
- [ ] Basic Self QA


cyndichin (approved):
looks good, approving based on code. But I think Bitrise is failing because we want to also add this to the `MockSubscriptionStore` class

PR #461, opened 2023-03-07
fix(coredata): removing temporary core data code

bassrock:
## Summary
Removes the temporary core data deletion code when we made pieces in core data not null.

## References 
* IN-1173


bassrock:
@nzeltzer in our next beta release notes, we should state that we made some behind the scenes changes, and if the app is crashing and not showing your full list to delete and reinstall.

Gio2018 (approved):

cyndichin (commented):
do we want to remove the launch argument as well in Tests iOS?

PR #463, opened 2023-03-07
ci(tests): Misc code tweaks to decrease flaky fails for several tests

audie1:
This PR attempts to fix a few flaky failing tests, summaries below:

test_addTagsToItemFromSaves_savesNewTags() - Removed the index check and now checking for the tag name on the Saves screen

test_navigatingToHomeTab_withClipboardURL_showsBannerAndSavedItem - Moved the check for the disappearance of the banner a few lines down, to allow more time for the UI to resolve


cyndichin (approved):
looks good! thanks for helping us with the flakey tests. seems there are some SwiftLint errors, could we apply `swiftlint --fix` before merging this PR in?

PR #464, opened 2023-03-07
feat(archive): improve handling of archive

bassrock:
## Summary

Improve the archive experience by allowing Archives to stay on the user's device.

In a follow-up we will update the store to clear out any old article (mArticle) data on archived content. IE Archived items will only have metadata stored locally.

## Implementation Details
* This merges the Saves and Archives view model because Archive is just a filter on saves.
    * This will allow us to implement features faster across Saves & Archives by not duplicating logic.
* Set the code to download 400 of the user's archives in a similar manner to how we download their saves. 

## Test Steps
* Open the app and ensure that Archives load from CoreData and do not download mArticle. This can be observed by clicking into an archived item and the reader view takes a few seconds to load from the server.

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA


PocketKit/Sources/PocketKit/MyList/ItemsList/SavedItemsList/SavedItemsListViewModel.swift
```diff
@@ -4,14 +4,26 @@ import Analytics
 import Combine
 import UIKit
 
+public enum SavesViewType {
+    case saves
+    case archive
```
cyndichin, line 9: just an FYI - we do have other existing `enums` not sure if we should repurpose them or create a new one.

PocketKit/Sources/PocketKit/Refresh/RefreshCoordinator.swift
```diff
@@ -54,7 +55,7 @@ class RefreshCoordinator {
             task.setTaskCompleted(success: false)
         }
 
-        source.refresh {
+        source.refreshSaves {
```
cyndichin, line 58: do we need a `refreshArchive`

PocketKit/Tests/PocketKitTests/SavedItemsListViewModelTests.swift
```diff
@@ -355,14 +358,14 @@ class SavedItemsListViewModelTests: XCTestCase {
...
-        source.initialDownloadState.send(.paginating(totalCount: 2))
+        source.initialSavesDownloadState.send(.paginating(totalCount: 2))
         itemsController.delegate?.controllerDidChangeContent(itemsController)
 
         wait(for: [receivedSnapshot], timeout: 1)
     }
 
     func test_receivedSnapshots_whenInitialDownloadIsStarted_insertsPlaceholderCells() throws {
-        source.initialDownloadState.send(.started)
+        source.initialSavesDownloadState.send(.started)
```
cyndichin, line 368: should we test `archive` version as well?

PocketKit/Sources/PocketKit/MyList/ItemsList/SavedItemsList/SavedItemsListViewModel.swift
```diff
@@ -349,7 +405,7 @@ class SavedItemsListViewModel: NSObject, ItemsListViewModel {
 
         let itemCellIDs: [ItemsListCell<ItemIdentifier>]
 
-        switch source.initialDownloadState.value {
+        switch source.initialSavesDownloadState.value {
```
cyndichin, line 408: do we also need to handle `initialArchiveDownloadState`?

cyndichin (approved):
ran the app and looked at the code on high level. Love how we're consolidating 🔥, thanks for doing this! some comments, but none blocker

PocketKit/Sources/PocketKit/Refresh/RefreshCoordinator.swift
bassrock: We have one, but I dont want to always be refreshing it. I think Archive should follow an first load + on demand refresh model.

cc @nzeltzer

PR #465, opened 2023-03-08
feat(pull-to-refresh): adding in custom pull to refresh indicator

bassrock:
## Summary
Uses our new loading animation for pull to refresh.

> **Note**
There appears to be a flash of the loading view in simulator before you fully pull to refresh. As far as I can tell this may be a fun iOS bug. It seems the spinner icon does the same thing you just don't notice it as much.

## Implementation Details
* Implements a custom UIRefreshControl that uses a LottieView.

## Test Steps
* Pull to refresh, see indicator!

## PR Checklist:
- [ ] Added Unit / UI tests
- [x] Basic Self QA

## Screenshots

https://user-images.githubusercontent.com/1010384/223614795-0a3ecbc6-29a1-4f60-846b-fcacb35e5d38.mp4


PocketKit/Sources/Textile/Views/LottieRefreshControl.swift
```diff
@@ -0,0 +1,69 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+// Modified from https://stackoverflow.com/questions/53123171/custom-uirefreshcontrol-with-lottie
+
+import UIKit
+import Lottie
+
+public class LottieRefreshControl: UIRefreshControl {
+    fileprivate let animationView: LottieAnimationView
+    fileprivate var isAnimating = false
```
cyndichin, line 12: curious could we make these `private` instead?

cyndichin (commented):
Nice! Only blocker is there is a bit weird on alignment for iPad. Do we want to address this before merging?
https://user-images.githubusercontent.com/6743397/223759881-264af687-127d-4443-ac66-791a617e8e07.mp4

PocketKit/Sources/Textile/Views/LottieRefreshControl.swift
bassrock: @cyndichin copy pasted from the original link in the header, i could if we want.

cyndichin: ah okay thanks np! just prefer to be on the side of most restricted if possible

is this ready for review again?

PocketKit/Sources/PocketKit/MyList/ItemsList/ItemsListViewController.swift
```diff
@@ -3,6 +3,7 @@ import Sync
 import CoreData
 import Combine
 import Kingfisher
+import Textile
```
cyndichin, line 6: do we still need this?

PocketKit/Sources/Textile/Animations/PocketAnimation.swift
```diff
@@ -0,0 +1,21 @@
+import Foundation
+
+import Lottie
+
+public enum PocketAnimation {
```
cyndichin, line 5: I like this refactor!

cyndichin (approved):
looks good! thanks for the work~

PR #474, opened 2023-03-09
fix(search): refactor search pagination

cyndichin:
## Summary
PR to fix issues where share sheet was dismissing due to triggering pagination for search results when not needed.

## References 
IN-1187

## Implementation Details
* Modified `SearchService` to only call once unless the user scrolls down the list. After the initial search call and the user scrolls down the list, I will check if the `pageInfo` from GraphQL response has `hasNextPage == true` and also create a variable `lastEndCursor` that gets hold by the service until the user submits another search or cancels search. 
* Created function `loadMoreResults` to trigger the next page call after a certain index has been reached in the list. This is an arbitrary constant that fits for both iPad and iPhone.

## Test Steps
Verify that all search functionality exists and works properly

**Confirm Bug Fix**
1. Navigate to Search and submit a search for a specific term
2. Tap on share icon for an item in the results view
3. Confirm that Share sheet does not auto dismiss and no network called for search query4. 

**Confirm Pagination Works Fix**
1. Navigate to Search and submit a search for a specific term
2. Scroll down the list and see how the list populates the results in the scroll bar3. 
3. Confirm analytics for `searchPage` is called.

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

## Screenshots

https://user-images.githubusercontent.com/6743397/224354359-685aa1cd-d53c-41e2-947b-a317d2caca36.mov

### DIFF
--- a/PocketKit/Sources/Analytics/AppEvents/Search.swift
+++ b/PocketKit/Sources/Analytics/AppEvents/Search.swift
@@ -306,27 +306,20 @@ public extension Events.Search {
     /**
      Fired when a user scrolls down the search list and triggers another results page call
      */
-    static func nextSearchResultsPage(
-        url: URL,
-        positionInList: Int,
+    static func searchResultsPage(
+        numberOfItems: Int,
         scope: SearchScope
     ) -> Event {
         return Engagement(
             .general,
             uiEntity: UiEntity(
-                .button,
-                identifier: "global-nav.search.nextResultsPage",
+                .page,
+                identifier: "global-nav.search.searchPage",
                 componentDetail: getScopeIdentifier(scope: scope),
-                index: positionInList
-            ),
-            extraEntities: [
-                ContentEntity(
-                    url: url
-                )
-            ]
+                index: numberOfItems
+            )
         )
     }
-
 
     /// Premium upsell in Search visible
     static func premiumUpsellViewed() -> Event {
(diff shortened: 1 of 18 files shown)


PocketKit/Sources/Analytics/AppEvents/Search.swift
```diff
@@ -303,6 +303,24 @@ public extension Events.Search {
         )
     }
 
+    /**
+     Fired when a user scrolls down the search list and triggers another results page call
+     */
+    static func searchResultsPage(
```
cyndichin, line 309: still working with Analytics to determine what the details should be

cyndichin: updated analytics in the following commit
`fix(analytics): update search page analytics`

Tests iOS/MyList/SearchTests.swift
```diff
@@ -594,6 +594,54 @@ class SearchTests: XCTestCase {
...
+        app.saves.searchView.searchItemCell(matching: "Item 27").wait().element.swipeUp(velocity: .fast)
+
+        wait(for: [firstExpectRequest, secondExpectRequest])
+
+        await snowplowMicro.assertBaselineSnowplowExpectation()
+
+        async let event0 = snowplowMicro.getFirstEvent(with: "global-nav.search.searchPage", index: 1)
+        async let event1 = snowplowMicro.getFirstEvent(with: "global-nav.search.searchPage", index: 2)
+
+        let events = await [event0, event1]
```
bassrock, line 636: 🥳  asyncing some awaits! yay!

bassrock (approved):
yay!

PR #475, opened 2023-03-09
fix(ui): cleaning up the account deletion ui

bassrock:
## Summary
* Doing some super small UI cleanup after review with design.


Gio2018 (commented):
Looks like some tests are failing here, is it expected?

bassrock:
@Gio2018 no, i think bitrise is flakey cause I didn't change any identifiers, just icons etc

Gio2018 (approved):

PR #477, opened 2023-03-10
fix(data): remove static data cleaner

bassrock:
## Summary

In local environments Home and all saves data would be deleted upon second launch of the app if logged in. Turns out this is because local environments are build <= 70.

This code was introduced in https://github.com/Pocket/pocket-ios/commit/f1e428bb0a2cd4d0e1ee73041cf851d78bbfd98d and is no longer needed.


PocketKit/Sources/PocketKit/Refresh/HomeRefreshCoordinator.swift
```diff
@@ -36,18 +36,30 @@ class HomeRefreshCoordinator: HomeRefreshCoordinatorProtocol {
         }
 
         if shouldRefresh(isForced: isForced), !isRefreshing {
-            Task {
+            Task { [weak self] in
+                guard let self = self else {
```
Gio2018, line 40: you could make this more compact, omitting `= self`  if you like: `guard let self else { ...`

Gio2018 (approved):

PR #487, opened 2023-03-11
fix(threads): moving all coredata to the background

bassrock:
## Summary
Move all CoreData and Apollo network processing to a background queue to prevent the Main thread from hanging. 

## References 
* IN-1204

## Implementation Details
There are multiple ways we can introduce CoreData concurrency into the application. One approach could be to make each operation occur in it's own thread and NSManagedContext with the views utilizing the viewContext (main thread context). Another is to do the same thing, but instead of each operation getting it's own thread and NSManagedContext we utilize a single background thread and context.

I chose to do the later. 

Space will now take in 2 contexts, 1 for background processing, and 1 for all view models/logic. I then provided proxy functions on Space to access the two NSManagedContexts. Moving forward all Data processing and FetchedController setup should come from a function within Space so that CoreData processing and rendering is contained to a single area of the codebase.

When we pass NSManagedObjects around, we should always be aware of the thread the code will be executing on and based on the thread we are on.

If you are doing view logic you should do:

```swift
guard let savedItem = space.viewObject(with: savedItem.objectID) as? SavedItem else {
    Log.capture(message: "Could not get SavedItem from viewContext while fetching details")
    return
}
```

If you are about to do write/update work you should do it like:

```swift
try space.performAndWait {
    guard let savedItem = space.backgroundObject(with: savedItem.objectID) as? SavedItem else {
        Log.capture(message: "Could not get SavedItem from backgroundContext while fetching details")
        return
    }
    savedItem.update(from: remoteSavedItem.fragments.savedItemParts, with: space)
    try space.save()
}
```

This will perform the data processing on the Thread that is specific to the NSManagedObjectContext. 

I also had to add `@MainActor` to all our view models and coordinators to ensure they operate on main threads.

>**Note**
All mock building functions in our test suite will build the objects in the backgroundContext. For them to be accessible to the viewContext ensure you call `space.save()` before using them. 

## Test Steps

Use app as normal.  Ensure no segfaults or crashes. This changes all underlying code.

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

>**Note**
There is more refactoring that needs to be done and comments to be added so that the app architecture is clearer for new developers and those actively adding to the project. That is out of scope for this work, and I will follow up with the changes.

### DIFF
--- a/PocketKit/Sources/PocketKit/AppBadgeTracker.swift
+++ b/PocketKit/Sources/PocketKit/AppBadgeTracker.swift
@@ -44,16 +44,13 @@ class AppBadgeSetup {
     }
 
     func manualCheckForSavedCount() {
-        let numberOfSavesRequest = Requests.fetchSavedItems()
-        var numberOfSaves: Int
+        var numberOfSaves: Int = 0
         let currentValue = userDefaults.bool(forKey: AccountViewModel.ToggleAppBadgeKey)
-        if currentValue == false {
-            numberOfSaves = 0
-        } else {
+        if currentValue != false {
             do {
-                numberOfSaves = try source.mainContext.fetch(numberOfSavesRequest).count
+                numberOfSaves = try source.unreadSaves()
             } catch {
-                numberOfSaves = 0
+                Log.capture(error: error)
             }
         }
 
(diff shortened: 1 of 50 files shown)


timc-mozilla (approved):
based on what Daniel and I talked about during our call, this lgtm

PR #491, opened 2023-03-13
ci(tests): fix for item actions flaky test

audie1:
This is a small fix for test_webview_includesCustomItemActions() in SavesTests.swift

This test has been flaky failing, and I put in a small guard conditional check to make sure that the share button is tapped correctly after the view loads. I hope this fixes the flaky fails, if not I'll have to try something else.


audie1:
The test failed in the CI, so I need to try another solution. I can't convert this to a draft for some reason. No review is required yet 🙌

bassrock:
@audie1 I think this is fixed in https://github.com/Pocket/pocket-ios/pull/500

PR #501, opened 2023-03-20
feat(premium): update receipt encoding

Gio2018:
## Summary
* This PR removes the url encoding from the App Store receipt. It also optimizes the code that sends the receipt out, only sending a "restore" receipt type to the backend if no purchased product was found (in which case a "purchase" type will be sent)

## Implementation Details
* Update `AppStoreReceiptService`, remove url encoding from the receipt
* Update `PocketSubscriptionService`, send the restore request to the backend only if no active product was found.

## Test Steps
* There's not much manual testing here.

## PR Checklist:
- [ ] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/PocketKit/PremiumUpgrade/Model/PocketSubscriptionStore.swift
+++ b/PocketKit/Sources/PocketKit/PremiumUpgrade/Model/PocketSubscriptionStore.swift
@@ -135,7 +135,7 @@ extension PocketSubscriptionStore {
         }
         do {
             // in case no subscription was found,
-            // still send the receipt App Store receipt to the backend
+            // still send the App Store receipt to the backend
             try await receiptService.send(nil)
         } catch {
             Log.capture(error: error)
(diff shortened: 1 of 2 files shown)


PocketKit/Sources/PocketKit/PremiumUpgrade/Model/PocketSubscriptionStore.swift
```diff
@@ -39,13 +39,6 @@ final class PocketSubscriptionStore: SubscriptionStore, ObservableObject {
             }
             // Restore a purchased subscription, if any
             await self.fetchActiveSubscription()
-
```
bassrock, line 42: are we still sending receipt on launch?

bassrock (approved):
just had one question, but not blocking

PocketKit/Sources/PocketKit/PremiumUpgrade/Model/PocketSubscriptionStore.swift
Gio2018: Yep, this method is called at `init()` time which happens on launch when all the other `Services` are initialized

Gio2018: What it does:
1. it will try to fetch a subscription to the App Store; if it does, it will send a receipt with the subscription data
2. if it does not, it will send a receipt with no subscription data, of type "restore", so if one exists that was not retrieved, the backend would.

timc-mozilla (approved):

PR #506, opened 2023-03-22
chore(tests): fixing logout tests

bassrock:
## Summary

Logout test was getting flakey lately on a lot of prs. 

## Implementation Details

I switched a method from using `.exists` to `waitForExistence`. This gives the app some time to render the view. It behaves the same as `exists` under the hood, the only difference is it adds a wait option to give the property a chance to turn true.

## Test Steps
* Run logout tests

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA


Tests iOS/SignOutTests.swift
```diff
@@ -58,10 +58,12 @@ class SignOutTests: XCTestCase {
         let cellCount = account.cells.count
         XCTAssertTrue(cellCount > 0)
 
-        let logoutRowTappedEvent = await snowplowMicro.getFirstEvent(with: "global-nav.settings.logout")
+        let events = await [snowplowMicro.getFirstEvent(with: "global-nav.settings.logout"), snowplowMicro.getFirstEvent(with: "global-nav.settings.logout-confirmed")]
```
cyndichin, line 61: is this for organizational purposes and not necessarily performance?

cyndichin (approved):
looks good, just one curious question!

Tests iOS/SignOutTests.swift
bassrock: @cyndichin performance! `getFirstEvent` makes an http call. By putting them in an array and awaiting the array it will perform the calls on seperate threads. Otherwise it has to work syncronously.

PR #508, opened 2023-03-22
fix(orientation): allow only reader to rotate

cyndichin:
## Summary
Fix rotation on iPhone and only allow Reader to be rotated.

## References 
IN-1134

## Implementation Details
Created static var `phoneOrientationLock` that is set to `portrait` by default for phone devices. Only case we want to be able to rotate the view is for our reader and added `lockOrientation` to be able to make the change for this view when it appears. 

## Test Steps
As a phone user, when I rotate my device to landscape orientation, the screen should not rotate, unless I am reading an article. 

1. Navigate to Premium Purchase Screen or any other screen that is not Reader
2. Attempt to rotate the screen and confirm view does not change

1. Navigate to Reader
2. Attempt to rotate the screen and confirm view does indeed change

## PR Checklist:
- [N/A] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/PocketKit/PocketAppDelegate.swift
+++ b/PocketKit/Sources/PocketKit/PocketAppDelegate.swift
@@ -10,6 +10,8 @@ import BackgroundTasks
 import SharedPocketKit
 
 public class PocketAppDelegate: UIResponder, UIApplicationDelegate {
+    static var phoneOrientationLock = UIInterfaceOrientationMask.portrait
+
     private let source: Source
     private let userDefaults: UserDefaults
     private let firstLaunchDefaults: UserDefaults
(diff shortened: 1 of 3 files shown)


CMasterson (approved):

PocketKit/Sources/PocketKit/PocketAppDelegate.swift
```diff
@@ -79,4 +81,14 @@ public class PocketAppDelegate: UIResponder, UIApplicationDelegate {
 
         return true
     }
+
+    /// Sets orientations to use for the views
+    /// - Parameters:
+    ///   - application: singleton app object
+    ///   - window: window whose interface orientations you want to retrieve
+    /// - Returns: orientations to use for the view
+    public func application(_ application: UIApplication, supportedInterfaceOrientationsFor window: UIWindow?) -> UIInterfaceOrientationMask {
```
dskuza, line 90: @cyndichin does this get called every time there's a view controller change (i.e appear, disappear)?

cyndichin: @dskuza yes it does but it also provides the simplest way to set all views to portrait from what I can find, do you have an alternative way that might be preferred?

dskuza: @cyndichin not that I know of in a SwiftUI world; the "old" setup was based on UIKit delegate methods

dskuza (approved):

PocketKit/Sources/PocketKit/PocketAppDelegate.swift
dskuza: That, and all of the usage of  `UITabBarControllerDelegate` and `UINavigationControllerDelegate`.

PR #511, opened 2023-03-23
fix(slate-detail): removing extra slate detail refresh

bassrock:
## Summary

Ensure that recommendations do not reload when a user clicks into a slate detail.

## References 
* IN-1230

## Implementation Details

Per conversation with @nzeltzer we are fetching more recs on initial load of home to make it so slate details does not reorder the results.

## Test Steps
* Load home
* Go into a slate detail, you should not see them re-order.

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA


cyndichin (approved):
nice, looks great on iPhone! slate detail does not reload anymore.

PR #513, opened 2023-03-24
fix(cache): improve caching and deletion of images

dskuza:
## Summary

Improve how images are (un)cached when there are Image changes in Source.

## References 

IN-1222

## Implementation Details

The usage of `ImagesController` is now simplified - we will handle Image changes when all changes have been saved, rather than on each insert / delete / etc. There is now some additional logic for when images should be downloaded:

1. Checked for any orphaned Images (Images that have no `.item`)
2. Generate a set of all URLs for the latest Images
3. Generate a set of all URLs for orphaned Images
4. Create sets of URLs that should be downloaded, skipped, and deleted
5. Delete the appropriate cached Images, download the appropriate uncached Images, and delete all orphan Images from source

## Test Steps

- [ ] Log in to Pocket
- [ ] Let Home / Saves load, and let thumbnails load for some (or all) items
  - A spinner should appear while a thumbnail is downloading
- [ ] Close and reopen the application
- [ ] Let Home / Saves appear, and thumbnails should be immediately loaded if previously visible; no spinner should appear

## PR Checklist:

- [x] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/PocketKit/ImageManager.swift
+++ b/PocketKit/Sources/PocketKit/ImageManager.swift
@@ -11,6 +11,8 @@ protocol ImageCacheProtocol {
                      callbackQueue: CallbackQueue,
                      completionHandler: (() -> Void)?
     )
+
+    func isCached(forKey key: String, processorIdentifier identifier: String) -> Bool
 }
 
 extension ImageCache: ImageCacheProtocol { }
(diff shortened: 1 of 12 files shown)


dskuza:
The overall changelog of this pull request, while may look "large", is rather small; there are a bunch of renames, and then updated tests (which definitely affect the total changes).

PocketKit/Sources/PocketKit/ImageManager.swift
```diff
@@ -53,77 +55,91 @@ class ImageManager {
...
+        let orphanURLs = Set(orphans.compactMap { $0.source })
+
+        // All URLs that are not orphans
+        let toDownload = allURLs.subtracting(orphanURLs)
+        // Skip deleting any URLs that are also orphans, as to not redownload
+        let skipDelete = toDownload.intersection(orphanURLs)
+        // Delete all orphan URLs that are not to be skipped
+        let toDelete = orphanURLs.subtracting(skipDelete)
+
+        toDelete.forEach { delete(url: $0) }
```
bassrock, line 132: @dskuza are these async methods? if not we should make them async

PocketKit/Sources/PocketKit/ImageManager.swift
```diff
@@ -53,77 +55,91 @@ class ImageManager {
...
-                if success {
-                    self?.source.download(images: [image])
-                }
-            }
-        case .delete:
-            delete(image: image)
-        case .move:
-            return
-        @unknown default:
+    func handle(images: [Image]?) {
```
timc-mozilla, line 112: NIT: can we name it other than "handle"?

timc-mozilla (approved):

bassrock (approved):

PocketKit/Sources/PocketKit/ImageManager.swift
dskuza: @bassrock not really… they "rely" on kingfisher code (behind the scenes and some protocols) which does things async, so there's not really anything synchronous here. the KF functions themselves aren't async, they're just performed async (using dispatch queues, i believe)

PocketKit/Sources/PocketKit/ImageManager.swift
timc-mozilla: handleImageCache?

dskuza: @timc-mozilla ¯\_(ツ)_/¯ but the thought of renaming this is surprisingly overwhelming so i'll leave it for now 😆

PR #515, opened 2023-03-26
fix(concurrency): clean up some of our threading and calls

bassrock:
## Summary

The UI was constantly hanging for me when logging out and in during testing. I believe this came down to how we were using threads, saving data and making network calls.

In this PR i fixed a few things:
* Ensure that we do not bombard our poor server, by:
    * Allowing a refresh of data only after 5 seconds from the last refresh has passed
    * Ensure that if we are restoring a failed FetchSaves/Archive operation we do not start another one.
* Ensure Home monitors relationship changes in [NSFetchedResults Controller](https://www.avanderlee.com/swift/nsfetchedresultscontroller-observe-relationship-changes/)

* Ensure that all images we cache are saved to device, and the respective models listen for the image being downloaded.
* Speed up initial sync by giving FetchSaves & FetchArchive their own queues so they can work in parallel instead of sequentially
* Speed up initial sync by performing a fetch of tags and saves at the same time.
* Ensure Home is sorted as expected by our server
    * This fixes a few glitches where mutliple hero cards could appear. Turns out the sectionNameKeyPath of NSFetchedResultsController needs to match the first sortDescriptor to work right.
 
Things I want to re-add but pulled when I could not make the test suite work:
* Ensure we save on background thread, so the next network request does not have to wait: https://github.com/Pocket/pocket-ios/pull/515/commits/a73b909fb09c99ef63da73ba718fa59f91cbcce7
    * When this was in place I saw massive initial sync improvements, because we spend more time writing to CoreData then we do making network requests. 

Things we need to follow up with:
*  If the initial FetchSaves operation never completed, and is restored, it should continue from where it left off, not from the beginning

## Implementation Details
* Followed some patterns from https://www.avanderlee.com/swift/nsfetchedresultscontroller-observe-relationship-changes/ to observe changes to relationships in NSFetchedResultsController.

## Test Steps
* Log out and in multiple times, and while doing so make sure that the UI does not hang (or is better then before) and that there are no weird anamolities.

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/Sync/CoreData+Relationships/RelationshipKeyPath.swift
+++ b/PocketKit/Sources/Sync/CoreData+Relationships/RelationshipKeyPath.swift
@@ -1,0 +1,59 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import Foundation
+import CoreData
+
+/// Describes a relationship key path for a Core Data entity.
+public struct RelationshipKeyPath: Hashable {
+    /// The source property name of the relationship entity we're observing.
+    let sourcePropertyName: String
+
+    let destinationEntityName: String
+
+    /// The destination property name we're observing
+    let destinationPropertyName: String
+
+    /// The inverse property name of this relationship. Can be used to get the affected object IDs.
+    let inverseRelationshipKeyPath: String
+
+    public init(keyPath: String, relationships: [String: NSRelationshipDescription]) {
+        let splittedKeyPath = keyPath.split(separator: ".")
+
+        if splittedKeyPath.count == 2 {
+            // Original code from https://www.avanderlee.com/swift/nsfetchedresultscontroller-observe-relationship-changes/ to monitor 1 level deep of relationships
+            sourcePropertyName = String(splittedKeyPath.first!)
+            destinationPropertyName = String(splittedKeyPath.last!)
+            let relationship = relationships[sourcePropertyName]!
+            destinationEntityName = relationship.destinationEntity!.name!
+            inverseRelationshipKeyPath = relationship.inverseRelationship!.name
+        } else if splittedKeyPath.count == 3 {
+            // Modified code to monitor 2 level deep of relationships.
+            // There is probaly a better way to do this with recursion and supporting more layers, but right now we only need 2, and, well, this works. 😅
+
+            let firstPropertyName = String(splittedKeyPath.first!)
+            sourcePropertyName = String(splittedKeyPath[1])
+
+            let relationship = relationships[firstPropertyName]!
+            let firstInverseKeyPath = relationship.inverseRelationship!.name
+
+            guard let secondRelationships = relationship.inverseRelationship?.entity.relationshipsByName else {
+                fatalError("Developer error, you should have a relationship if you are monitoring this path")
+            }
+
+            let secondRelationship = secondRelationships[sourcePropertyName]!
+
+            destinationPropertyName = String(splittedKeyPath.last!)
+            destinationEntityName = secondRelationship.destinationEntity!.name!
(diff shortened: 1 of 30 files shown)


PocketKit/Sources/Sync/CoreData+Relationships/RelationshipKeyPath.swift
```diff
@@ -0,0 +1,59 @@
...
+
+    let destinationEntityName: String
+
+    /// The destination property name we're observing
+    let destinationPropertyName: String
+
+    /// The inverse property name of this relationship. Can be used to get the affected object IDs.
+    let inverseRelationshipKeyPath: String
+
+    public init(keyPath: String, relationships: [String: NSRelationshipDescription]) {
```
Gio2018, line 21: this method feels a bit hacky with string parsing and explicit array indexes; on the other hand, because we are using `#keypath` this should ensure that we don't accidentally have mistyped strings, so.. 😄

Gio2018 (approved):
Works for me @bassrock ! Just a few optional comments!

PocketKit/Sources/Sync/CoreData+Relationships/RelationshipKeyPath.swift
bassrock: Yea i added comments in the file. Its not meant to be pretty. There are probably better ways to do this. i am all for cleanup here.

PocketKit/Sources/Sync/Operations/FetchArchive.swift
```diff
@@ -27,6 +27,15 @@ class FetchArchive: SyncOperation {
 
     func execute() async -> SyncOperationResult {
         do {
+            if lastRefresh.lastRefreshArchive != nil {
```
CMasterson, line 30: Minor: Is this 'if' necessary? We guard lastRefreshArchive on the next line.

PocketKit/Sources/Sync/Operations/FetchSaves.swift
```diff
@@ -30,9 +30,18 @@ class FetchSaves: SyncOperation {
 
     func execute() async -> SyncOperationResult {
         do {
-            try await fetchSaves()
-            try await fetchTags()
+            if lastRefresh.lastRefreshSaves != nil {
```
CMasterson, line 33: Minor: Similar to above, the if seems unnecessary.

PocketKit/Sources/Sync/PocketSource.swift
```diff
@@ -35,9 +35,28 @@ public class PocketSource: Source {
...
+        q.name = "com.mozilla.pocket.fetch.saves"
+        return q
+    }()
+
+    private let fetchArchiveQueue: OperationQueue = {
+        let q = OperationQueue()
+        q.maxConcurrentOperationCount = 1
+        q.qualityOfService = .background
+        q.name = "com.mozilla.pocket.fetch.archive"
         return q
```
CMasterson, line 60: Medium: These could be consolidated into one function that takes the name as a parameter.

CMasterson (commented):
Half way there. Livin' on a prayer.

PocketKit/Sources/Sync/Operations/FetchSaves.swift
```diff
@@ -30,9 +30,18 @@ class FetchSaves: SyncOperation {
...
+                    // However many states may not come from a user, IE. Instant Sync, Persistent Tasks that never finished, Retries
+                    return .success
+                }
+            }
 
+            async let saves: Void = fetchSaves()
+            async let tags: Void = fetchTags()
+            _ = await [try saves, try tags]
             lastRefresh.refreshedSaves()
             return .success
```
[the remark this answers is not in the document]
CMasterson: Feel free to ignore. It's a lot to remove one if statement.

bassrock:
thanks @CMasterson we can followup with more improvements

PR #522, opened 2023-03-27
feat(reader): add unarchive button

cyndichin:
## Summary
* Include short description of what this PR accomplishes

## References 
IN-929

## Implementation Details
* Added a new button for unarchiving an item or moving it back to saves.
* Modify the reader for both when accessing an item from home and from saves

## Test Steps
- [ ] WHEN the user opens an item in the READER
AND the opened item is archived
THEN there should be an unarchive button in the top navigation bar 

- [ ] WHEN the user taps the unarchive button 
THEN the item should be added to the top of the user’s saves
AND the item should be removed from the list of user’s archived articles
AND the unarchive button should become the archive button

## PR Checklist:
- [ ] Added Unit / UI tests
- [ ] Basic Self QA


timc-mozilla (approved):

PocketKit/Sources/PocketKit/Article/Presenters/UnsupportedComponentPresenter.swift
```diff
@@ -27,8 +27,4 @@ class UnsupportedComponentPresenter: ArticleComponentPresenter {
     private func handleShowInWebReaderButtonTap() {
         readableViewModel?.showWebReader()
     }
-
-    private func handleArchiveArticleButtonTap() {
-        readableViewModel?.archiveArticle()
-    }
```
cyndichin, line 33: wasn't being used - so removed

PocketKit/Sources/PocketKit/Article/ReadableViewModel/RecommendationViewModel.swift
```diff
@@ -80,12 +80,8 @@ class RecommendationViewModel: ReadableViewModel {
...
     }
 
-    func moveToSaves() {
-        guard let savedItem = recommendation.item?.savedItem else {
-            return
-        }
-
-        source.unarchive(item: savedItem)
+    var isArchived: Bool {
+        return recommendation.item?.savedItem?.isArchived ?? false
```
CMasterson, line 84: This line reaches very deeply into recommendation.
Can we add a function to recommendation that returns if it's saved item is archived instead?
Otherwise we are tightly coupling the ViewModel to it.

PocketKit/Sources/PocketKit/Article/ReadableViewModel/ReadableEvent.swift
```diff
@@ -2,6 +2,7 @@ import Sync
 
 enum ReadableEvent {
     case archive
+    case moveToSaves
```
CMasterson, line 5: Both 'moveToSaves' and 'unarchive' are used interchangeably in the PR. Consider sticking to one (personally I'd pick unarchive as it has little chance of being confused for saving a new article)

CMasterson:
The icon used is just the Pocket icon, that is intended?

PocketKit/Sources/PocketKit/Home/HomeViewController.swift
```diff
@@ -403,7 +403,7 @@ extension HomeViewController {
             switch event {
             case .contentUpdated:
                 break
-            case .archive, .delete:
```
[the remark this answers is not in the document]
cyndichin: good point, I sent the video of the feature, but let me confirm with this specific part.

PocketKit/Sources/PocketKit/Article/ReadableViewModel/ReadableEvent.swift
cyndichin: haha agreed! even had a battle with myself when to use which one. I'll keep existing as is, but new changes to use unarchive.

PocketKit/Sources/PocketKit/Article/ReadableViewModel/RecommendationViewModel.swift
cyndichin: discussed over slack, as of now, will keep this change as we're currently doing this in multiple places. If we want to refactor this, we can create a separate ticket to rethink our architecture in general when reaching for the `isArchived` property for our view models in general

PocketKit/Sources/PocketKit/Article/ReadableViewModel/RecommendationViewModel.swift
```diff
@@ -228,13 +212,23 @@ extension RecommendationViewModel {
         track(identifier: .itemUnfavorite)
     }
 
-    private func archive() {
+    func moveToSaves() {
+        guard let savedItem = recommendation.item?.savedItem else {
+            return
```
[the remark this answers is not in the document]
cyndichin: will update with error! thanks!

cyndichin:
> The unarchive icon used is just the Pocket icon, that is intended?

yes, as we are using it for the `moveToSaves`
also confirming feature via https://pocket.slack.com/archives/C01UH57EJKD/p1680015188909139?thread_ts=1679943867.910309&cid=C01UH57EJKD

PocketKit/Sources/PocketKit/Item/ReadableHostViewController.swift
```diff
@@ -99,7 +91,40 @@ class ReadableHostViewController: UIViewController {
...
+        readableViewModel.archive()
+    }
+
+    @objc
+    private func unarchiveArticle() {
+        readableViewModel.unarchive { [weak self] success in
+            if success, let archiveButton = self?.getArchiveButton() {
+                self?.navigationItem.rightBarButtonItems?[2] = archiveButton
+            }
+        }
```
cyndichin, line 103: Should I add a check for the `rightBarButtonItems.count` > 3, but we know this already when initializing it?

PocketKit/Sources/PocketKit/Home/HomeViewController.swift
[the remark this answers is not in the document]
cyndichin: thanks again! updated!

PocketKit/Sources/PocketKit/Item/ReadableHostViewController.swift
CMasterson: Boundary checks would be good here. It works now but is brittle and may break on change.
Maybe an Enum with each buttons index, to avoid crossover?

PocketKit/Sources/PocketKit/Item/ReadableHostViewController.swift
```diff
@@ -99,7 +91,40 @@ class ReadableHostViewController: UIViewController {
...
+    }
+
+    private func getArchiveButton() -> UIBarButtonItem {
```
dskuza, line 106: there are multiple options for doing something like this -- another is `lazy var`. something we could talk about as a team to determine if there's a pattern we should follow.

PocketKit/Sources/PocketKit/Item/ReadableHostViewController.swift
```diff
@@ -99,7 +91,40 @@ class ReadableHostViewController: UIViewController {
...
-        readableViewModel.archiveArticle()
+        readableViewModel.archive()
+    }
+
+    @objc
+    private func unarchiveArticle() {
+        readableViewModel.unarchive { [weak self] success in
+            if success, let archiveButton = self?.getArchiveButton(), let count =  self?.navigationItem.rightBarButtonItems?.count, count >= 3 {
+                self?.navigationItem.rightBarButtonItems?[2] = archiveButton
+            }
```
dskuza, line 102: i'd like to see this done a little more safely. if, for whatever reason, index 2 doesn't exist, this will cause a crash. if we do something like, for example, maintain a reference to the unarchive button, we can find it in the right bar button items array (if it exists) and replace it safely.

dskuza (changes requested):

PocketKit/Sources/PocketKit/Article/ReadableViewModel/ReadableViewModel.swift
```diff
@@ -26,14 +26,15 @@ protocol ReadableViewModel: ReadableViewControllerDelegate {
     var domain: String? { get }
     var publishDate: Date? { get }
     var url: URL? { get }
+    var isArchived: Bool { get }
 
     func delete()
     func openExternally(url: URL?)
-    func archiveArticle()
+    func archive()
+    func unarchive(completion: (Bool) -> Void)
```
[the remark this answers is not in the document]
cyndichin: hmm seems like @CMasterson preferred `unarchive`, should this be a larger team discussion? I can go either way.

PocketKit/Sources/PocketKit/Item/ReadableHostViewController.swift
cyndichin: If index 2 doesn't exist, then it shouldn't be able to hit line 101 right? Maintaining a reference will utilize the `lazy var` you suggested right?

PocketKit/Sources/PocketKit/Item/ReadableHostViewController.swift
cyndichin: Ref: https://pocket.slack.com/archives/C01UH57EJKD/p1680188131450979

PocketKit/Sources/PocketKit/Article/ReadableViewModel/ReadableViewModel.swift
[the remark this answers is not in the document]
cyndichin: Ref: https://pocket.slack.com/archives/C01UH57EJKD/p1680188418032409

PocketKit/Sources/PocketKit/Item/ReadableHostViewController.swift
cyndichin: updated! does this reflect what you were thinking?

PocketKit/Sources/PocketKit/Item/ReadableHostViewController.swift
```diff
@@ -98,8 +114,17 @@ class ReadableHostViewController: UIViewController {
...
-        readableViewModel.archiveArticle()
+    private func archive() {
+        readableViewModel.archive()
+    }
+
+    @objc
+    private func moveFromArchiveToSaves() {
+        readableViewModel.moveFromArchiveToSaves { [weak self] success in
+            if success, let items = navigationItem.rightBarButtonItems, let index = items.firstIndex(of: getMoveFromArchiveToSavesButton) {
+                self?.navigationItem.rightBarButtonItems?[index] = getArchiveButton
```
dskuza, line 125: :+1:

dskuza (approved):

PR #528, opened 2023-03-28
fix(home): ensuring home does not always refresh

bassrock:
## Summary

It was reported that Home was constantly refreshing for @mkoidin. Upon 👁️ I noticed that the Network code was calling the updateHandler multiple times and we were always having home force refresh every time.

Instead forceRefresh should only be reserved for a user pulling to refresh. 

We should also always observe network status on a background queue.

## Test Steps
* Change network status and ensure that home does not do a hard refresh (where items change) when toggling network status.

## PR Checklist:
- [ ] Added Unit / UI tests
- [ ] Basic Self QA

### DIFF
--- a/PocketKit/Sources/PocketKit/Home/HomeViewModel.swift
+++ b/PocketKit/Sources/PocketKit/Home/HomeViewModel.swift
@@ -132,9 +132,7 @@ class HomeViewModel: NSObject {
 
         networkPathMonitor.updateHandler = { [weak self] path in
             if path.status == .satisfied {
-                DispatchQueue.main.async {
-                    self?.refresh(isForced: true) { }
-                }
+                self?.refresh(isForced: false) { }
             }
         }
         fetch()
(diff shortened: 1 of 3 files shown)


timc-mozilla (approved):

PocketKit/Sources/PocketKit/Home/HomeViewModel.swift
```diff
@@ -146,11 +144,13 @@ class HomeViewModel: NSObject {
...
-            try self.recomendationsController.performFetch()
-        } catch {
-            Log.capture(error: error)
+        DispatchQueue.main.async {
+            do {
+                try self.recentSavesController.performFetch()
+                try self.recomendationsController.performFetch()
+            } catch {
+                Log.capture(error: error)
+            }
```
dskuza, line 153: Does this affect the thread on which the fetched results controller works? This seems like the opposite of what our changes to what queues we use is.

bassrock: its the oddest stuff. im double checking

PR #531, opened 2023-03-29
fix(storekit1): add storekit 1 methods

bassrock:
## Summary

Pocket's backend only knows about StoreKit 1 for the time being. It seems the receipt that gets saved in the receipt URL area is a storeKit 2 receipt and is not compatible with our backend.

So we need to explcitly create a StoreKit 1 receipt request and send that to our backend.

## Implementation Details
* Checked continuations are so cool.

## Test Steps
* Test out purchasing and it should succeed.

## PR Checklist:
- [ ] Added Unit / UI tests
- [ ] Basic Self QA

### DIFF
--- a/PocketKit/Sources/PocketKit/Home/HomeViewController.swift
+++ b/PocketKit/Sources/PocketKit/Home/HomeViewController.swift
@@ -168,7 +168,7 @@ class HomeViewController: UIViewController {
             task.setTaskCompleted(success: true)
         }
     }
-    
+
     override func viewWillAppear(_ animated: Bool) {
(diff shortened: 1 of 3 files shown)


cyndichin (approved):
code looks good in terms of adding the storeKit 1 method, curious to understand the context of why we are adding them. Do we have to handle cases where receipts were created with StoreKit 1?

Gio2018 (approved):

bassrock:
@cyndichin Our server only understands sStoreKit 1, so im explicitly loading a store kit 1 receipt. I dont think StoreKit 2 is working as expected.

PR #535, opened 2023-03-29
fix(receipts): ensuring there is a receipt

bassrock:
## Summary

On iOS simulators when you are logged out of Apple, a system prompt will show if you request a receipt. Instead we should see if a recept exists first

## PR Checklist:
- [ ] Added Unit / UI tests
- [ ] Basic Self QA

### DIFF
--- a/PocketKit/Sources/PocketKit/PremiumUpgrade/Model/ReceiptService.swift
+++ b/PocketKit/Sources/PocketKit/PremiumUpgrade/Model/ReceiptService.swift
@@ -27,6 +27,9 @@ class AppStoreReceiptService: NSObject, ReceiptService {
     }
 
     func send(_ product: Product?) async throws {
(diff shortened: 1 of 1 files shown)


PocketKit/Sources/PocketKit/PremiumUpgrade/Model/ReceiptService.swift
```diff
@@ -27,6 +27,9 @@ class AppStoreReceiptService: NSObject, ReceiptService {
     }
 
     func send(_ product: Product?) async throws {
+        // First make sure a receipt even exists before we try and get one.
+        _ = try getReceipt()
```
Gio2018, line 31: wondering if, instead of calling `getReceipt()` twice, we can just handle the whole thing inside `getReceipt()` where we already check the existence on disk?

Gio2018 (approved):

PR #537, opened 2023-03-29
fix(flakey): run flakey tests nightly

bassrock:
## Summary

* Run flakey tests nightly
* move to xcode 14.3
* remove building QA Build on every pr, instead a dev can trigger it manually in Bitrise as needed.

## Test fixes

* Search tests - changed the mocking to use expectations instead of checkedContinuations. 
* Tests using threads
   *  Set the mock classes, when they write to `calls` to perform that write in a queue using sync, to prevent race conditions that could occur.


timc-mozilla (approved):

PocketKit/Sources/Sync/PocketSource.swift
```diff
@@ -229,12 +229,10 @@ public class PocketSource: Source {
     // Exposed to tests to facilitate waiting for all operations to finish
     // Should not be used outside of a testing context
     func drain(_ completion: @escaping () -> Void) {
-        DispatchQueue.global(qos: .background).async {
```
bassrock, line 232: only called on tests

.swiftlint.yml
```diff
@@ -65,7 +65,9 @@ opt_in_rules: # some rules are only opt-in
   # swiftlint rules
   - anyobject_protocol
   - array_init
-  - attributes
+  - attributes:
+      always_on_line_above:
+        - '@Published'
```
bassrock, line 70: @cyndichin Lastet version of [swiftlint](https://github.com/realm/SwiftLint/releases/tag/0.51.0) causes 100+ violations without this change, do we need to update to match the rest of Mozilla?

cyndichin: let me follow up with Roux on this because I don't see `attributes` in their `.swiftlint.yml` file, but it's marked as Done in the Mozilla SwiftLint spreadsheet. I was going to create a PR with the updated rules next Monday.

I'm also not seeing the violations. Do you have a screenshot of what it looks like?

bassrock: @cyndichin update to the latest swiftlint `brew update && brew upgrade swiftlint`

cyndichin: ah okay, I see it. Roux says it has not been incorporated in their main branch yet but they did not have to add the additional field. Most likely due to us using a lot of @Published property wrappers. I'll make sure to check this when updating the rules - thanks!

PR #541, opened 2023-03-30
Fix a memory leak when opening the Reader

Gio2018:
## Summary
* This PR fixes a leak happening when opening an item in the reader

For reference, here's what was happening before

<p align=center>

</p>

## Implementation Details
* Update `ReadableViewModel`, add an `unowned` reference to `self` in the closure that defines the `layout` property to avoid the cycle

## Test Steps
* If you're curious about testing, you can open an item in the reader a few times and check the memory graph making sure there's no more than one instance of `ReadableViewModel` (zero if you're not in the reader anymore)
* If you wanna see the behavior in the picture, just remove unowned and repeat the test.

## PR Checklist:
- [ ] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/PocketKit/Article/ReadableViewController.swift
+++ b/PocketKit/Sources/PocketKit/Article/ReadableViewController.swift
@@ -42,7 +42,7 @@ class ReadableViewController: UIViewController {
         collectionViewLayout: layout
     )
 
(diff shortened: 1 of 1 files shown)


timc-mozilla (approved):

bassrock (approved):

nzeltzer (commented):
lgtm (amaze)

PR #544, opened 2023-03-31
refactor(tags): create reusable components and functionality for Add / Edit Tags view

cyndichin:
## Summary
Refactor tags list to use tag cell components and set up for Recent Tags to be consistent across both screens

## References 
IN-1268

## Implementation Details
* Separate Add Tags View to three components: `InputTagsView`, `TagsListView`, `TagsCell`
* Utilize `TagsCell` component in the `Edit Tags screen`. This required that the tag types be the same, so created `TagType` that will serve the purpose of stating if the tag is a recent tag, normal tag or not tagged tag. 
* Refactor how we implement the list in terms of arranging tags and created a protocol `TagsList` to be shared across Edit Tags / Add Tags

## Test Steps
Go through the Edit Tags and Add Tags screen to verify that functionality has not changed

## PR Checklist:
- N / A Added Unit / UI tests
- [x] Basic Self QA


PocketKit/Sources/Textile/Views/Tags/TagType.swift
```diff
@@ -0,0 +1,17 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+public enum TagType: Hashable {
+    case notTagged
+    case tag(String)
```
cyndichin, line 7: will also add a case for `recentTag` in a future PR

PocketKit/Sources/Textile/Views/Tags/TagsList.swift
```diff
@@ -0,0 +1,23 @@
...
+        let fetchedTags = tags.reversed()
+        if fetchedTags.count > 3 {
+            let topRecentTags = Array(fetchedTags)[..<3]
+            let sortedTags = Array(fetchedTags)[3...].sorted()
+            allTags.append(contentsOf: topRecentTags)
+            allTags.append(contentsOf: sortedTags)
+        } else {
+            allTags.append(contentsOf: fetchedTags)
+        }
+        return allTags.compactMap { TagType.tag($0) }
```
cyndichin, line 21: logic will be revisited in future recent tags PR

PocketKit/Sources/Textile/Views/Tags/TagsList.swift
```diff
@@ -0,0 +1,26 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+public protocol TagsList {
+    func arrangeTags(with tags: [String]) -> [TagType]
+}
```
CMasterson, line 7: Why is this a protocol?

PocketKit/Sources/Textile/Views/Tags/TagsList.swift
```diff
@@ -0,0 +1,26 @@
...
+
+public extension TagsList {
+    /// Arranges the tags in a list by sorting with three recent tags and then alphabetizing the rest
+    /// - Parameter tags: list of users tags
+    /// - Returns: a list of tags to display to the user
+    func arrangeTags(with tags: [String]) -> [TagType] {
```
CMasterson, line 13: Is it possible that we try to arrange tag strings that can't be rendered into TagTypes

PocketKit/Sources/Textile/Views/Tags/TagsList.swift
```diff
@@ -0,0 +1,26 @@
...
+        var allTags: [String] = []
+        let fetchedTags = tags.reversed()
+        if fetchedTags.count > 3 {
+            let topRecentTags = Array(fetchedTags)[..<3]
```
CMasterson, line 17: Can we for sure say we will always be getting 3 tags as recent tags?

If you imagine a new user who just added one tag to an article, how would they get 3?

PocketKit/Sources/Textile/Views/Tags/TagsList.swift
CMasterson: The more I think about it, the more I believe that this is backend work.
We can't experiment with tag orders or the like with this code.

PocketKit/Sources/Textile/Views/Tags/TagsList.swift
cyndichin: So that it can be used for both the view models for add tags (i.e. `PocketAddTagsViewModel`)  and edit/filter tags (i.e. `TagsFilterViewModel`)

PocketKit/Sources/Textile/Views/Tags/TagsList.swift
cyndichin: In that case, if tags is less than three, then we won't show the tags as recent tags. They will be displayed in alphabetical order.

PocketKit/Sources/Textile/Views/Tags/TagsList.swift
cyndichin: > Is it possible that we try to arrange tag strings that can't be rendered into TagTypes
Tag Types can take in a String, so not sure what care it would be not possible.

> The more I think about it, the more I believe that this is backend work.
Interesting point, I agree that it will be nice to have it backend. Post MVP - recent tags will be a backend call, but as for now we are using recent tags created locally. 

> We can't experiment with tag orders or the like with this code.
Can you expand on this? thanks!

PocketKit/Sources/Textile/Views/Tags/TagsList.swift
CMasterson: Protocols (imo) are for declaring abstract interfaces so outsiders don't need to know the concrete type they are interacting with.

If (as it seems) no one outside is trying to call `arrangeTags` then a protocol seems like an odd choice.

If what you wanted was for PocketAddTagsVM and TagsFilterVM to be able to arrangeTags, I would have gone with a simple helper function, potentially declared in the same file (but not the same Struct as TagType.

PocketKit/Sources/Textile/Views/Tags/TagsList.swift
CMasterson: Alright, fair.

CMasterson (approved):

PR #555, opened 2023-04-03
feat(Braze) Enable Migration Event IN-1213

CMasterson:
## Summary
Add new Braze event 'SIGNED_IN_USER_UPGRADE_DID_START' when user begins migration (presumably from 7 to 8)

## Implementation Details
Added `signedInUserDidBeginMigration()` func to PocketBraze. When a migration is to be performed a closure should be passed into the perform function that contains a call to the above function.

## Test Steps
Set up a user with a v7 app and account
Log into v8 and check Braze for the event.

## PR Checklist:


PocketKit/Sources/PocketKit/PocketAppDelegate.swift
```diff
@@ -94,6 +94,7 @@ public class PocketAppDelegate: UIResponder, UIApplicationDelegate {
         do {
             let attempted = try legacyUserMigration.perform()
             if attempted {
+                Services.shared.braze.signedInUserDidBeginMigration()
```
bassrock, line 97: Does this run after or before the migration runs? We want this as soon as possible so we can trigger an In app message

Gio2018:
> Why are the README updates in here?!

Probably rebasing `develop` should fix?

PocketKit/Sources/PocketKit/PocketAppDelegate.swift
Gio2018: as far as I can tell it'll run after a successful migration (happens inside `perform()`) and before anything gets loaded, am I right @CMasterson ?

bassrock: ok @CMasterson we want this to happen after we determined they need a migration, but before it is run

Gio2018, line 97: Would it be possible to add a `braze` (or similar name) property, similarly to what we do with `source`, `userDefaults` etc instead of directly calling the singleton here? This would make it easier if we switch to a full dependency injection pattern here.

Gio2018 (commented):
Hey @CMasterson , the code

PocketKit/Sources/PocketKit/PocketAppDelegate.swift
Gio2018: There is an `isRequired` property in `LegacyUserMigration` that might come in handy

CMasterson: I think it's very slightly after, but we know that the migration was actually attempted.
I can dig down a little and see if there's a spot we know the migration will be tried.

CMasterson: This is actually already the case. The `braze` property is of type `BrazeProtocol`

CMasterson: Modified to pass a closure into the `perform` function.
That closure is executed if a migration is about to be performed.

Gio2018: Actually what I meant (sorry for the confusion) is to replace

`Services.shared.braze.signedInUserDidBeginMigration()`

with 

`brazeService.signedInUserDidBeginMigration()`

`brazeService` would be an additional property in `PocketAppDelegate`, initialized from `Services`, similarly to the other existing properties.

PocketKit/Sources/SharedPocketKit/Migrations/LegacyUserMigration.swift
```diff
@@ -68,18 +67,30 @@ public class LegacyUserMigration {
...
         do {
-            legacyStore = try JSONDecoder().decode(LegacyStore.self, from: decryptedData)
+            return try JSONDecoder().decode(LegacyStore.self, from: decryptedData)
         } catch {
             throw LegacyUserMigrationError.failedDeserialization(error)
         }
+    }
+
+    @discardableResult
+    public func perform(migrationAnalytics: () -> Void) throws -> Bool {
```
Gio2018, line 82: can we rename `migrationAnalytics`? If I understand correctly this is meant to launch a braze message instead of tracking analytics with SnowPlow?

Gio2018 (approved):
Looks good to me @CMasterson , just one non blocking comment.

PocketKit/Tests/PocketKitTests/Support/MockPocketBraze.swift
```diff
@@ -46,4 +46,8 @@ extension MockPocketBraze {
 
         return calls[index] as? DidRecieveUserNotifcationImpl
     }
+
+    func signedInUserDidBeginMigration() {
+
+    }
```
Gio2018, line 52: is this needed

CMasterson: It's part of the protocol so it must be implemented in the mock.

PocketKit/Sources/SharedPocketKit/Migrations/LegacyUserMigration.swift
CMasterson: I wanted to keep "braze" out of LegacyUserMigration as much as possible.
`fireMigrationEvent`?

Gio2018: I agree that we don't need to put the braze name: looks like that closure executes a pre-migration task, so whichever name resembles that task is fine for me.

PR #556, opened 2023-04-03
Core data - Refactor Space to use child contexts when needed

Gio2018:
## Summary
- This PR does some Core Data related refactors:
    
## Implementation Details
- Update `Space`
        - Reorders methods by entity
        - Add `makeChildBackgroundContext()` to create a child background context to `backgroundContext`, to better isolate fetch/update operations
        - Add methods `updateSlate(_ :)` and `updateLineup(_ :)` to abstract Core Data logic away from `SlateService`. The above methods use a child context
        - Injects a context instead of hardcoding `backgroundContext` in fetch and update methods

> **Note**
       This is still WIP, there will still be some methods were we still use `backgroundContext`, this will change in future PRs.

 - Update `Pocket (iOS)` Scheme
        - Add the following launch args:
           - -com.apple.CoreData.SQLDebug 1 (disabled)
           - -com.Apple.CoreData.ConcurrencyDebug 1 (enabled)
       - Enables the following diagnostics
           - Thread Sanitizer
       - Disables the following diagnostics
           - Zombie Objects
- Update `Source`
    - Remove method `fetchSlate(_ :)` related implementations, calls and unit tests.
## Test Steps
* Checkout/build/run this branch
* Use the app as much as you can and check that it behaves as expected and does not crash. Test in particular:
    - Login/Logout with an account that has many items
    - Save/unsave, archive/unarchive, favorite/unfavorite
    - Scroll the list, navigate between tabs
    - Stop and relaunch the app several times
    - Go offline/online

## Optional tests
 - Do a raindance
 - Test in different weather conditions, outside, inside, night, day, climb a mountain and test, etc etc

## PR Checklist:
- [x] ~~Added Unit / UI tests~~ removed some tests due to refactor
- [x] Basic Self QA

## Screenshots
NA


PocketKit/Sources/Sync/Space.swift
```diff
@@ -272,57 +324,112 @@ public class Space {
...
+            SlateLineup(context: context, remoteID: remote.id, expermimentID: remote.experimentId, requestID: remote.requestId)
+
+            lineup.update(from: remote, in: self, context: context)
+        }
+        // save the child context
+        try context.performAndWait {
+            guard context.hasChanges else {
+                return
+            }
+            try context.save()
```
bassrock, line 354: @Gio2018 I think we should do the batch deleting of orphans before we save. That would represent a full logical transaction. And ensure we save within this code path only once after we do everything we need to do to trigger a view update.

PocketKit/Sources/Sync/Space.swift
```diff
@@ -272,57 +324,112 @@ public class Space {
...
-        backgroundContext.performAndWait {
-            backgroundContext.refresh(object, mergeChanges: mergeChanges)
+        try context.performAndWait {
+            guard context.hasChanges else {
+                return
+            }
+            // save the child context
+            try context.save()
+            // then save the parent context
+            try save()
```
bassrock, line 382: Remembering to save the parent context, is gonna be fun 😬

PocketKit/Sources/Sync/Space.swift
dskuza: fwiw I don't think batch deleting affects anything on a save, since its behavior is different, so i'm not convinced that order necessarily matters, but i understand what you're suggesting.

> [NSManagedObjectContext](https://developer.apple.com/documentation/coredata/nsmanagedobjectcontext) doesn’t automatically merge a request’s deletions because they happen at the SQL level. Subsequently, you must remove any deleted objects from memory after the request finishes. To determine the objects a request deletes, configure it to return the [NSManagedObjectID](https://developer.apple.com/documentation/coredata/nsmanagedobjectid) of each deleted object and use those identifiers to update your contexts, as the following example shows:

https://developer.apple.com/documentation/coredata/nsbatchdeleterequest

PocketKit/Sources/Sync/Space.swift
```diff
@@ -16,201 +15,241 @@ public class Space {
...
+            guard let object = backgroundObject(with: object.objectID) else {
+                return
+            }
+            context.delete(object)
+        }
     }
 
-    func fetchSavedItems(bySearchTerm searchTerm: String, userPremium isPremium: Bool) throws -> [SavedItem]? {
-        return try fetch(Requests.fetchSavedItems(bySearchTerm: searchTerm, userPremium: isPremium))
+    func delete(_ byIDs: [NSManagedObjectID], for entity: NSEntityDescription) throws {
```
dskuza, line 47: Do we also want to update this function to accept a context?

dskuza (approved):
Just some comments, but nothing blocking.

PocketKit/Sources/Sync/Space.swift
Gio2018: Thanks for the comments @bassrock @dskuza ! I updated the code in favor of non-batch deletion of orphaned items, since we are not dealing with a lot of data here.

PocketKit/Sources/Sync/Space.swift
Gio2018: I agree @cyndichin! I'd only postpone this to a future PR because it feels a but odd to have to pass one context while having the other one hardcoded. A bit more refactoring will come and, at that point, an helper method will come in handy for sure.

Pocket.xcodeproj/xcshareddata/xcschemes/Pocket (iOS).xcscheme
```diff
@@ -346,13 +355,6 @@
             isEnabled = "NO">
          </EnvironmentVariable>
       </EnvironmentVariables>
-      <AdditionalOptions>
-         <AdditionalOption
-            key = "NSZombieEnabled"
-            value = "YES"
-            isEnabled = "YES">
```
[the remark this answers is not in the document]
Gio2018: It's a debugging aid, that keeps deallocated objects forcibly alive to see if something calls them after deallocation, but it also keeps them unnecessarily in memory, so we decided to remove those (was discussed in an Office Hours session)

PocketKit/Sources/Sync/Space.swift
```diff
@@ -16,201 +15,241 @@ public class Space {
...
         backgroundContext.persistentStoreCoordinator?.managedObjectID(forURIRepresentation: url)
     }
 
-    func fetchSavedItem(byRemoteID remoteID: String) throws -> SavedItem? {
-        return try fetch(Requests.fetchSavedItem(byRemoteID: remoteID)).first
-    }
-
-    func fetchSavedItem(byRemoteItemID remoteItemID: String) throws -> SavedItem? {
-        return try fetch(Requests.fetchSavedItem(byRemoteItemID: remoteItemID)).first
+    /// Performs the given request on the specified context. If context is nil, the default `backgroundContext` is used
```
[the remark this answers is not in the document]
Gio2018: See discussion on Slack!!

Pocket.xcodeproj/xcshareddata/xcschemes/Pocket (iOS).xcscheme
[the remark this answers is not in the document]
bassrock: @cyndichin we never want NSZombies enabled unless debugging. In a prod environment a zombie will crash the app. So while we develop, we should be feeling the crashes.

PocketKit/Sources/Sync/Space.swift
```diff
@@ -16,201 +15,241 @@ public class Space {
...
+                return
+            }
+            context.delete(object)
+        }
     }
 
-    func fetchSavedItems(bySearchTerm searchTerm: String, userPremium isPremium: Bool) throws -> [SavedItem]? {
-        return try fetch(Requests.fetchSavedItems(bySearchTerm: searchTerm, userPremium: isPremium))
+    func delete(_ byIDs: [NSManagedObjectID], for entity: NSEntityDescription) throws {
+        try backgroundContext.performAndWait {
```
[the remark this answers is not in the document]
Gio2018: See comment [here](https://github.com/Pocket/pocket-ios/pull/556#discussion_r1158774463) related to the same line.

PocketKit/Sources/Sync/Space.swift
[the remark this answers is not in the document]
bassrock: childContext always. @Gio2018 is working on removing `backgroundContext` usage, iteratively.

PR #557, opened 2023-04-04
feat(coordinators): enabling background refresh coordinators

bassrock:
## Summary

Pocket needs to background process data so that when you open the app, the views you see are up to date without much waiting for the latest.

This PR aims to accomplish the following:
* Abstract Background refresh logic consisting of:
    * Background scheduling
    * Session handling of refresh data
    * Reloading data when the user opens the app
    * Handling 
* Create a consistent way to refresh data from our ViewModels without loading via Source.

We will handle refreshing data in the following way:
* Home is setup to refresh as a BGTaskRefreshRequest, which is designed to keep app data up to date. and we can only have 1 of these at a time.
    * It is allowed to refresh every 6 hours minimum, unless a user pulls to refresh, at which point we will reset the 6 hours. 
* Saves, Archives, and UnresolvedSaveItems are setup as a BGTaskProcessingRequests, which is designed for data processing and we can have 10 scheduled at a time and specify that we need Network Connectivity.
* Tags does not BGrefresh data because of how the backend APIs are designed. Instead it will ever only do a full refresh once, unless a user manually does a pull to refresh (not yet implemented, but this pr got too big and I see @cyndichin and I may conflict if i tocuh it, so ill wait to add it.), or a Saved/Archived item comes in with new Tag data. 
     * Note that any time a tag is changed across the Pocket platform, we will get any Saved/Archived Item affected by the tag change returned by the Saves/Archive api with the new tags. (this is how the old api worked)

## References 
* IN-1288

## Implementation Details
I created an AbstractRefreshProtocol that we can implement anytime we need to handle refreshing of data behavior. It hides a lot of the business logic so that if we need to change how bg refreshing is done it will take effect across the app and lets our indfividual refresh class focus on their specific refresh behavior.

## Test Steps
* 
## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA


PocketKit/Sources/PocketKit/Refresh/AbstractRefreshCoordinatorProtocol.swift
```diff
@@ -0,0 +1,166 @@
...
+    case processing
+    /// Instructs us to create a Refresh Request, we are allowed 1 of thus type schdeduled at a time
+    case refresh
+}
+
+protocol AbstractRefreshCoordinatorProtocol: AnyObject {
+    var notificationCenter: NotificationCenter { get set }
+    var taskScheduler: BGTaskSchedulerProtocol { get set }
+    var appSession: AppSession { get set }
+    var subscriptions: [AnyCancellable] { get set }
```
Gio2018, line 22: it looks like, aside from `subscriptions`, the others could just be `{ get }`?

PocketKit/Sources/PocketKit/Refresh/AbstractRefreshCoordinatorProtocol.swift
```diff
@@ -0,0 +1,166 @@
...
+            try taskScheduler.submit(request)
+        } catch {
+            Log.warning("Could not submit background task request for \(type(of: self))")
+            Log.capture(error: error)
+        }
+    }
+
+    /// Private function that calls the underlying refresh function
+    /// - Parameter completion: Completion to call when done refreshing data
+    private func refreshData(_ completion: @escaping () -> Void) {
```
Gio2018, line 143: nit: we could make the closure optional if there are cases (like some above) where we don't want to pass one?

PocketKit/Sources/PocketKit/Refresh/ArchiveRefreshCoordinator.swift
```diff
@@ -0,0 +1,37 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import Foundation
+import Sync
+import SharedPocketKit
+import Combine
+
+/// Refresh coordinator to handle the refreshing of a Users Archive data
+class ArchiveRefreshCoordinator: AbstractRefreshCoordinatorProtocol {
+    var taskID: String = "com.mozilla.pocket.refresh.archive"
```
Gio2018, line 12: does this (and all the others) need to be a `var` or could it be a `let`?

PocketKit/Sources/PocketKit/Refresh/HomeRefreshCoordinator.swift
```diff
@@ -1,52 +1,60 @@
...
+    ///   - isForced: True when a user manually asked for a refresh
+    ///   - completion: The completion call back to run when the code is done running
     func refresh(isForced: Bool, _ completion: @escaping () -> Void)
 }
 
-class HomeRefreshCoordinator: HomeRefreshCoordinatorProtocol {
-    static let dateLastRefreshKey = "HomeRefreshCoordinator.dateLastRefreshKey"
-    private let notificationCenter: NotificationCenter
-    private let userDefaults: UserDefaults
+class HomeRefreshCoordinator: AbstractRefreshCoordinatorProtocol, HomeRefreshCoordinatorProtocol {
```
Gio2018, line 19: what's the need of conforming to `HomeCoordinatorProtocol` since `AbstractRefreshCoordinatorProtocol` already has a `refresh(_ :, _ :)` method?

PocketKit/Sources/SharedPocketKit/LastRefresh.swift
```diff
@@ -0,0 +1,113 @@
+import Foundation
+
+public protocol LastRefresh {
+    var lastRefreshSaves: Double? { get }
+    var lastRefreshArchive: Double? { get }
+    var lastRefreshTags: Double? { get }
+    var lastRefreshHome: Double? { get }
```
Gio2018, line 7: nit: because they are time intervals, for readability you could use the `TimeInterval` typeAlias here

Gio2018 (approved):
Looks good @bassrock ! Left a few non blocking comments, some of which we discussed already, other than that :shipit: !

PocketKit/Sources/PocketKit/Refresh/AbstractRefreshCoordinatorProtocol.swift
```diff
@@ -0,0 +1,166 @@
...
+import SharedPocketKit
+
+enum BackgroundRequestType {
+    /// Instructs us to create a Processing Request, we are allowed 10 different tasks of this type scheduled at a time
+    case processing
+    /// Instructs us to create a Refresh Request, we are allowed 1 of thus type schdeduled at a time
+    case refresh
+}
+
+protocol AbstractRefreshCoordinatorProtocol: AnyObject {
```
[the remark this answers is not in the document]
bassrock: I kinda of like at leasting having the word Protocol mainly cause its easier to search for. but idc either way

PocketKit/Sources/PocketKit/Refresh/AbstractRefreshCoordinatorProtocol.swift
bassrock: i think we should always pass one but thats just me.

PocketKit/Sources/PocketKit/Refresh/AbstractRefreshCoordinatorProtocol.swift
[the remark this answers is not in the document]
Gio2018: either way is fine for me too

PR #559, opened 2023-04-04
refactor: use shared user defaults

dskuza:
## Summary

Use shared user defaults rather than separate ones (for things such as first sign out). Additionally, do not clean up legacy user defaults (since these will be needed until full app migration / transition). This also introduces the use of a new enum, `UserDefaults.Key`, to consolidate our "library" of Pocket-used keys, and also aid in clearing `UserDefaults` of our used keys as necessary.

## References 

IN-1201

## Implementation Details

- Consolidates all usage of `UserDefaults` throughout the app into a single suite - the shared app group.

- Introduce `UserDefaults.Key`
  - A consolidated set of all `UserDefaults` keys that are used within the Pocket project
  - Extends `UserDefaults` to support the use of `Key` rather than `String`
  - Extends `AppStorage` to support the use of `Key` rather than `String`
  - Adds `UserDefaults.resetKeys()` to remove all (permitted) keys from an instance of `UserDefaults`

Currently, the keys' raw values match the existing keys used in the app prior to this pull request.

## Test Steps

- [ ] Basic app regression testing

## PR Checklist:

- [ ] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/SharedPocketKit/UserDefaultsKey.swift
+++ b/PocketKit/Sources/SharedPocketKit/UserDefaultsKey.swift
@@ -1,0 +1,102 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import Foundation
+import SwiftUI
+
+public extension UserDefaults {
+    enum Key: String, CaseIterable {
+        case hasAppBeenLaunchedPreviously = "hasAppBeenLaunchedPreviously"
+        case recentSearches = "Search.recentSearches"
+        case toggleAppBadge = "AccountViewModel.ToggleAppBadge"
+        case appBadgeToggle = "Settings.ToggleAppBadge"
+        case legacyUserMigration = "com.mozilla.pocket.next.migration.legacyUser"
+        case dateLastRefresh = "HomeRefreshCoordinator.dateLastRefreshKey"
+        case lastRefreshedTagsAt = "lastRefreshedTagsAt"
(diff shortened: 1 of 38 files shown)


Gio2018 (approved):
Looks good @dskuza !

PocketKit/Sources/SharedPocketKit/UserDefaultsKey.swift
```diff
@@ -0,0 +1,102 @@
...
+        case appBadgeToggle = "Settings.ToggleAppBadge"
+        case legacyUserMigration = "com.mozilla.pocket.next.migration.legacyUser"
+        case dateLastRefresh = "HomeRefreshCoordinator.dateLastRefreshKey"
+        case lastRefreshedTagsAt = "lastRefreshedTagsAt"
+        case lastRefreshedArchiveAt = "lastRefreshedArchiveAt"
+        case lastRefreshedSavesAt = "lastRefreshedSavesAt"
+        case listSelectedSortForSaved = "listSelectedSortForSaved"
+        case listSelectedSortForArchive = "listSelectedSortForArchive"
+        case readerFontSizeAdjustment = "readerFontSizeAdjustment"
+        case readerFontFamily = "readerFontFamily"
```
bassrock, line 22: @dskuza in a followup can we add comments for the intended use of each key? While I know due to pocket experience what readerFontFamily is, others may not know right away.

bassrock (approved):

PocketKit/Sources/SharedPocketKit/UserDefaultsKey.swift
dskuza: :+1: that'll be a cross-team effort, but a good idea

PR #563, opened 2023-04-05
fix(tests): upping all our timeouts

bassrock:
## Summary

I believe that a lot of our tests are all flakey because the timeout is too tight.

Our tests should not test performance, hence our timeouts should be allowed to be looser.

## Implementation Details
In the words of @nzeltzer 
> The reason to include timeouts is so that tests fail, not to enforce performance expectations. It’s reasonable to expect that if a test doesn’t pass in 10 seconds that it never will. A 1 second wait suggests to me that we’re simultaneously trying to test that something is happening fast enough. We should test performance explicitly through performance monitoring tests and tools, not incidentally as part of a test of works/fails.


cyndichin (approved):
just a thought, do we just want to up only the ones that we have seen as being flakey as opposed to all of them? Or going forward we should default by 10 to avoid potential flakiness.

Gio2018 (approved):

PR #566, opened 2023-04-06
feat: add and update utm_source query params for auth or sharing

dskuza:
## Summary

Updates the `utm_source` value on login / signup and add the `utm_source` query param when sharing items (from home, saves, and reader).

## References 

IN-1273
IN-942

### Auth

We were already adding a `utm_source` query parameter to the appropriate auth endpoint, so this value just had to be updated.

### Share

Added a `premiumShareURL` helper function that will append / replace an existing `utm_source` query parameter using the correct source value based on the context in which the activity item was generated (such as home, saves, reader).

After discussing with the analytics team, the three possible sources are:

```
pocket_home
pocket_saves
pocket_reader
```

Unit tests for this, as well as the `pocketPremiumURL` helper function, were added.

## Test Steps

- [ ] When logged out, after tapping login / signup, a network monitor will show `ios_next` as the `utm_source` for the page that was loaded
- [ ] When sharing an item from "Home", selecting Copy (link) and pasting the link will show the correct `utm_source` value based on its use within the context of home, saves, or the reader.

## PR Checklist:

- [x] Added Unit / UI tests
- [x] Basic Self QA


Gio2018 (approved):
:shipit:

PR #573, opened 2023-04-07
fix(user): ensuring we only refresh the user when we have a session

bassrock:
## Summary

Ensure that user data only refreshes when there is an active session.

## References 
* IN-1238

## Implementation Details
* Switched user data to the refresh coordinator pattern

## Test Steps
* Ensure there are no FetchUser errors in the console on logout and login.

## PR Checklist:
- [ ] Added Unit / UI tests
- [ ] Basic Self QA


Gio2018 (approved):
:shipit:

PR #568, opened 2023-04-06
feat(tags): add recent tags

cyndichin:
## Summary
Add Recent Tags section to both Edit and Add Tags screen

* This work does not include guarding recent tags with premium check (this will be done in another PR)

## References 
IN-1268

## Implementation Details
Created `RecentTag` component and add a case of `recent(String)`  to enum `TagType` . Added `x` to the input tags in Add Tags view. Created a `RecentTagProvider` class to hold the logic of recent tags to be used across the three view models related to tags `PocketAddTagsViewModel`, `SaveToAddTagsViewModel` and `TagsFilterViewModel`.

Moved some files that were related to tags to `SharedPocketKit` instead of `Textile` as it made more sense for it to belong there.

## Test Steps
Verify the AC below: 

There should be two sections: (1) Recent Tags, which shows the three most-recently used tags, and (2) All Tags, which shows all of a user’s tags. Both sections should be sorted alphabetically. 
1. Recent Tags for both screens should be consistent 
2. Recent Tags should appear first in the tags list 
3. Recent Tags should be distinguishable in some way 
4. Recent Tags should be sorted by most-recent use (adding a tag to an item)
5. All Tags should be sorted by alphabetical order
6. Add x to right side of tag for Add Tags screen

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Self Review (review, clean up, documentation, run tests)
- [x] Basic Self QA

## Screenshots

https://user-images.githubusercontent.com/6743397/230919581-294613cc-875c-4130-a432-508d64beb26b.mp4

PR #576, opened 2023-04-10
fix(tests): battle hardened tests against multiple device sizes

bassrock:
## Summary

@timc-mozilla discovered that our test was not battle hardended against small device sizes. This makes use of a scrollTo function to ensure that we scroll to a specific element instead of unless scrolling or assuming something is on screen.

This will also fixes flakyness I discovered around tags where tags were always appearing in a different order. This should be fully fixed in https://github.com/Pocket/pocket-ios/pull/568 and any work here is a stop gap for tests and sanity.

## Implementation Details
There are a few helper [libraries](https://github.com/PGSSoft/AutoMate) we could import that would help with our tests, but I chose to use a very basic scrollTo function instead since the one I found had not been updated in a while.

## Test Steps
* Try the add tags tests on a iphone mini

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/PocketKit/Tags/TagsFilterViewModel.swift
+++ b/PocketKit/Sources/PocketKit/Tags/TagsFilterViewModel.swift
@@ -21,7 +21,8 @@ class TagsFilterViewModel: ObservableObject {
     }
 
     func getAllTags() -> [TagType] {
-        arrangeTags(with: fetchedTags?.compactMap({ $0.name }) ?? [])
+        // Finding unique elements using set
+        arrangeTags(with: Array(Set(fetchedTags?.compactMap({ $0.name }) ?? [])))
     }
 
     func trackEditAsOverflowAnalytics() {
--- a/PocketKit/Sources/Textile/Views/Tags/TagType.swift
+++ b/PocketKit/Sources/Textile/Views/Tags/TagType.swift
@@ -20,15 +20,5 @@ public enum TagType: Hashable {
 /// - Parameter tags: list of users tag names
 /// - Returns: converts users tags to display a list of `TagType`
 public func arrangeTags(with tags: [String]) -> [TagType] {
-    var allTags: [String] = []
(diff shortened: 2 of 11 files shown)


PocketKit/Sources/PocketKit/Tags/TagsFilterViewModel.swift
```diff
@@ -21,7 +21,8 @@ class TagsFilterViewModel: ObservableObject {
     }
 
     func getAllTags() -> [TagType] {
-        arrangeTags(with: fetchedTags?.compactMap({ $0.name }) ?? [])
+        // Finding unique elements using set
+        arrangeTags(with: Array(Set(fetchedTags?.compactMap({ $0.name }) ?? [])))
```
bassrock, line 25: I had to do this, because at times the FilterTagsView would contain duplicate tags. Should be fixed in https://github.com/Pocket/pocket-ios/pull/568

PocketKit/Sources/Textile/Views/Tags/TagType.swift
```diff
@@ -20,15 +20,5 @@ public enum TagType: Hashable {
...
-    if fetchedTags.count > 3 {
-        let topRecentTags = Array(fetchedTags)[..<3]
-        let sortedTags = Array(fetchedTags)[3...].sorted()
-        allTags.append(contentsOf: topRecentTags)
-        allTags.append(contentsOf: sortedTags)
-    } else {
-        allTags.append(contentsOf: fetchedTags)
-    }
-    return allTags.compactMap { TagType.tag($0) }
+    return tags.sorted().compactMap { TagType.tag($0) }
```
bassrock, line 23: removing recent tags instead for https://github.com/Pocket/pocket-ios/pull/568

timc-mozilla (approved):
looks cool! excited for these changes

PR #579, opened 2023-04-10
Add Analytics tracking to v7 -> v8 migration

CMasterson:
## Summary
Add new tracking types and fire them from PocketKits and SaveToPocketKit

## References 
https://getpocket.atlassian.net/browse/IN-1277

## Implementation Details
Add new Event type, System. Other event types relate to some manner of user interaction and were therefore unsuitable for use here in a purely passive event.
A System Event is qualified by it's SystemEventType (Which as of right now only supports a UserMigration event)
To support the UserMigration event a UserMigrationState enum was also added to add granularity to the stage of migration the user is currently at.

## Test Steps

- Log into a v7 instance of Pocket.
- Close the app and update to v8
- Launch the app or use the SaveToPocketKit extension
- Check for suitable Analytics events.

## PR Checklist:
- [ ] Added Unit / UI tests
- [x] Basic Self QA


PocketKit/Sources/Analytics/AppEvents/Migration.swift
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     Fired when the user succeeds migrating from V7 to V8
+     */
+    static func MigrationTo_v8DidSucceed(source: MigrationSource) -> Background {
+        return Background(type: .userMigration(.succeeded), source: source)
+    }
+
+    /**
+     Fired when the user begins migrating from V7 to V8
```
CMasterson, line 34: Missed this comment.

PocketKit/Sources/Analytics/AppEvents/Migration.swift
```diff
@@ -0,0 +1,45 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import Foundation
+import SharedPocketKit
+
+/// Used to track migration event source
+public enum MigrationSource: String {
```
CMasterson, line 9: I don't think this needs to be a string.

timc-mozilla (approved):

PR #580, opened 2023-04-11
fix(saves): ensure that dual refreshing saves does not reset the loading state

bassrock:
## Summary

Moves the sending the `.start` state within our operation queue, where it is guarded from being called multiple times before it is finished loading.

## References 
* IN-1311

## Implementation Details
* we need to think about a better way to do placeholders and pagination, but not atm

## Test Steps
* Try and perform the replication as done by Matt here: https://pocket.slack.com/archives/C4DL4QSU8/p1681169244912969?thread_ts=1680970901.903769&cid=C4DL4QSU8
* Note you must pull to refresh again, as soon as you see your first set of list results.

## PR Checklist:
- [ ] Added Unit / UI tests
- [ ] Basic Self QA

### DIFF
--- a/PocketKit/Sources/Sync/PocketSource.swift
+++ b/PocketKit/Sources/Sync/PocketSource.swift
@@ -264,10 +264,6 @@ public class PocketSource: Source {
 // MARK: - Saves/Archive items
 extension PocketSource {
     public func refreshSaves(completion: (() -> Void)? = nil) {
-        if lastRefresh.lastRefreshSaves == nil {
-            initialSavesDownloadState.send(.started)
-        }
-
         let operation = operations.fetchSaves(
             apollo: apollo,
             space: space,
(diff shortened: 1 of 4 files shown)


PocketKit/Sources/Sync/PocketSource.swift
```diff
@@ -264,10 +264,6 @@ public class PocketSource: Source {
 // MARK: - Saves/Archive items
 extension PocketSource {
     public func refreshSaves(completion: (() -> Void)? = nil) {
-        if lastRefresh.lastRefreshSaves == nil {
```
bassrock, line 267: This function could be called mutltiple times, moving this send state to the Operation is safer because the operation can only be called one at a time.

cyndichin (approved):

Gio2018 (approved):
pretty straightforward @bassrock ! :shipit:

PR #583, opened 2023-04-11
feat(tags): refine recent tags experience

cyndichin:
## Summary
Refining recent tags experience: 
* Allow only Premium users to view the recent tags and if their tags are greater than 3
* Updating UI and Unit Tests
* Adding Analytics for tapping on recent tags

## References 
IN-1268

## Implementation Details
* Added a check to `recentTags` to be displayed if user is Premium and if the number of all their tags is greater than 3.
* Added two tracking called `addTagsRecentTagTapped` and `filterTagsRecentTagTapped`
* Added `user` in `Services` for shared extension
* Moved `RecentTag` to its own file
* Created test for recent tags

## Test Steps
Please 
There should be two sections: (1) Recent Tags, which shows the three most-recently used tags, and (2) All Tags, which shows all of a user’s tags (including the three most-recently used). Both sections should be sorted alphabetically. 
1. Recent Tags and Add Tags screens should be consistent 
2. Recent Tags should appear first in the tags list 
3. Recent Tags should be distinguishable in some way 
4. Recent Tags should be sorted by most-recent use
5. All Tags should be sorted by alphabetical order
6. Recent Tags should only be visible to Premium Users
7. Add x to right side of tag for Add Tags screen

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

## Screenshots

https://user-images.githubusercontent.com/6743397/231266937-728ce0ce-3bbe-4236-b5b1-76abfd368a1b.mp4

### DIFF
--- a/PocketKit/Tests/PocketKitTests/PocketAddTagsViewModelTests.swift
+++ b/PocketKit/Tests/PocketKitTests/PocketAddTagsViewModelTests.swift
@@ -120,7 +120,7 @@ class PocketAddTagsViewModelTests: XCTestCase {
         XCTAssertNotNil(source.addTagsToSavedItemCall(at: 0))
     }
 
-    func test_recentTags_withLessThanThreeTags_andPremiumUser_returnsNoRecentTags() {
+    func test_recentTags_withThreeTags_andPremiumUser_returnsNoRecentTags() {
         let item = space.buildSavedItem(tags: [])
         let expectFetchAllTagsCall = expectation(description: "expect source.fetchAllTags()")
         expectFetchAllTagsCall.assertForOverFulfill = false
--- a/PocketKit/Tests/PocketKitTests/PocketAddTagsViewModelTests.swift
+++ b/PocketKit/Tests/PocketKitTests/PocketAddTagsViewModelTests.swift
@@ -2,6 +2,7 @@ import XCTest
 import Combine
 import Analytics
 import Textile
+import SharedPocketKit
 
 @testable import Sync
 @testable import PocketKit
@@ -40,6 +41,7 @@ class PocketAddTagsViewModelTests: XCTestCase {
         source: Source? = nil,
         tracker: Tracker? = nil,
         userDefaults: UserDefaults? = nil,
+        user: User? = nil,
         saveAction: @escaping () -> Void,
         networkPathMonitor: NetworkPathMonitor? = nil
     ) -> PocketAddTagsViewModel {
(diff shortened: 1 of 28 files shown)


PocketKit/Tests/PocketKitTests/PocketAddTagsViewModelTests.swift
```diff
@@ -118,26 +120,85 @@ class PocketAddTagsViewModelTests: XCTestCase {
         XCTAssertNotNil(source.addTagsToSavedItemCall(at: 0))
     }
 
-    func test_recentTags_withTags_returnsRecentTags() {
+    func test_recentTags_withLessThanThreeTags_andPremiumUser_returnsNoRecentTags() {
```
dskuza, line 123: The test name doesn't match the implementation, which is a  little confusing. Within the test, there _are_ three tags, but the test name implies 2 or less, and the _true_ implementation of this functionality is > 3 tags.

PocketKit/Tests/SaveToPocketKitTests/SaveToAddTagsViewModelTests.swift
```diff
@@ -104,9 +117,68 @@ class SaveToAddTagsViewModelTests: XCTestCase {
...
+
+    func test_recentTags_withMoreThanThreeTags_andPremiumUser_returnsRecentTags() throws {
+        let item = space.buildSavedItem(tags: [])
+        try space.save()
+        let viewModel = subject(
+            item: space.viewObject(with: item.objectID) as! SavedItem,
+            user: MockUser(status: .premium),
+            retrieveAction: { _ in
+                var tags: [Tag] = []
+                for index in 1...4 {
```
dskuza, line 131: This seems to be building a different array of tags than the assertion is expecting (which is a total of 3, but we're generating 4). Why is that?

dskuza (commented):
Just a few comments, mainly re: tests. Other than that, looks good. I'll just await a brief discussion before giving a :+1:

PocketKit/Tests/SaveToPocketKitTests/SaveToAddTagsViewModelTests.swift
cyndichin: I wanted to test that even if there are four tags, the recent tags should only gather three at most.

PocketKit/Tests/PocketKitTests/PocketAddTagsViewModelTests.swift
cyndichin: yes, you're right! thanks for the catch, updated `withThreeTags`. Wanted to capture if user had a total of three tags or less, then don't show recent tags.

PR #588, opened 2023-04-11
chore(apollo): upgrading apollo to 1.1.1

bassrock:
## Summary

Update Apollo to 1.1.1


cyndichin (approved):
update looks good from code side - approving ~ seems like we'll need to update the failing tests.

PR #593, opened 2023-04-12
fix(braze): ensure braze has a scene

bassrock:
## Summary

Ensure that braze has a scene to show the message on

## Implementation Details
* Still working to figure out why the IAM  via a trigger is not showing on device.

## Test Steps
* Try the commented out code in the init with and without the windowScene code.


cyndichin (commented):
I remember that when we were testing IAM via the Braze dashboard, it worked for me without any modification to the code. Curious to understand why we would need a scene now. Also, testing steps refer to commenting in / out, but do I need to trigger an IAM from Braze too?

PocketKit/Sources/PocketKit/Notifications/PocketBraze.swift
```diff
@@ -108,7 +107,22 @@ extension PocketBraze: BrazeInAppMessageUIDelegate {
         _ ui: BrazeInAppMessageUI,
         prepareWith context: inout BrazeInAppMessageUI.PresentationContext
     ) {
-        // Customize the in-app message presentation here using the context
+        // We need to set our scene for Braze because our UIKit + SwiftUI stuff does not seem to work with Braze's default
```
cyndichin, line 110:
```suggestion
        // We need to set our scene to show our local Braze in-app messaging because our UIKit + SwiftUI stuff does not seem to work with Braze's default
```

> @cyndichin that stuff is in the comment, can you help me make it more clear? That code creates a local in app message.
Ohh okay, added suggestion for the comment. I was more curious to understand why we need this local message when it worked previously without this code (we can also discuss live in Unleash meeting).

bassrock: @cyndichin that i can't answer, i dug through their sdk, and cant see why. But our codes been changing a lot. This is an area where im just accepting something changed, and this fixes it. 🤷‍♂️

cyndichin (approved):
tested with and without the additional code, looks good!

PR #597, opened 2023-04-13
feat(listen): putting playlists behind a feature flag

bassrock:
## Summary

Put's listen's playlist support behind a feature flag, to be released at a later date post MVP.

## Test Steps
* Add your user to `temp.ios.listen.tag_playlists` and `temp.ios.listen`
* Select a tag
* then select listen and notice that the items shown are those that were in the tag

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA


Tests iOS/Support/XCTestCase+extensions.swift
```diff
@@ -15,6 +15,10 @@ extension XCTestCase {
         wait(for: expectations, timeout: 10)
     }
 
+    func fulfillment(of expectations: [XCTestExpectation]) async {
```
bassrock, line 18: This is needed instead of wait starting with Swift 6 but xcode 14.3 has started warning us.

Tests iOS/Support/Elements/PocketAppElement.swift
```diff
@@ -177,3 +177,18 @@ struct PocketAppElement {
         app.activate()
     }
 }
+
+/// Listen
+/// Hacky helper extenstion until we add accessibility identifier helpers to Listen
+extension PocketAppElement {
```
bassrock, line 183: @nzeltzer note that we need to add accessibility identifiers to listen so we can test it.

bassrock: Using the accessibility inspector, i did not see any currently.

PocketKit/Sources/PocketKit/FeatureFlagService.swift
```diff
@@ -60,12 +60,15 @@ class FeatureFlagService: FeatureFlagServiceProtocol {
...
     case listen = "temp.ios.listen"
+    case listenTagsPlaylists = "temp.ios.listen.tag_playlists"
 
     /// Description to use in a debug menu
     var description: String {
         switch self {
-        case.listen:
+        case .listen:
             return "Enable the Listen feature"
+        case .listenTagsPlaylists:
```
dskuza, line 70: I'm missing something; is the debug menu something that's coming, or something that exists? If that's the case, can we create a protocol or something that adds a `debugDescription` to make it more obvious that the description is for debugging purposes? Just an idea, not a blocker.

PocketKit/Sources/PocketKit/MyList/ItemsList/SavedItemsList/SavedItemsListViewModel.swift
```diff
@@ -615,6 +615,14 @@ extension SavedItemsListViewModel {
 
         switch filter {
         case .listen:
+            // If the user selected a filter, and the user is not in the Listen tags playlist feature flag
+            // Remove any filters and sorts they selected and re-fetch data before showing listen.
```
dskuza, line 619: My interpretation: if a user has a tag / other filter selected, if they tap the listen button and they are _not_ in the playlist group, this will "reset" the list? This is just so the "playlist" is the entire list, correct? If my interpretation is correctly, I feel like this _might_ be jarring as a user to see their view "reset", but I think I understand its purpose.

Tests iOS/MyList/ListenTests.swift
```diff
@@ -0,0 +1,115 @@
...
+
+        app.launch().tabBar.savesButton.wait().tap()
+        app.saves.itemView(matching: "Item 1").wait()
+
+        wait(for: [flagsLoaded])
+
+        // do a refresh because the flag prob loaded in the background.
+        app.saves.pullToRefresh()
+
+        app.saves.filterButton(for: "Listen").wait().tap()
```
dskuza, line 56: Is there an expectation that's needed after this to test that Listen actually displayed? However, that'd rely on Listen having an accessibility identifier. Or, maybe there's a way of drilling down the view hierarchy for the right view? …though that's an ugly solution.

dskuza: I think I see it in a test below - `app.listenPlay.wait()` ?

PocketKit/Sources/PocketKit/MyList/ItemsList/SavedItemsList/SavedItemsListViewModel.swift
bassrock: @dskuza correct. This is what we want to do on purpose for now. We will revert this behavior in a point release. Can explain the reasoning in slack.

dskuza (approved):

PR #602, opened 2023-04-13
fix(extension): address feedly bug

cyndichin:
## Summary
Fix bug where we cannot save from Feedly due to URL initialization returning nil

## References 
* IN-1254

## Implementation Details
Create a helper function to parse a URL from a string. Modified from https://www.hackingwithswift.com/example-code/strings/how-to-detect-a-url-in-a-string-using-nsdatadetector

This only applied to the items that are of the type `public.plain-text`

## Test Steps
Nav to the Feedly app
Share any item to Pocket
Verify that error no longer appears and you are able to save the item

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/SaveToPocketKit/SavedItem/SavedItemViewModel.swift
+++ b/PocketKit/Sources/SaveToPocketKit/SavedItem/SavedItemViewModel.swift
@@ -143,7 +143,7 @@ extension SavedItemViewModel {
 
             if provider.hasItemConformingToTypeIdentifier(plainTextUTI) {
                 guard let string = try? await provider.loadItem(forTypeIdentifier: plainTextUTI, options: nil) as? String,
-                      let url = URL(string: string) else {
+                      let url = retrieveURLFromString(with: string) else {
                     continue
                 }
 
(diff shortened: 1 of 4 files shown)


PocketKit/Sources/SaveToPocketKit/SavedItem/SavedItemViewModel.swift
```diff
@@ -143,7 +143,7 @@ extension SavedItemViewModel {
 
             if provider.hasItemConformingToTypeIdentifier(plainTextUTI) {
                 guard let string = try? await provider.loadItem(forTypeIdentifier: plainTextUTI, options: nil) as? String,
-                      let url = URL(string: string) else {
```
cyndichin, line 146: We were receiving `"The Continentals first trailer takes the John Wick series to the 70s https://www.theverge.com/23680378/the-continental-teaser-trailer-john-wick-peacock"` as the string and so this value returned nil

bassrock (approved):

bassrock:
@cyndichin can you also UI test this?

cyndichin:
> @cyndichin can you also UI test this?

I can possible hack one if I figure out how to type in the Reminders app. 
1. Create a reminder with the text "Get Pocket get pocket.com"
2. Highlight that text
3. Navigate to Share setting
4. Verify it has been saved

I was unable to find an app on Simulator that shared in the same format. 
Do you have a better alternative and worth investing time into the steps above?

bassrock:
Ah i see. thats ok then i didnt think about what we needed to simulate this in ui.

cyndichin:
> Ah i see. thats ok then i didnt think about what we needed to simulate this in ui.

Oh I just saw this, let me know if you think we should keep the tests or not. It passes locally.

bassrock:
@cyndichin if the tests work lets keep it! I think they are important

PR #612, opened 2023-04-14
fix(logs): adding explict logging and error handling to mutations

bassrock:
## Summary

While this doesn't solve the root cause on IN-1306. It does start capturing more specific data for us. With the logging added the root problem appears to be the CoreData store SavedItem object somehow becoming invalid and causing the store save method to rollback which in turn does not update the UI. 

However when you are online the problem does not appear, because the server will do a full object refresh with the SavedItem in our store. 

--

I am also proposing that we implement the ability to post an error from Sync or similar to NotificationCenter that will trigger a banner or something to indicate there is an error like this.

## References 
* IN-1306

## Implementation Details

## Test Steps
* Manipulate items and see breadcrumbs in the logs.

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA


nzeltzer (approved):
lgtm

PR #613, opened 2023-04-15
feat(api): moving to api domain

bassrock:
## Summary

In order to seperate traffic on the backend, we need to move clients to api.getpocket.com. Our android app has already been on this domain for a few years, and we should ensure iOS is on it before launch.


cyndichin (approved):
thanks! do we also update here as well?

image

timc-mozilla (approved):

PR #619, opened 2023-04-18
fix(tags): update rename tag functionality

cyndichin:
## Summary
Fix rename tag function so that it does not deactivate upon tag selection after renaming 1 tag.

## References 
IN-940

## Implementation Details
* Separated `EditBottomBar` and moved logic into `Textile`. Update alert UI to utilize iOS 16 Alert View with Textfield and remove UIKit implementation. Updated rename tests. Adding `selectedItems.removeAll()` fixes bug where rename breaks after 1 tag is renamed. 

* As a nice polish, also add scrolling ability after a user renames a tag since the list is sorted alphabetically.

* Added analytics for `renameTag` and `deleteTag`

## Test Steps
1. Nav to Archive > Tagged > Edit
2. Select a tag name
3. Rename it
==> after 1 tag is renamed, the 'Rename' function should not break and works as expected upon another tag name selection 

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA


PocketKit/Sources/PocketKit/Tags/TagsFilterViewModel.swift
```diff
@@ -60,22 +59,24 @@ class TagsFilterViewModel: ObservableObject {
...
+            Log.capture(message: "Unable to rename tag")
+            return
+        }
+        let newName = newName.lowercased()
+        guard let tag: Tag = fetchedTags.filter({ $0.name == oldName }).first,
+              !fetchedTags.compactMap({ $0.name }).contains(newName) else { return }
         source.renameTag(from: tag, to: newName)
-        refreshView = true
+        trackTagRename()
+        completion()
```
cyndichin, line 79: open to suggestions if a better way to wait until this function completes before performing the scrolling logic

PocketKit/Sources/PocketKit/Tags/TagsFilterViewModel.swift
```diff
@@ -60,22 +59,24 @@ class TagsFilterViewModel: ObservableObject {
...
     }
 
-    func rename(from oldName: String, to newName: String) {
-        let event = SnowplowEngagement(type: .general, value: nil)
-        let contexts: Context = UIContext.button(identifier: .tagsSaveChanges)
-        tracker.track(event: event, [contexts])
-        guard let tag: Tag = fetchedTags.filter({ $0.name == oldName }).first else { return }
+    func rename(from oldName: String?, to newName: String, completion: () -> Void) {
+        guard let oldName else {
+            Log.capture(message: "Unable to rename tag")
```
bassrock, line 71: @cyndichin you should call the completion here.

Tests iOS/EditTagsTests.swift
```diff
@@ -28,44 +30,50 @@ class EditTagsTests: XCTestCase {
         try server.start()
     }
 
-    override func tearDownWithError() throws {
+    @MainActor
+    override func tearDown() async throws {
         try server.stop()
         app.terminate()
+        await snowplowMicro.assertNoBadEvents()
```
bassrock, line 37: @cyndichin lets use https://github.com/Pocket/pocket-ios/blob/develop/Tests%20iOS/Support/SnowplowMicro.swift#L296-L298 instead

PocketKit/Sources/Analytics/AppEvents/Tags.swift
```diff
@@ -136,4 +136,26 @@ public extension Events.Tags {
             )
         )
     }
+
+    /// Fired when a user adds tag to an input list in the `Add Tags` screen for an item
+    static func renameTag() -> Event {
```
bassrock, line 141: Im curious for analytics purposes do they need to know the stuff being renamed?

Tests iOS/EditTagsTests.swift
cyndichin: ah nice catch! I realize I was also adding that to individual tests, but can prob remove it.

PocketKit/Sources/Analytics/AppEvents/Tags.swift
cyndichin: let me check with them, I was under the assumption that it could be useful to keep track of how many users are using these functionality.

PocketKit/Sources/PocketKit/Tags/TagsFilterViewModel.swift
cyndichin: I don't want to trigger the code in the completion block, should I return completion(true) or completion(false) in that case then?

Tests iOS/EditTagsTests.swift
cyndichin: updated!

PocketKit/Sources/Sync/PocketSource.swift
```diff
@@ -524,10 +523,10 @@ extension PocketSource {
             let operation = operations.savedItemMutationOperation(
                 apollo: apollo,
                 events: _events,
-                mutation: TagUpdateMutation(input: TagUpdateInput(id: remoteID, name: name))
+                mutation: TagUpdateMutation(input: TagUpdateInput(id: oldTag.remoteID ?? "", name: name))
```
cyndichin, line 526: @Gio2018 with the changes in terms of using name over remoteID, does this look okay?

PocketKit/Sources/PocketKit/Tags/TagsFilterViewModel.swift
cyndichin: triggering the code doesn't have an effect from the user perspective, so will add a completion block there as well. thanks!

PocketKit/Sources/PocketKit/Tags/TagsFilterViewModel.swift
```diff
@@ -50,22 +49,25 @@ class TagsFilterViewModel: ObservableObject {
...
+            completion()
+            return
+        }
+        let newName = newName.lowercased()
+        guard let tag: Tag = fetchedTags.filter({ $0.name == oldName }).first,
+              !fetchedTags.compactMap({ $0.name }).contains(newName) else { return }
```
cyndichin, line 67: oops forgot one more completion and should prob log this too

PocketKit/Sources/PocketKit/Tags/TagsFilterViewModel.swift
bassrock: General rule of thumb, I follow is the function here doesn't know what the completion does or doesnt do, so when you return you need to call it. In this case since the completion doesn't take any error parameters I think it should always be called to signify any work this function does is complete.

PocketKit/Sources/PocketKit/Tags/TagsFilterViewModel.swift
cyndichin: will most likely change with this ticket https://getpocket.atlassian.net/browse/IN-1350

PocketKit/Sources/Analytics/AppEvents/Tags.swift
cyndichin: ooooh sorry I misread that, I can follow up with her on that. Not sure what purposes that would serve.

bassrock: yea no biggie. I was just curious, prob doesn't matter im sure they can corelate on the data model if ever needed.

PocketKit/Sources/PocketKit/Tags/TagsFilterViewModel.swift
bassrock: @cyndichin how come we cant wait for the function to just return? Am i missing something?

Tests iOS/EditTagsTests.swift
```diff
@@ -28,44 +30,49 @@ class EditTagsTests: XCTestCase {
...
-    override func tearDownWithError() throws {
+    @MainActor
+    override func tearDown() async throws {
         try server.stop()
         app.terminate()
+        await snowplowMicro.assertBaselineSnowplowExpectation()
     }
 
-    // TODO: Should be updated with https://getpocket.atlassian.net/browse/IN-940
-//    func test_editTagsView_renamesTag() {
```
bassrock, line 37: I like it when a test comes back

bassrock (approved):
Approving, but i had a question or two about the completion use

PocketKit/Sources/PocketKit/Tags/TagsFilterViewModel.swift
cyndichin: hmm, originally I wanted to make sure that `.onChange(of: renameTagText)` was only being called if we rename the tag. But since we're adding the completion block in both cases, then it doesn't matter anymore actually. I can remove them.

PocketKit/Sources/Analytics/AppEvents/Tags.swift
cyndichin: okay, let me bring a bigger question to her since don't think we're doing that for other tag tracking calls either. will do a follow up PR if anything

PR #626, opened 2023-04-19
feat(flags): updating feature flags to always attach

bassrock:
## Summary

After a recent nimbus integration meeting, we realized it would be useful and best practice to always attach any feature flag entity we may be assigned to our analytics events similar to APIUser and User. 

## Implementation Details
* I added a method that gets called in the refresh coordinators to have a log out call back so we can clear the feature entities on logout.

## PR Checklist:
- [ ] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/PocketKit/Refresh/FeatureFlagsRefreshCoordinator.swift
+++ b/PocketKit/Sources/PocketKit/Refresh/FeatureFlagsRefreshCoordinator.swift
@@ -6,6 +6,7 @@ import Foundation
 import Sync
 import SharedPocketKit
 import Combine
+import Analytics
 
 /// Refresh coordinator to handle the refreshing of feature flags
 class FeatureFlagsRefreshCoordinator: RefreshCoordinator {
@@ -24,14 +25,20 @@ class FeatureFlagsRefreshCoordinator: RefreshCoordinator {
     private let lastRefresh: LastRefresh
 
     private let source: Source
+    private let tracker: Tracker
     var isRefreshing: Bool = false
 
-    init(notificationCenter: NotificationCenter, taskScheduler: BGTaskSchedulerProtocol, appSession: AppSession, source: Source, lastRefresh: LastRefresh) {
+    init(notificationCenter: NotificationCenter, taskScheduler: BGTaskSchedulerProtocol, appSession: AppSession, source: Source, lastRefresh: LastRefresh, tracker: Tracker) {
         self.notificationCenter = notificationCenter
         self.taskScheduler = taskScheduler
         self.appSession = appSession
         self.source = source
         self.lastRefresh = lastRefresh
+        self.tracker = tracker
+    }
+
+    func didLogout() {
+        tracker.resetPersistentFeatureEntities([])
     }
 
     func refresh(isForced: Bool = false, _ completion: @escaping () -> Void) {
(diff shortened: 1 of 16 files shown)


PocketKit/Sources/PocketKit/Refresh/FeatureFlagsRefreshCoordinator.swift
```diff
@@ -46,7 +49,19 @@ class FeatureFlagsRefreshCoordinator: RefreshCoordinator {
...
                 do {
                     self.isRefreshing = true
-                    try await self.source.fetchAllFeatureFlags()
+                    let flags = try await self.source.fetchAllFeatureFlags()
+                    
+                    let flagEntities: [FeatureFlagEntity] = flags.compactMap { flag in
+                        guard let name = flag.name else {
+                            return nil
+                        }
+                        return FeatureFlagEntity(
```
bassrock, line 58: @kirill-demtchouk this makes the change we discussed, in the nimbus call. I think we want to expand the feature falg entity to include `assigned` true or false to make this work the way we like.

kirill-demtchouk: Wow, ok, so pretty easy on iOS! I'm curious whether Nimbus expects to get _all_ feature flags back, with true/false assignments, or just the ones that a client is enrolled in.

bassrock: I can filter to assigned, if we want. or i can just pass it through to the data modeling layer with all.

bassrock: Android should be able to operate in the same manner here.

kirill-demtchouk: @bassrock Would this also include all feature flags that a user is assigned to with `variant_enroll` events?

We currently assume that `variant_enroll` events are only sent for one experiment at a time in data modeling, specifically the experiment that the user was assigned to that initiated that event in the first place.

bassrock: 🤔  It would. Although. Question, (in perfect world) do we need `variant_enroll` if all flags were always sent. Wouldnt we just need an event to signify the user felt the change? (which could be an analytics event on the actual thing we are doing)

kirill-demtchouk: > Wouldnt we just need an event to signify the user felt the change?
That's exactly right! It's imperative for experiment analysis to know the exact moment a user was enrolled in a test. This is the current purpose of the `variant_enroll` event and why I'm asking about it 😅 

We could infer the moment of their enrollment in a specific test from an event stream with all of their feature flags on it, but it's a lot more error-prone on the modeling side than having the client emit a single event with a single feature flag.

In the future, I imagine we could move the specific feature flag relevant to a `variant_enroll` event into the event properties for that event and then include an array of all feature flags in the `feature_flag` entity (so the handling of the `feature_flag` entity isn't different for any given event type).

timc-mozilla (approved):

bassrock:
moving to draft to address https://github.com/Pocket/pocket-ios/pull/626#discussion_r1173995664

kirill-demtchouk:
@bassrock Thanks for moving to draft! Let's discuss this some more once we're clear of the public release 😄

PR #605, opened 2023-04-14
Core Data - Refactor entities to match required and non-nullable fields with the graph

Gio2018:
## Summary
* This PR:
    - Sets the code gen to `Manual/none` for all Core Data entities
    - Matches the nullability and uniqueness of fields with the objects in the graph

## Implementation Details
* All entities were updated as described in the summary.
> **Note**
since we are updating fields, Core Data will attempt a lightweight migration. This might cause instability and/or crashes while updating from a previous build. It is recommended to delete and reinstall. This issue will not affect production users migrating from Pocket 7 since they won't have a previous version of the Core Data DB.

## Test Steps
* Build/run this branch and make sure that the app works as expected. Especially look for crashes or any data inconsistency in Home and Saves

## PR Checklist:
- [ ] Added Unit / UI tests
- [x] Self Review (review, clean up, documentation, run tests)
- [x] Basic Self QA

PR #628, opened 2023-04-19
ci(tests): fix unit tests after core data changes

cyndichin:
## Summary
Fix tests, see original PR for context
https://github.com/Pocket/pocket-ios/pull/605

## References 
N/A

## Implementation Details
* We updated SavedItem and Item to have unique properties (i.e. given_url and url and remoteID), but the items to set up these tests were using the same values for the saved Items causing errors. Setting up the items to have unique values, causes the tests pass.
* Need to create `remoteID` with tag creation in `fetchOrCreateTag` (needs to be revisited)
* Added `createdAt` when building pending saved item

## Test Steps
Make sure tests are passing.

## PR Checklist:
- N / A Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/Sync/Space/Space.swift
+++ b/PocketKit/Sources/Sync/Space/Space.swift
@@ -393,12 +393,14 @@ extension Space {
         }
         let createTag = Tag(context: context)
         createTag.name = name
-        createTag.remoteID = "remoteID\(name)"
         return createTag
     }
 
-    func fetchTag(byID id: String) throws -> Tag? {
-        try fetch(Requests.fetchTag(byID: id)).first
+    func fetchTag(by name: String, context: NSManagedObjectContext? = nil) throws -> Tag? {
+        let context = context ?? backgroundContext
+        let fetchRequest = Requests.fetchTag(byName: name)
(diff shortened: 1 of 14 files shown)


PocketKit/Sources/Sync/Space/Space.swift
```diff
@@ -388,10 +388,13 @@ extension Space {
...
-        let fetchedTag = (try? fetch(fetchRequest, context: context).first) ?? Tag(context: context)
-        guard fetchedTag.name == nil else { return fetchedTag }
-        fetchedTag.name = name
-        return fetchedTag
+        if let fetchedTag = (try? fetch(fetchRequest, context: context).first) {
+            return fetchedTag
+        }
+        let createTag = Tag(context: context)
+        createTag.name = name
+        createTag.remoteID = "remoteID\(name)"
```
cyndichin, line 396: need to discuss how we would like to approach this

cyndichin: possibly UUID, but will wait after touching based with @daniel as we discussed on the call @Gio2018

PR #629, opened 2023-04-19
feat(session): write session data to encrypted store on log in / out

dskuza:
## Summary

- On user log in / out, the session is saved to an encrypted store, which can later be reused by the legacy migration. This is required for re-authenticating a user after org transfer, where we can force re-run the legacy user migration for re-authentication.

- Fixes an issue where analytic events and migration errors were being captured for scenarios that were not true "errors", but special cases that had to be handled.

## References 

IN-1182

## Implementation Details

In order to save session data to a file, `SessionBackupUtility` was used. This listens to both `.userLoggedIn` and `.userLoggedOut` notifications, and appropriately adding / updating the required keys within an encrypted store. Upon requesting file access, this store will be created on-disk if it did not previously exist. Otherwise, this file will be updated as required.

The basic steps for the utility are:
1. Create the store file as necessary
2. Upon session update, decrypt the store
3. Update the required keys (`uid`, `guid`, and `accessToken`) by either inserting, updating, or removing the keys depending on if a user logged in or logged out
4. Encrypt the store, and write to disk

In order to update migrations being incorrectly tracked and logged, two additional cases can be thrown - `noData` (which can possibly happen when Pocket 8 is launched for the first time), and `noSession` (which can happen when a user is logged out). Both of these paths would previously cause an analytic event to be tracked, as well as an error logged, however, these are both valid cases that require custom handling, rather than be treated as other errors.

Additionally, we bail out early of running a migration if we have discovered it has been run, via `UserDefaults`. This helps minimize the number of errors reported, since version comparison is done by first decrypting the store, which is unnecessary if already run.

## Test Steps

This pull request, due to encryption / decryption, is best tested locally with a debugger. 

### Pocket 7 -> 8

- Pocket 7 Logged Out User
  - After updating to Pocket 8, and logging in, the store should contain session data mimicking that contained within the keychain
  - No analytic event should be tracked for failing to decrypt the store due to a missing session.
- Pocket 7 Logged In User
  - After updating to Pocket 8, the user should still be logged in. No store changes are necessary.
  - After signing out within Pocket 8, the store should be updated to remove the session data (no keys will exist)
  - Analytic event should be tracked for the migration succeeding
  
### Pocket 8
- Initially Pocket 8 Logged Out User
  - The store should be created on-disk, but should not contain any session data (empty or otherwise)
- Newly Signed In User
  - The store should be updated with the session data that is within the keychain
- On Sign Out
  - The store should be updated by removing the session data (no keys will exist) 

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/SharedPocketKit/Session/SessionBackupUtility.swift
+++ b/PocketKit/Sources/SharedPocketKit/Session/SessionBackupUtility.swift
@@ -1,0 +1,130 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import Foundation
+import Combine
+import RNCryptor
+
+/**
+ A utility class that can back up a `SharedPocketKit.Session` to an encrypted legacy Pocket store.
+ This is done by adding / modifying existing key/value pairs in the store.
+ - Note: The utility of this class is that we will want to force-run the legacy migration
+ after transferring orgs, and a Pocket 8-only user won't have a pre-existing (required) store.
+ By backing up a session, we can create this store (or update a pre-existing one) as necessary.
+ */
+public class SessionBackupUtility {
+    /// The decryption key used in conjunction with `store`.
+    private let password: String
+
+    private let userDefaults: UserDefaults
+    private let store: LegacyEncryptedStore
+    private let notificationCenter: NotificationCenter
+
+    private let notificationQueue = DispatchQueue(label: "SessionBackupUtility", qos: .utility)
+    private var subscriptions: Set<AnyCancellable> = []
+
+    public init(userDefaults: UserDefaults, store: LegacyEncryptedStore, notificationCenter: NotificationCenter) {
+        self.userDefaults = userDefaults
+        self.store = store
+        self.notificationCenter = notificationCenter
+
+        // Return an existing decryption key, else create + store a new one, and return that
+        if let existing = userDefaults.string(forKey: LegacyUserMigration.decryptionKey) {
+            self.password = existing
+        } else {
+            // Generate a password the same way as the legacy app
+            let password = RNCryptor.randomData(ofLength: 32).base64EncodedString()
+            userDefaults.set(password, forKey: LegacyUserMigration.decryptionKey)
+            self.password = password
+        }
+    }
+
+    public func start() {
+        // Utilize NotificationCenter to save a user session on-disk after a user has logged in
+        notificationCenter.publisher(for: .userLoggedIn)
+            .receive(on: notificationQueue)
+            .sink { [weak self] notification in
+                guard let self = self, let session = notification.object as? Session else {
+                    return
+                }
+
+                do {
+                    try self.save(session: session)
+                    Log.breadcrumb(category: "session", level: .info, message: "Successfully backed up logged in session.")
+                } catch {
+                    Log.capture(error: error)
+                }
+            }.store(in: &subscriptions)
+
+        // Utilize NotificationCenter to reset a user session on-disk after a user has logged out
+        notificationCenter.publisher(for: .userLoggedOut)
+            .receive(on: notificationQueue)
+            .sink { [weak self] notification in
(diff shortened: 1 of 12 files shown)


PocketKit/Sources/SharedPocketKit/Session/SessionBackupUtility.swift
```diff
@@ -0,0 +1,113 @@
...
+    public func start() {
+        // Utilize NotificationCenter to save a user session on-disk after a user has logged in
+        notificationCenter.publisher(for: .userLoggedIn)
+            .receive(on: notificationQueue)
+            .sink { [weak self] notification in
+                guard let session = notification.object as? Session else {
+                    return
+                }
+
+                try? self?.save(session: session)
```
bassrock, line 52: We should capture these errors to sentry.

PocketKit/Sources/SharedPocketKit/Session/SessionBackupUtility.swift
```diff
@@ -0,0 +1,113 @@
...
+            self.password = password
+        }
+    }
+
+    public func start() {
+        // Utilize NotificationCenter to save a user session on-disk after a user has logged in
+        notificationCenter.publisher(for: .userLoggedIn)
+            .receive(on: notificationQueue)
+            .sink { [weak self] notification in
+                guard let session = notification.object as? Session else {
```
bassrock, line 48:
```suggestion
                guard let session = notification.object as? Session, let self else {
```
simplify the self conditional?

PocketKit/Sources/SharedPocketKit/Session/SessionBackupUtility.swift
```diff
@@ -0,0 +1,113 @@
...
+                try? self?.save(session: nil)
+            }.store(in: &subscriptions)
+    }
+
+    /// Saves a `SharedPocketKit.Session` to `store`, adding / modifying existing key/value pairs
+    ///  within the store to contain the requested Session data.
+    private func save(session: Session?) throws {
+        let data: Data
+        do {
+            // `decryptStore` will throw if the on-disk store was
```
bassrock, line 68: Do we need to log or send any of these states to sentry?

PocketKit/Sources/SharedPocketKit/Session/SessionBackupUtility.swift
```diff
@@ -0,0 +1,113 @@
...
+            data = try store.decryptStore(securedBy: password) ?? Data()
+        } catch {
+            data = Data()
+        }
+
+        let initial: [String: Any]
+
+        if data.isEmpty { // data will be empty if the store has been created in Pocket 8
+            initial = [:]
+        } else { // data will not be empty if previously created, either in Pocket 7 or 8
```
bassrock, line 82: Should we add a version field or similar to an entry in the file to be explicit and use that?

bassrock (commented):
Couple of comments. Can you check the tests are not caused by any changes here? Looks like all UI tests failed

PocketKit/Sources/SharedPocketKit/Session/SessionBackupUtility.swift
dskuza: I'm not sure it necessarily matters _which_ version it was created in, within the context of this utility. The file either is there, or isn't, which is what's more crucial. However, we _can_ update the version stored to the version of the application, but not use it.

PocketKit/Sources/SharedPocketKit/Session/SessionBackupUtility.swift
dskuza: We might want to capture what's in the `catch`, even though it's technically a valid code path. Is a breadcrumb with the error enough, here? I don't want to bloat the issues with errors when it could possibly just be because the file didn't exist on-disk (which is legitimate - a fresh install of Pocket won't have it).

bassrock (approved):

dskuza:
@bassrock if you'd like to take _one_ more peek, please, the latest 3 commits are some important changes, but feel free to look over the entire pull request again

bassrock (approved):

PR #637, opened 2023-04-20
feat: add peek n' pop, bulk edit fake door

dskuza:
## Summary

Adds peek n' pop to both Saves and Search, appropriately previewing via web view or the reader. An "Edit" option is included under the preview, which will present a banner saying that editing will come in a later release.

## References 

IN-1247

## Implementation Details

The main change here is the addition of a `BannerPresenter`, which maintains the state of whether an app-wide banner should be presented, and with what data. This type listens to a `.bannerRequested` notification, which when posted, should also include the banner data to present. The pre-existing `.banner` modifier is used to show / dismiss the banner as appropriate, which is presented within `MainView`, as to maintain visibility across all views. When `Edit` is tapped both within the context of Saves and Search, the correct notification is posted with the banner data for the context in which an item was previewed. These strings have also been added to `Localizable.strings`, and regenerated with `swiftgen`.

## Test Steps

- [ ] Saves
  - [ ] Tap and hold on an item in Saves or Archive (both should act the same)
  - [ ] The item should be previewed in a web view if not available for offline viewing, otherwise the reader
  - [ ] When tapping `Edit`, the preview should dismiss, and a banner should be visible stating that editing will come later
  - [ ] When tapping the banner, it should dismiss
- [ ] Search
  - [ ] Tap and hold on an item in Search in any of the 3 scopes
  - [ ] The item should be previewed in a web view if not available for offline viewing, otherwise the reader
  - [ ] When tapping `Edit`, the preview should dismiss, and a banner should be visible stating that editing will come later
  - [ ] When tapping the banner, it should dismiss

## PR Checklist:

- [ ] Added Unit / UI tests
- [x] Basic Self QA

## Screenshots

Simulator Screenshot - iPhone 14 - 2023-04-20 at 11 19 14

Simulator Screenshot - iPhone 14 - 2023-04-20 at 11 19 18

### DIFF
--- a/PocketKit/Sources/PocketKit/BannerPresenter.swift
+++ b/PocketKit/Sources/PocketKit/BannerPresenter.swift
@@ -6,6 +6,7 @@ import Foundation
 import Textile
 import Combine
 import SwiftUI
+import SharedPocketKit
 
 class BannerPresenter: ObservableObject {
     var bannerData: BannerModifier.BannerData?
@@ -19,15 +20,18 @@ class BannerPresenter: ObservableObject {
     }
 
     func listen() {
-        notificationCenter.publisher(for: .bannerRequested).sink { [weak self] notification in
-            guard let data = notification.object as? BannerModifier.BannerData else {
-                // TODO: Log?
-                return
-            }
+        notificationCenter
+            .publisher(for: .bannerRequested)
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] notification in
+                guard let data = notification.object as? BannerModifier.BannerData else {
+                    Log.capture(message: "Requesting banner, but no banner data was supplied")
+                    return
+                }
 
-            self?.bannerData = data
-            self?.shouldPresentBanner = true
-        }.store(in: &subscriptions)
+                self?.bannerData = data
+                self?.shouldPresentBanner = true
+            }.store(in: &subscriptions)
 
         $shouldPresentBanner.filter({ $0 == false }).sink { [weak self] value in
             self?.bannerData = nil
--- a/PocketKit/Sources/PocketKit/BannerPresenter.swift
+++ b/PocketKit/Sources/PocketKit/BannerPresenter.swift
@@ -1,0 +1,36 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import Foundation
+import Textile
+import Combine
+import SwiftUI
+
+class BannerPresenter: ObservableObject {
+    var bannerData: BannerModifier.BannerData?
+    @Published var shouldPresentBanner = false
+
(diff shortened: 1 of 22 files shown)


PocketKit/Sources/PocketKit/BannerPresenter.swift
```diff
@@ -0,0 +1,36 @@
...
+    private var subscriptions: Set<AnyCancellable> = []
+
+    init(notificationCenter: NotificationCenter) {
+        self.notificationCenter = notificationCenter
+    }
+
+    func listen() {
+        notificationCenter.publisher(for: .bannerRequested).sink { [weak self] notification in
+            guard let data = notification.object as? BannerModifier.BannerData else {
+                // TODO: Log?
```
bassrock, line 24: why not just log ;)

PocketKit/Sources/PocketKit/BannerPresenter.swift
```diff
@@ -0,0 +1,36 @@
...
+
+    private let notificationCenter: NotificationCenter
+    private var subscriptions: Set<AnyCancellable> = []
+
+    init(notificationCenter: NotificationCenter) {
+        self.notificationCenter = notificationCenter
+    }
+
+    func listen() {
+        notificationCenter.publisher(for: .bannerRequested).sink { [weak self] notification in
```
bassrock, line 22: Can we add an explicit main thread listener? This could be posted from another one.

PocketKit/Sources/PocketKit/MyList/ItemsList/SavedItemsList/SavedItemsListViewModel.swift
```diff
@@ -202,6 +202,34 @@ class SavedItemsListViewModel: NSObject, ItemsListViewModel {
         source.retryImmediately()
     }
 
+    func preview(for cell: ItemsListCell<NSManagedObjectID>) -> (ReadableViewModel, Bool)? {
+        guard case .item(let itemID) = cell else {
+            return nil
```
bassrock, line 207: Are these errors? Can this ever happen? Should we log?

PocketKit/Sources/PocketKit/BannerPresenter.swift
dskuza: yeah, we should. i think SwiftUI expects things to be published on the main thread, so i think if we receive on the main thread, and set on the main thread, we should be okay.

PocketKit/Sources/PocketKit/BannerPresenter.swift
dskuza: good thing i use TODOs!

bassrock (approved):

PocketKit/Sources/PocketKit/BannerPresenter.swift
bassrock: Yea the main thing i care about is receiving on the main thread

timc-mozilla (approved):

PR #645, opened 2023-04-21
chore(tests): start using apollo built mocks

bassrock:
## Summary

Instead of using raw json responses, lets start setting up basic responses using the types that are available on the Graph, apollo auto generation and inspecting the responses that the client sends. You will see a new autogenerated package canned PocketGraphTestMocks that helps us.

This will ensure that as we test, we generate the correct expected responses using the type safety that our Graph provides us.

Take for example, all our delete/favorite/unfavorite mutations. They all take an ID and return the same id confirming success. So we should inspect the request, get the id and return the same id in the response.

This pattern will also let us to start programmatically build responses, (say generate 30 items in a repeatable expected pattern that conforms to a set of known expectations reducing copy and paste burdens

Note: Everything in Sources/PocketGraphTestMocks was autogenerated by apollo client with fixes added (see comment and linked apollo issue).

## References 
* https://www.apollographql.com/docs/ios/testing/test-mocks/

## Implementation Details
While Apollo provides TestMock classes they were not originally designed to be used in Mock HTTP servers and UI testing, however they are very close to being in the state to do so. I have made some slight modifications (see end of Response+factories.swift) to support this functionality and will reach out to our Account Representative and the GraphQL community to work to mainline some of the feature and functionality we are using and extending.

## PR Checklist:
- [ ] Added Unit / UI tests
- [ ] Basic Self QA


bassrock:
@cyndichin @Gio2018 lets hold off on more UI fixture tests some of the mutations will require dynamic data and I think there can be a better way i want us to Spike. https://www.apollographql.com/docs/ios/testing/test-mocks/

PocketKit/Sources/PocketGraphTestMocks/UnleashAssignment+Mock.graphql.swift
```diff
@@ -0,0 +1,33 @@
...
+
+public class UnleashAssignment: MockObject {
+  public static let objectType: Object = PocketGraph.Objects.UnleashAssignment
+  public static let _mockFields = MockFields()
+  public typealias MockValueCollectionType = Array<Mock<UnleashAssignment>>
+
+  public struct MockFields {
+    @Field<Bool>("assigned") public var assigned
+    @Field<String>("name") public var name
+    @Field<String?>("payload") public var payload
```
bassrock, line 15: Note there is a bug in Apollo GraphQL: https://github.com/apollographql/apollo-ios/issues/2977

bassrock:
There is more to do, but marking PR for review so we can merge and iterate on the pattern.

Tests iOS/Support/RequestMatchers.swift
```diff
@@ -132,6 +136,14 @@ struct ClientAPIRequest {
 extension ClientAPIRequest {
 
     var itemIdVariable: String {
-        self.apolloRequestBody!.variableDict["itemId"] as! String
+        self.apolloRequestBody!.variableDict["itemID"] as! String
```
cyndichin, line 143: itemId?

cyndichin (approved):
looks good high level! one comment on `itemID` vs. `itemId`, would like to walkthrough setting one of these up together during our 1:1 thanks~!

Tests iOS/Support/RequestMatchers.swift
bassrock: Nope, its itemID in the graph when encoded. https://github.com/Pocket/pocket-ios/blob/develop/PocketKit/Sources/PocketGraph/Operations/Mutations/FavoriteItemMutation.graphql.swift#L11

bassrock: actually. i guess this would be in the input dict too, but as id. 🤔 i'll look at this in a followup.

PR #664, opened 2023-04-26
fix(analytics): ensure we always init the api user entity

bassrock:
## Summary

This ensure that we always attach the APIUser entity and if there is a session, the UserEntity to analytics at app launch. This was primarly done in RootViewModel, but RootViewModel is not loaded until after the App finish launching method and with the 7 -> 8 migration, we are now emitting analytics before the RootViewModel loads.

## References 
* https://pocket.slack.com/archives/C4DL4QSU8/p1682463311167069?thread_ts=1682447329.175119&cid=C4DL4QSU8

## Implementation Details
* I think we can do a cleaner job of analytics initing but I am not going to include that work in MVP.

## PR Checklist:
- [ ] Added Unit / UI tests
- [ ] Basic Self QA

### DIFF
--- a/PocketKit/Sources/Analytics/Events/System.swift
+++ b/PocketKit/Sources/Analytics/Events/System.swift
@@ -25,12 +25,23 @@ public struct System: Event, CustomStringConvertible {
         }
     }
 
-    public func toSelfDescribing() -> SelfDescribing {
-        let base = SelfDescribing(schema: System.schema, payload: [
-            "identifier": NSString(string: self.description),
-        ])
+    public var value: String? {
+        switch type {
+        case .userMigration(let userMigrationState):
+            return userMigrationState.id(source)
+        }
+    }
 
-        return base
+    public func toSelfDescribing() -> SelfDescribing {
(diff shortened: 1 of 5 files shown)


PocketKit/Sources/Analytics/Events/System.swift
```diff
@@ -25,12 +25,23 @@ public struct System: Event, CustomStringConvertible {
...
     public func toSelfDescribing() -> SelfDescribing {
-        let base = SelfDescribing(schema: System.schema, payload: [
+        var payload = [
             "identifier": NSString(string: self.description),
-        ])
+        ]
 
-        return base
+        if let value = value {
+            payload["value"] = NSString(string: value)
```
bassrock, line 41: @CMasterson I added this, given we have this field in snowplow we should use it to capture error descriptions if they exist to help inform how we proceed.

CMasterson: Totes, seems like an oversight on my part.

PR #671, opened 2023-04-26
feat(analytics): add tag name(s) as tag event value

dskuza:
## Summary

Adds tag name(s) as tag event values to tag events sent from both the app and the Save extension.

## References 

IN-1355

## Implementation Details

A `tag(Name)` parameter was added to the appropriate static functions to add a `value` to the `Engagement` event. The correct functions were then updated to supply the correct name(s), either by utilizing a string in a calling function, or a `TagType` as appropriate.

## Test Steps

- [ ] When performing tag engagements, the Engagement events sent should include a `value` field.

## PR Checklist:
- [ ] Added Unit / UI tests
- [x] Basic Self QA


timc-mozilla (approved):

Gio2018 (approved):
Looks good to me @dskuza ! side note: it's be nice if we could see those values represented in the console somehow, but I feel this is a broader scope than just tag-related engagement events, so it does not concern this particular PR.

PocketKit/Sources/Analytics/AppEvents/SaveTo.swift
```diff
@@ -79,9 +83,13 @@ public extension Events.SaveTo.Tags {
     }
 
     /// Fired when user taps on a tag name and removes it from the list of input tags in `Add Tags` screen for an item
-    static func remoteInputTag(itemUrl: URL) -> Event {
+    /// - Parameters:
+    ///     - tag: The tag that was removed from the list of input tags.
+    ///     ///     - itemUrl: The url of the item to which a tag was removed.
```
cyndichin, line 88: nit: extra `///` also in `Tags.swift`

cyndichin (approved):
analytics call look good! thanks for picking this up! one teeny nit I noticed.

PR #685, opened 2023-05-01
In 1298 - Migrate from wait to async fulfillment

CMasterson:
## Summary
In async tests we must stop using wait and migrate to async fulfillment.
This will be an error in Swift 6

## References 
https://getpocket.atlassian.net/browse/IN-1298

## Implementation Details
The majority of the call signature remains the same, simply swap
`wait(for:` with `async fulfillment(of:`
e.g.
-        wait(for: [startExpectation], timeout: 10)
+        await fulfillment(of: [startExpectation], timeout: 10)

I also threw in some tidying of other warnings, stuff like unused vars and the like

## Test Steps
Tests should continue to pass


bassrock:
@CMasterson looks like the test build is failing to build

PR #692, opened 2023-05-02
fix(analytics): updating search open analytics to have destination

bassrock:
## Summary
* While digging through some search code noticed a duplicate event being fired in legacy version of analytics.

### DIFF
--- a/PocketKit/Sources/Analytics/AppEvents/Search.swift
+++ b/PocketKit/Sources/Analytics/AppEvents/Search.swift
@@ -289,9 +289,11 @@ public extension Events.Search {
     static func searchCardContentOpen(
         url: URL,
         positionInList: Int,
-        scope: SearchScope
+        scope: SearchScope,
+        destination: ContentOpen.Destination
     ) -> ContentOpen {
         return ContentOpen(
+            destination: destination,
             contentEntity:
                 ContentEntity(url: url),
             uiEntity: UiEntity(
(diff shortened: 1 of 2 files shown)


PocketKit/Sources/Analytics/AppEvents/Search.swift
```diff
@@ -289,9 +289,11 @@ public extension Events.Search {
     static func searchCardContentOpen(
         url: URL,
         positionInList: Int,
-        scope: SearchScope
+        scope: SearchScope,
+        destination: ContentOpen.Destination
     ) -> ContentOpen {
         return ContentOpen(
+            destination: destination,
```
cyndichin, line 296: when should we be including destination in our Content Opens? Looks like it's differentiating web view vs. reader view

bassrock: cc @kirill-demtchouk can you advise?

kirill-demtchouk: `destination` is a required field on the `content_open` event, so we should always include it. Can we set it to `internal` if the item opens in Article view and `external` if the item opens in Web view? Otherwise we can always set it to `internal` given that the mobile apps never hand off the user to an external app when they open content.

bassrock: We can do that.

PR #701, opened 2023-05-03
fix(listen): ensure listen stops playing when closed

bassrock:
## Summary
* Ensure that listen stops playing when the drawer is closed

## References 
* https://github.com/Pocket/pocket-ios-listen/pull/2

## Test Steps
* Play an item in listen
* ensure it stops when you close the drawer.


nzeltzer (approved):
lgtm

PR #706, opened 2023-05-04
fix(share): moving share extension to its old id

bassrock:
## Summary

Because we switched ids on our Share extension iOS saw it as a new extension instead of an upgraded one and changed users settings for how it was displayed.

IN-1429


timc-mozilla (approved):

nzeltzer (approved):
lgtm

PR #714, opened 2023-05-05
feat(launch-screen) [IN-1439]

nzeltzer:
## Summary

Adds launch screen story board

## References 

- Lots of fun in [this Slack conversation](https://pocket.slack.com/archives/C01UH57EJKD/p1683322069310879), but nothing mandatory.

- [SVG Assets in Xcode for Single Scale Images](https://www.avanderlee.com/xcode/svg-image-assets/)

- [Launch screens in Xcode: All the options explained](https://www.avanderlee.com/xcode/launch-screen/)—credit to @bassrock for noticing the explicit mention of SVG; I missed it entirely. 

## Implementation Details

Adds a new Launch Screen storyboard file. You guys like to keep things in Capitalized directories, so I followed that convention. 

You can't use PDFs as vector images for launch screens, but SVGs are fair game: in order to get this to work, Tais exported SVG variants of each of the tab bar icons. Unfortunately, it turned out that the version of the Saves icon she had was slightly different than the one we used in the app. 

She was kind enough to generate a new pair of selected/deselected Saves icons in SVG, and I used them both for the Launch storyboard _and_ the in-app tab bar. Everything lines up perfectly, but now we have one set of tab bar icons in SVG, and all the rest as PDFs. 

**Note**: [we should prolly use SVGs moving forward](https://www.avanderlee.com/xcode/svg-image-assets/): support was added in iOS 13, and they're smaller, more flexible, and have greater compatibility (e.g., launch screens) than PDFs. 

## [Pedanticism](https://en.wiktionary.org/wiki/pedantic#English)

It's not talked about much anymore, but the iOS graphics layer is built on Quartz.

[Quartz 2D](https://en.wikipedia.org/wiki/Quartz_(graphics_layer)) is an Apple-specific [stack-based programming language](https://en.wikipedia.org/wiki/Stack-oriented_programming#:~:text=PostScript%20is%20an%20example%20of,the%20following%20conveyor%20belt%20analogy.) that implements and extends [Postscript](https://en.wikipedia.org/wiki/PostScript) through its inheritance from an early version of the PDF standard. 

[PDF](https://en.wikipedia.org/wiki/PDF), in turn, is Adobe's own implementation of a Postscript superset. 

Statements that Quartz "uses PDF internally" are technically accurate, but also mildly misleading: while Quartz was originally based on PDF Specification 1.4, the two have since evolved into distinct languages, each embodying different core utilities. 

Both Quartz and PDF are primarily used to describe two-dimensional vector graphics, which can be scaled to effectively any resolution without loss of detail. Apple's decision to use Quartz over a more-traditional pixel-buffer rendering technology is probably a major factor in why it was so easily able to support high-resolution screens—Apple's been rendering all of its UI as vectors for nearly 25 years!

It's probably the close familial relationship between Quartz and PDF that prompted Apple to implement PDF support before they did support for SVG—after all, Quartz has long contained APIs to read and render PDF files in a Quartz canvas. That said, while Cocoa's Postscript/PDF compatibility is a genuine [wonder](https://developer.apple.com/library/archive/documentation/GraphicsImaging/Conceptual/drawingwithquartz2d/Introduction/Introduction.html), PDFs really were intended for print proofing, and not screen rendering. [SVG](https://en.wikipedia.org/wiki/SVG), on the other hand, is an open format intended specifically for rendering vector assets to a computer screen (the different core utilities mentioned, _supra_). 

**tl;dr**: all macOS and iOS UIs are _basically_ just giant, interactive PDFs, but SVGs are probably the better choice for vector assets intended for display on a computer screen. 

## Test Steps

If you want to test this yourself, the Simulator is your best friend: set the Simulator debug menu "Slow Animations" to on and launch and re-launch the app to your heart's content.

You can toggle to Dark Mode via Developer settings in the Settings app: the launch storyboard should match our Dark mode appearance settings.  

## PR Checklist:
- [ ] ~Added Unit / UI tests~ _NA_
- [ ] ~Self Review (review, clean up, documentation, run tests)~ _NA_
- [x] Basic Self QA

#### Light Mode 

https://user-images.githubusercontent.com/5227069/236585407-43404b54-86c6-404a-a120-f572a0625e39.mov

#### Dark Mode

https://user-images.githubusercontent.com/5227069/236644731-03aabe0e-c8b5-4ab9-b476-e20bfbcc18b0.mov


cyndichin (approved):
looks great! love the new filename format. Curious to see if we thought about not using the storyboard. Also, thanks for the PR writeup, its amazing 😍 

ref: https://sarunw.com/posts/launch-screen-using-plist/#:~:text=Before%20Xcode%2012%20and%20iOS,you%20configure%20your%20launch%20screen.

Gio2018 (approved):

nzeltzer:
> looks great! love the new filename format. Curious to see if we thought about not using the storyboard. Also, thanks for the PR writeup, its amazing 😍
> 
> ref: https://sarunw.com/posts/launch-screen-using-plist/#:~:text=Before%20Xcode%2012%20and%20iOS,you%20configure%20your%20launch%20screen.

Wow, @cyndichin, I had no idea we could do this in SwiftUI now. I think we'll end up redoing our launch screen in the near future if we move away from the tab bar; we should use SwiftUI when we do that.

PR #716, opened 2023-05-08
feat(debug): add debug menu

cyndichin:
## Summary
PR to add debug menu to iOS app

## References 
N / A 

## Implementation Details
Added basic debug menu to Settings that shows user identifier and sets up UI for snowplow configuration and feature flag values.

## Test Steps
* Add your user ID to `https://featureflags.readitlater.com/projects/default/features/perm.ios.debug.menu` if you would like the debug menu feature flag to be turned on
* Navigate to Settings and view the Debug Menu as shown in screenshot

## PR Checklist:
- N / A Added Unit / UI tests
- [x] Basic Self QA

## Screenshots
| Entry |  Debug Menu  | 
| ------------- |  ------------- |
| Simulator Screenshot - iPhone 13 mini - 2023-05-08 at 15 52 32 | Simulator Screenshot - iPhone 13 mini - 2023-05-08 at 15 49 25|

### DIFF
--- a/PocketKit/Sources/PocketKit/Settings/AccountViewModel.swift
+++ b/PocketKit/Sources/PocketKit/Settings/AccountViewModel.swift
@@ -16,7 +16,7 @@ class AccountViewModel: ObservableObject {
     private let notificationCenter: NotificationCenter
     private let restoreSubscription: () async throws -> Void
     private let networkPathMonitor: NetworkPathMonitor
-    let featureFlags: FeatureFlagServiceProtocol
+    private let featureFlags: FeatureFlagServiceProtocol
 
     // Factories
     private let premiumUpgradeViewModelFactory: PremiumUpgradeViewModelFactory
@@ -105,6 +105,10 @@ class AccountViewModel: ObservableObject {
             self.userDefaults.setValue(!currentValue, forKey: AccountViewModel.ToggleAppBadgeKey)
             self.notificationCenter.post(name: .listUpdated, object: nil)
         }
+    }
+
+    var showDebugMenu: Bool {
+        featureFlags.isAssigned(flag: .debugMenu)
     }
 }
 
--- a/PocketKit/Sources/PocketKit/Settings/SettingsView.swift
+++ b/PocketKit/Sources/PocketKit/Settings/SettingsView.swift
@@ -27,7 +27,7 @@ struct SettingsForm: View {
     var body: some View {
         Form {
             Group {
-                if model.featureFlags.isAssigned(flag: .debugMenu) {
+                if model.showDebugMenu {
                     Section(header: Text("Developer mode").style(.settings.header)) {
                         SettingsRowButton(title: "Debug menu", icon: SFIconModel("ladybug.fill")) {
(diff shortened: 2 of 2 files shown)


PocketKit/Sources/PocketKit/Settings/SettingsView.swift
```diff
@@ -27,6 +27,17 @@ struct SettingsForm: View {
     var body: some View {
         Form {
             Group {
+                if model.featureFlags.isAssigned(flag: .debugMenu) {
```
dskuza, line 30: Perhaps `featureFlags` could be made internal, and the `isAssigned` functionality could be moved into the view model. This would at least eliminate the implementation detail of the feature flags service.

dskuza (approved):

CMasterson (approved):
We _probably_ don't need to add any localization here.

PocketKit/Sources/PocketKit/Settings/DebugMenuView.swift
```diff
@@ -0,0 +1,91 @@
...
+                Text("This is used to change the snowplow endpoint to be used in the application.")
+            }
+
+            Section {
+                ForEach(CurrentFeatureFlags.allCases, id: \.self) { featureFlag in
+                    // TODO: Change to use toggle and force override for feature flags
+                    DebugMenuRow(
+                        title: featureFlag.rawValue,
+                        detail: featureFlag.description,
+                        value: Services.shared.featureFlagService.isAssigned(flag: featureFlag).description
```
cyndichin, line 55: I didn't create a view model to inject these as I felt it was overhead for a debug menu, but happy to refactor if needed

dskuza:
The only other real comment I have is that we're utilizing the `Services.shared` singleton within a view, rather than passing down any dependencies _in_ as a dependency, but this is a debug view and I don't think that's something to think too hard about. Just pointing it out.

bassrock (approved):

cyndichin:
> We _probably_ don't need to add any localization here.

Yup, I didn't add it due to it being the debug menu, thanks for considering this tho!

cyndichin:
> The only other real comment I have is that we're utilizing the `Services.shared` singleton within a view, rather than passing down any dependencies _in_ as a dependency, but this is a debug view and I don't think that's something to think too hard about. Just pointing it out.

good point! Thought it might add overhead so didn't add a view model or inject the dependencies. thanks!

PocketKit/Sources/PocketKit/Settings/AccountViewModel.swift
```diff
@@ -16,6 +16,7 @@ class AccountViewModel: ObservableObject {
     private let notificationCenter: NotificationCenter
     private let restoreSubscription: () async throws -> Void
     private let networkPathMonitor: NetworkPathMonitor
+    let featureFlags: FeatureFlagServiceProtocol
```
dskuza, line 19: @cyndichin can we make this private now?

bassrock:
@dskuza imagine one of many possible futures when we get to SwiftUI we could set all the services as environment variables if we wanted and just yoink them out 😬

PR #718, opened 2023-05-08
fix(error): ensuring we dont capture internal server errors

bassrock:
## Summary
* Do not capture any server errors into Sentry since they are errors on the backend
* Alert the user if they hit a 500 error. @nzeltzer to review copy.

## Implementation Details
* Added to @CMasterson work on 429 errors.

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Self Review (review, clean up, documentation, run tests)
- [ ] Basic Self QA

PR #720, opened 2023-05-08
fix(retry): disabling retry on a 500

bassrock:
## Summary
* A 500 indicates an internal server error, without any backoff logic the code will try the same operation again within seconds of it failing. It is highly unlikely that an internal server error will be solved in that time. 

Until we have a backoff strategy removing the retries on 500 errors

## References 
* https://pocket.slack.com/archives/C01UH57EJKD/p1683583090774729

## Implementation Details
* Making 500 errors be a general failure.
* Pairs with https://github.com/Pocket/pocket-ios/pull/718

### DIFF
--- a/PocketKit/Sources/Sync/Operations/SaveItemOperation.swift
+++ b/PocketKit/Sources/Sync/Operations/SaveItemOperation.swift
@@ -41,7 +41,7 @@ class SaveItemOperation: SyncOperation {
             case ResponseCodeInterceptor.ResponseCodeError.invalidResponseCode(let response, _):
                 switch response?.statusCode {
                 case .some((500...)):
-                    return .retry(error)
+                    return .failure(error)
                 default:
                     return .failure(error)
                 }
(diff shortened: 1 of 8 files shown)


PocketKit/Sources/Sync/Operations/SaveItemOperation.swift
```diff
@@ -41,7 +41,7 @@ class SaveItemOperation: SyncOperation {
             case ResponseCodeInterceptor.ResponseCodeError.invalidResponseCode(let response, _):
                 switch response?.statusCode {
                 case .some((500...)):
```
cyndichin, line 43: Just to confirm we are disabling retry on an errors that are 500 or greater. Do we want to just fail only on 500 instead?

bassrock:
@dskuza done

PocketKit/Sources/Sync/Operations/SaveItemOperation.swift
bassrock: @cyndichin yes that is corrrect.

dskuza (approved):

PocketKit/Sources/Sync/Operations/SaveItemOperation.swift
bassrock: 500+ are general server errors and not on the client https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#server_error_responses

PR #721, opened 2023-05-09
fix(sentry): adding profiling

bassrock:
## Summary

Add in Sentry profiling.

## Implementation Details
We have made this setup a feature flag so that we can enable and disable Sentry profiling at will due to our Sentry contract supporting a limited number of "profile" traces. Note that the value is controlled via the Payload of the feature flag. We will set this up to 0.25 (25%) for traceSampling and 0.50 (50%) for profileSampling

Note that profile sampling is relative to traceSampling:
```
IE. if tracesSampler responds with 100%, profilesSampler will be called 100% of the time,
            // if traces responds with 50%, profileSamples will be called 50% of the time.
```

Note Feature Flags only update values once every 48 hours, currently asking SRE if we need to change this cadence and asking android what they do.


nzeltzer (approved):
lgtm 🤞

bassrock:
Waiting on an answer from Sentry about how to do change options after we initialize the SDK

PR #722, opened 2023-05-09
fix(data): fixing tests missing a piece of data

bassrock:
## Summary
* With https://github.com/Pocket/pocket-ios/pull/717 the mock data stopped loading.

This was a hidden error in Apollo Client complaining about missing fields from a graphql request response.

I propose we should do 1 of 2 things to mitigate in the future:
* Detect if we are in a testing environment and make this type of missing data error crash the app for tests
* Show this error to the user, and make our tests ensure that the error is not shown.


nzeltzer (approved):
lgtm

PR #733, opened 2023-05-11
feat(sentry): add user feedback reporting for errors

cyndichin:
## Summary
Add custom UI for user to report an issue that gets sent to Sentry and populates the user feedback section. This custom report an issue view can be accessed through searches current error views (banner and empty state). 

## References 
IN-1162

## Implementation Details
Create new UI for reporting an issue `ReportIssueView` that utilizes the `ReportField` component. Added a parameter for our `PocketButtonStyle` to consider `small` button types. This is needed to add a small button to our banner view. This is a workaround, but when we refactor buttons we should be using the font size and padding to create over static button height (up to debate). Also, added a `buttonType` to differentiate a `premium` vs. `normal` button to be used in our empty states.

Also, added a new method `sentryCaptureUserFeedback` in our `Log.swift` file to call Sentry API for user feedback.

## Test Steps
1. Set up environment to return error message (whether through modifying code or using Proxyman)
2. Tap on send a report button
3. Fill out report form and verify that email is populated
4. Verify that event is sent to Sentry

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

## Screenshots
| Light Mode | Dark Mode | 
| ---- | ---- | 
| Simulator Screenshot - iPhone 14 - 2023-05-25 at 12 12 18 |  Simulator Screenshot - iPhone 14 - 2023-05-25 at 12 12 08 |

| Banner Mode | Empty State Mode | 
| ---- | ---- | 
| simulator_screenshot_5FF5052C-1544-48F2-B791-F6670F3B772E | simulator_screenshot_90617B28-5373-4D8E-98D5-F1A48C58D1B1 |


PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchView.swift
```diff
@@ -103,18 +106,48 @@ struct SearchEmptyView: View {
     }
 
     var body: some View {
-        if let text = viewModel.buttonText {
-            EmptyStateView(viewModel: viewModel) {
-                GetPocketPremiumButton(text: text)
+        if let buttonType = viewModel.buttonType {
```
cyndichin, line 109: main reason why I decided to create a Button Type is because we were relying on `buttonText` to create a Premium button. However, now we have to consider another button type. Future cases may be needed, but added `normal` for now

Tests iOS/MyList/SearchTests.swift
```diff
@@ -363,6 +365,19 @@ class SearchTests: XCTestCase {
         searchField.typeText("item\n")
 
         XCTAssertTrue(app.saves.searchView.hasBanner(with: "Limited search results"))
+        // TODO: Fix tests after addressing issue with other server error banner
```
cyndichin, line 368: unable to write accurate tests due to this issue
https://pocket.slack.com/archives/C01UH57EJKD/p1685029047272619

PocketKit/Sources/Localization/Resources/en.lproj/Localizable.strings
```diff
@@ -5,6 +5,15 @@
...
 "general.error.serverError" = "Something went wrong with your request. The Pocket team has been notified. Please try again later.";
+"general.error.sendReport" = "Send a report";
+
+// Report an Issue
+"reportIssue.header" = "Report an issue";
+"reportIssue.description" = "Thanks for helping us improve Pocket! Please give us more details about the problem you've experienced.";
+"reportIssue.name" = "Name";
+"reportIssue.email" = "E-mail";
+"reportIssue.comment" = "What happened?";
+"reportIssue.submitIssue" = "Submit issue";
```
CMasterson, line 16: Perhaps?

```suggestion
"reportIssue.submitIssueButton.title" = "Submit issue";
```

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchView.swift
CMasterson: So the naming seems uneven;
.premium is for upselling to premium
.normal is for reporting an issue

It seems like there is a 1-1 connection between the button type and it's purpose, as in I do see a way we could use .normal for anything other than reporting an issue here. So why don't we name it .reportIssue instead?

PocketKit/Sources/Textile/Style/PocketButtonStyle.swift
```diff
@@ -57,73 +66,90 @@ public struct PocketButtonStyle: ButtonStyle {
...
     struct PocketButton: View {
         let configuration: ButtonStyle.Configuration
 
         let variation: Variation
+        let size: Size
 
         @Environment(\.isEnabled)
         private var isEnabled: Bool
 
         var body: some View {
```
CMasterson, line 87: This function is massive and very hard to read.
I'll post some suggestions inline.

PocketKit/Sources/Textile/Style/PocketButtonStyle.swift
```diff
@@ -57,73 +66,90 @@ public struct PocketButtonStyle: ButtonStyle {
...
-            HStack {
-                Spacer()
+            // TODO: Refactor this to match our components in Figma
+            if size == .small && variation == .primary {
                 configuration.label
```
CMasterson, line 90: Consider pulling each block into it's own function.

PocketKit/Sources/Textile/Views/Banner/BottomNotification.swift
```diff
@@ -196,6 +196,27 @@ struct BannerModifier_PreviewProvider: PreviewProvider {
         .previewDisplayName("Action - Dark")
         .preferredColorScheme(.dark)
 
+        VStack {
+            Spacer()
+            Text("Some Screen!")
```
CMasterson, line 201: Some Screen?

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchView.swift
CMasterson: I hadn't seen the emptyView case...
Drop this to a nitpick then, and go with your gut.

CMasterson:
Do we want to do any email validation in this PR?
I'm genuinely unsure as we get the initial value from the model, but it looks as though the user could edit it.
If we do we may have conflicts with the backend's validation rules where we won't allow something that works everywhere else, so we'd need to be on the ball now and if the backend ever updates their rules.

cyndichin:
> Do we want to do any email validation in this PR?
> I'm genuinely unsure as we get the initial value from the model, but it looks as though the user could edit it.
> If we do we may have conflicts with the backend's validation rules where we won't allow something that works everywhere else, so we'd need to be on the ball now and if the backend ever updates their rules.

What backend validation are you referring to? This is to send to Sentry the user's name, email and comment. The worst case is that we won't be able to contact the user. I considered whether we want to even allow the user to input an email, but maybe the user may want to enter a different email in which they respond to this issue. I can bring this up in the ticket and can submit another PR to address this if we would like to proceed with email validation. thanks for the input!

PocketKit/Sources/Textile/Views/Banner/BottomNotification.swift
cyndichin: this is for previews

PocketKit/Sources/Textile/Style/PocketButtonStyle.swift
cyndichin: thanks! I agree, I added a TODO to look into refactoring our Buttons to match our Figma files at a later date

CMasterson:
> > Do we want to do any email validation in this PR?
> > I'm genuinely unsure as we get the initial value from the model, but it looks as though the user could edit it.
> > If we do we may have conflicts with the backend's validation rules where we won't allow something that works everywhere else, so we'd need to be on the ball now and if the backend ever updates their rules.
> 
> What backend validation are you referring to? This is to send to Sentry the user's name, email and comment. The worst case is that we won't be able to contact the user. I considered whether we want to even allow the user to input an email, but maybe the user may want to enter a different email in which they respond to this issue. I can bring this up in the ticket and can submit another PR to address this if we would like to proceed with email validation. thanks for the input!

I had figured the email from the model was coming from the user account. I also assumed backed would be doing some kind of validation on account emails. Those could both be incorrect, of course.

PocketKit/Sources/Textile/Style/PocketButtonStyle.swift
cyndichin: I think we should refactor this entirely instead so didn't want to make too many modifications on existing code

PocketKit/Sources/Textile/Views/Banner/BottomNotification.swift
```diff
@@ -196,6 +196,27 @@ struct BannerModifier_PreviewProvider: PreviewProvider {
...
+            Spacer()
+        }
+        .banner(
+            data: BannerModifier.BannerData(
+                image: .warning, title: "Limited search results",
+                detail: "We're experiencing an error and can't show you full search results. Please try again later.",
```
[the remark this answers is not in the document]
cyndichin: this is for previews, so localization not necessary here.

PocketKit/Sources/Textile/Views/Report/ReportField.swift
```diff
@@ -0,0 +1,41 @@
...
+        static let cornerRadius: CGFloat = 10
+        static let lineWidth: CGFloat = 1
+    }
+    let userInput: Binding<String>
+    let header: String
+    let height: CGFloat
+
+    var body: some View {
+        Section(header: Text(header).style(.recommendation.textStyle).textCase(nil)) {
+            TextField("", text: userInput)
```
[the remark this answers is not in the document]
cyndichin: I can bring this up, but removed it to matched the Figma file

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchView.swift
cyndichin: I incorporated this nitpick, since no preference either way. Let me know if you prefer it!

PocketKit/Sources/Textile/Style/PocketButtonStyle.swift
```diff
@@ -57,73 +66,90 @@ public struct PocketButtonStyle: ButtonStyle {
...
+                    if variation == .internalInfoLink {
+                        SFIcon(SFIconModel("chevron.right", size: CGFloat(Constants.InternalInfoLink.style.fontDescriptor.size)-2, color: configuration.isPressed ? Constants.InternalInfoLink.pressedForegorundColor : Constants.InternalInfoLink.foregroundColor))
+                    }
+                    Spacer()
+                }
+                .multilineTextAlignment(SwiftUI.TextAlignment(Constants.style.paragraph.alignment))
+                .lineSpacing(Constants.style.paragraph.lineSpacing ?? 0)
+                .opacity(isEnabled ? 1.0 : 0.5)
+                .if(variation == .secondary) {  view in
+                   view
```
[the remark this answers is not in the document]
cyndichin: I would like to push back on refactoring this entire function and how we do buttons in a new ticket instead of this PR.

cyndichin:
> > > Do we want to do any email validation in this PR?
> > > I'm genuinely unsure as we get the initial value from the model, but it looks as though the user could edit it.
> > > If we do we may have conflicts with the backend's validation rules where we won't allow something that works everywhere else, so we'd need to be on the ball now and if the backend ever updates their rules.
> > 
> > 
> > What backend validation are you referring to? This is to send to Sentry the user's name, email and comment. The worst case is that we won't be able to contact the user. I considered whether we want to even allow the user to input an email, but maybe the user may want to enter a different email in which they respond to this issue. I can bring this up in the ticket and can submit another PR to address this if we would like to proceed with email validation. thanks for the input!
> 
> I had figured the email from the model was coming from the user account. I also assumed backed would be doing some kind of validation on account emails. Those could both be incorrect, of course.



> > > Do we want to do any email validation in this PR?
> > > I'm genuinely unsure as we get the initial value from the model, but it looks as though the user could edit it.
> > > If we do we may have conflicts with the backend's validation rules where we won't allow something that works everywhere else, so we'd need to be on the ball now and if the backend ever updates their rules.
> > 
> > 
> > What backend validation are you referring to? This is to send to Sentry the user's name, email and comment. The worst case is that we won't be able to contact the user. I considered whether we want to even allow the user to input an email, but maybe the user may want to enter a different email in which they respond to this issue. I can bring this up in the ticket and can submit another PR to address this if we would like to proceed with email validation. thanks for the input!
> 
> I had figured the email from the model was coming from the user account. I also assumed backed would be doing some kind of validation on account emails. Those could both be incorrect, of course.

Yes, we're getting the email from the User entity. However, it can be modified, but sending it to Sentry will just take the value as it is.

cyndichin:
Discussed with @CMasterson, other follow up work will be addressed in separate tickets / follow up PRs:
* Make email unchangeable + add not required as placeholders
* Button Refactoring

PR #736, opened 2023-05-11
feat(tags): set tags action title based on tags count

dskuza:
## Summary

Updates tags actions to say "Add tags" if there are no tags for a given item, otherwise "Edit tags"

## References 

IN-1252

## Implementation Details

- [x] Update Saves to set the correct action title based on tag count
- [x] Update Archive to set the correct action title based on tag count
- [x] Update reader to set the correct action title based on tag count
- [x] Update search to set the correct action title based on tag count
- [x] Update Save extension to set the correct action title based on tag count  

## Test Steps

For the views mentioned under "Implementation Details", check the following:
- [ ] When viewing the tag action for a saved item with no tags (typically via overflow, except the Share extension), the tag item should say "Add tags"
- [ ] After tapping "Add tags" and adding at least one tag, viewing the tag action the next time should say "Edit tags"
- [ ] After tapping "Edit tags" and removing all tags, viewing the tag action the next time should say "Add tags"

Note: the only place one will not see this change after updating tags is in search; the action title matches that of the search item. Since search items currently do not update to show newly update tags (added / removed), the action also does not update as to stay in sync with what's visually present.

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA


PocketKit/Sources/PocketKit/ItemAction.swift
```diff
@@ -79,6 +79,19 @@ extension ItemAction {
...
     }
 
+    static func editTags(_ handler: @escaping (Any?) -> Void) -> ItemAction {
+        return ItemAction(
+            title: Localization.ItemAction.editTags,
+            identifier: .addTagsItem,
+            accessibilityIdentifier: "item-action-add-tags",
+            image: UIImage(asset: .tag),
+            handler: { sender in
+                Haptics.addTagsTap()
```
bassrock, line 89: yay!

bassrock (approved):
lgtm

PR #742, opened 2023-05-15
Pocket open source project

CMasterson:
## Summary
Make Pocket iOS more welcoming to outside contributors.
Add missing secrets examples
Fix a crash that happens with an invalid GroupID

### DIFF
--- a/PocketKit/Sources/Sync/PersistentContainer.swift
+++ b/PocketKit/Sources/Sync/PersistentContainer.swift
@@ -57,10 +57,14 @@ public class PersistentContainer: NSPersistentContainer {
     }
 
     private func getStoreDescription(with storage: Storage, _ groupID: String) -> NSPersistentStoreDescription {
-        if case .shared = storage,
-            let sharedContainerURL = FileManager.default
+        if case .shared = storage {
+            guard let sharedContainerURL = FileManager.default
                 .containerURL(forSecurityApplicationGroupIdentifier: groupID)?
-                .appendingPathComponent("PocketModel.sqlite") {
+                .appendingPathComponent("PocketModel.sqlite") else {
+                Log.warning("Invalid GroupID \(groupID). Falling back to inMemory Storage")
+                break
+            }
+
             Log.debug("Store URL: \(sharedContainerURL)")
             return NSPersistentStoreDescription(url: sharedContainerURL)
         }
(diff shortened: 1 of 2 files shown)


PocketKit/Sources/Sync/PersistentContainer.swift
```diff
@@ -36,21 +36,7 @@ public class PersistentContainer: NSPersistentContainer {
...
-                .containerURL(forSecurityApplicationGroupIdentifier: groupID)!
-                .appendingPathComponent("PocketModel.sqlite")
-
-            Log.debug("Store URL: \(sharedContainerURL)")
-            storeDescription = NSPersistentStoreDescription(url: sharedContainerURL)
-        }
-
-        guard let storeDescription else {
-            fatalError("no store description")
-        }
```
Gio2018, line 53: Not sure if I'm getting this right, but with this change, it seems that there is a path where, if we initialize with the `shared` storage and, for some reason, the on disk db is not available, we' re still initializing the app with a volatile storage. Is this what we want? cc @bassrock

PocketKit/Sources/Sync/PersistentContainer.swift
```diff
@@ -15,7 +15,7 @@ public class PersistentContainer: NSPersistentContainer {
     }()
 
     private (set) var spotlightIndexer: CoreDataSpotlightDelegate?
-    private (set) var storeDescription: NSPersistentStoreDescription?
+    private (set) var storeDescription: NSPersistentStoreDescription!
```
Gio2018, line 18: what's the advantage of implicitly unwrapping here?

Gio2018 (commented):
Had a couple of comments @CMasterson

PocketKit/Sources/Sync/PersistentContainer.swift
CMasterson: `getStoreDescription` returns non-optional.
So I removed the optionalness of `storeDescription` and xcode complained about not all properties being initalised before the class called `super.init`.
`storeDescription` will be valid before it's accessed or something horrifying has happened and so we should probably crash out.

PocketKit/Sources/Sync/PersistentContainer.swift
CMasterson: The scenario I coded for was if we ask for .shared storage with a bad groupID, we fall back to .inMemory. Maybe we need a more explicit `if groupID is the default garbage value { fall back to .inMemory }` but that seems specific and brittle.

Gio2018: moving a bit of conversation to slack on this

CMasterson:
Closing this as I have been approved to add GroupID to the default secrets, making this change unnecessary.

PR #744, opened 2023-05-16
feat(article): open article in web view if it has no article components

cyndichin:
## Summary
Update our logic to consider only showing articles that have article components in reader view. If an article does not have marticle data associated with it, then we show the web view.

## References 
IN-1497

## Implementation Details
Created new boolean property `hasArticleComponents` to determine if an article is not nil or its components are not empty. Added a check for `checkForArticleData` to show item in web view if the article's components are missing. Also, had to modify the `PocketSource` methods `fetchDetails` to return a Bool. This allows us to also check if the remote has article data when it has not been updated locally. 

Note: Although I don't love this, it seems to work. Happy to hear alternatives.

## Test Steps
Currently, no known article to test with, but verified with unit / UI tests. If you have an article in which isArticle is true, but there are no marticle data associated, then perform the following steps.

1. Navigate to the article
2. Tap on the article to open and verify that it opens in web view
3. Verify that this works for syndicated articles, saves and archive articles

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/PocketKit/Article/ReadableViewModel/RecommendationViewModel.swift
+++ b/PocketKit/Sources/PocketKit/Article/ReadableViewModel/RecommendationViewModel.swift
@@ -112,8 +112,8 @@ class RecommendationViewModel: ReadableViewModel {
 
         Task {
             do {
-                try await source.fetchDetails(for: recommendation)
-                checkForArticleData()
+                let remoteHasArticle = try await source.fetchDetails(for: recommendation)
+                checkForArticleData(with: remoteHasArticle)
             } catch {
                 Log.capture(message: "Failed to fetch details for RecommendationViewModel: \(error)")
             }
@@ -121,9 +121,9 @@ class RecommendationViewModel: ReadableViewModel {
     }
 
     /// Check to see if item has article components to display in reader view, else display in web view
-    /// - Parameter item: item that the user wants to open
-    private func checkForArticleData() {
-        if recommendation.item.hasArticleComponents {
+    /// - Parameter remoteHasArticle: condition if the remote in `fetchDetails` has article data
+    private func checkForArticleData(with remoteHasArticle: Bool) {
+        if recommendation.item.hasArticleComponents || remoteHasArticle {
             _events.send(.contentUpdated)
         } else {
             showWebReader()
--- a/PocketKit/Sources/PocketKit/Article/ReadableViewModel/RecommendationViewModel.swift
+++ b/PocketKit/Sources/PocketKit/Article/ReadableViewModel/RecommendationViewModel.swift
@@ -113,10 +113,20 @@ class RecommendationViewModel: ReadableViewModel {
         Task {
             do {
                 try await source.fetchDetails(for: recommendation)
-                _events.send(.contentUpdated)
+                checkForArticleData()
             } catch {
-                Log.capture(message: "Failed to fetch details for recommendation: \(error)")
+                Log.capture(message: "Failed to fetch details for RecommendationViewModel: \(error)")
             }
+        }
+    }
+
+    /// Check to see if item has article components to display in reader view, else display in web view
+    /// - Parameter item: item that the user wants to open
+    private func checkForArticleData() {
+        if recommendation.item.hasArticleComponents {
+            _events.send(.contentUpdated)
+        } else {
+            showWebReader()
         }
     }
 
--- a/PocketKit/Sources/PocketKit/Article/ReadableViewModel/RecommendationViewModel.swift
+++ b/PocketKit/Sources/PocketKit/Article/ReadableViewModel/RecommendationViewModel.swift
@@ -113,7 +113,7 @@ class RecommendationViewModel: ReadableViewModel {
         Task {
             do {
                 let remoteHasArticle = try await source.fetchDetails(for: recommendation)
-                checkForArticleData(with: remoteHasArticle)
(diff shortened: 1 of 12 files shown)


PocketKit/Sources/PocketKit/Item/Item+Extensions.swift
```diff
@@ -40,6 +40,10 @@ public extension Item {
                 // then we should open in web view
                 return true
             } else {
+                // if the item is an article and has no article components (marticle), then open in web view
+                if isArticle && !hasArticleComponents {
```
bassrock, line 44: @cyndichin theres a case (ike archive results) where we purposedly did not download marticle ocmponents, but we download them on demand. Will this ensure that if that download fails then we fall back to webview

cyndichin: ah okay, no wonder tests were breaking for archive and syndicated, I updated to use the logic in our readableviewmodels instead. thanks!

PocketKit/Sources/PocketKit/Article/ReadableViewModel/RecommendationViewModel.swift
```diff
@@ -112,14 +112,24 @@ class RecommendationViewModel: ReadableViewModel {
...
             } catch {
-                Log.capture(message: "Failed to fetch details for recommendation: \(error)")
+                Log.capture(message: "Failed to fetch details for RecommendationViewModel: \(error)")
             }
         }
     }
 
+    /// Check to see if item has article components to display in reader view, else display in web view
+    /// - Parameter remoteHasArticle: condition if the remote in `fetchDetails` has article data
+    private func checkForArticleData(with remoteHasArticle: Bool) {
```
CMasterson, line 125: Should this function be named `displayArticle` or similar?

PocketKit/Sources/PocketKit/Article/ReadableViewModel/SavedItemViewModel.swift
```diff
@@ -123,8 +123,22 @@ class SavedItemViewModel: ReadableViewModel {
         }
 
         Task {
-            try? await self.source.fetchDetails(for: self.item)
+            do {
+                let remoteHasArticle = try await self.source.fetchDetails(for: self.item)
```
CMasterson, line 127: Consider adding a breadcrumb here of what we are trying to fetch.
Then we will have more context if the catch hits.

CMasterson (approved):

PocketKit/Sources/PocketKit/Item/Item+Extensions.swift
```diff
@@ -65,4 +65,9 @@ public extension Item {
     var isImage: Bool {
         hasImage == .isImage
     }
+
+    var hasArticleComponents: Bool {
+        guard let article else { return false }
+        return !article.components.isEmpty
```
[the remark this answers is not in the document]
CMasterson: irrc you don't need a return in a single-line getter. But maybe we should put one just for readability.

PocketKit/Sources/Sync/PocketSource.swift
```diff
@@ -546,26 +546,29 @@ extension PocketSource {
         try? space.filterTags(with: input, excluding: tags)
     }
 
-    public func fetchDetails(for savedItem: SavedItem) async throws {
+    public func fetchDetails(for savedItem: SavedItem) async throws -> Bool {
```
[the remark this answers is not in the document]
cyndichin: agreed, think `RecommendationViewModel` and `SavedItemViewModel` share a lot of similarities. we can consider for the future

PocketKit/Sources/PocketKit/Article/ReadableViewModel/SavedItemViewModel.swift
cyndichin: there's a breadcrumbs added in `fetchDetails`, do we find more value to add another, if so what details would you suggest?

 Log.breadcrumb(category: "recommendations", level: .debug, message: "Loading details for Recomendation: \(String(describing: recommendation.remoteID))")
 
Log.breadcrumb(category: "sync", level: .debug, message: "Fetching details for item with id \(String(describing: savedItem.remoteID))")

PocketKit/Sources/PocketKit/Item/Item+Extensions.swift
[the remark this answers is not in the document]
cyndichin: updated!

PocketKit/Sources/Sync/PocketSource.swift
```diff
@@ -546,26 +546,28 @@ extension PocketSource {
...
+        return try space.performAndWait {
             guard let savedItem = space.backgroundObject(with: savedItem.objectID) as? SavedItem else {
                 Log.capture(message: "Could not retreive item from background context for mutation")
-                return
+                return false
             }
             savedItem.update(from: remoteSavedItem.fragments.savedItemParts, with: space)
             try space.save()
+
+            return remoteSavedItem.item.asItem?.marticle?.isEmpty == false
```
cyndichin, line 570: discussed with roux, curious to see if there was a better way to make it so we can use async / await by waiting for core data to return the status of successfully saving an item, we would not have to return this boolean and make that synchronous instead. It seems like there isn't a way, but opening up for the team in case there is.

PR #745, opened 2023-05-16
Apollo - Always log error code to Sentry

Gio2018:
## Summary
* This PR updates Sentry logging from Apollo operations, making sure to always send the error code to Sentry

## Implementation Details
* Update `ApolloClientProtocol+Extensions`, refactor `handleServerError()` method to always log the code to Sentry.

## Test Steps

## PR Checklist:
- [ ] Added Unit / UI tests
- [ ] Basic Self QA

### DIFF
--- a/PocketKit/Sources/Sync/ApolloClientProtocol+Extensions.swift
+++ b/PocketKit/Sources/Sync/ApolloClientProtocol+Extensions.swift
@@ -79,16 +79,20 @@ public extension ApolloClientProtocol {
         // Codes we wish to notify the user about
         let notifiableErrorCodes = [429, 500, 503]
 
-        guard let responseError = error as? ResponseCodeInterceptor.ResponseCodeError,
-              case .invalidResponseCode(let response, _) = responseError,
-              let code = response?.statusCode else {
-            Log.capture(message: "GraphQl Error - unknown response/status code.", filename: filename, line: line, column: column, funcName: funcName)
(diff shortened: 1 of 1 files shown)


PocketKit/Sources/Sync/ApolloClientProtocol+Extensions.swift
```diff
@@ -72,17 +68,28 @@ public extension ApolloClientProtocol {
...
         // Codes we wish to notify the user about
-        let serverErrorCodes = [429, 500, 503]
+        let notifiableErrorCodes = [429, 500, 503]
 
         guard let responseError = error as? ResponseCodeInterceptor.ResponseCodeError,
               case .invalidResponseCode(let response, _) = responseError,
-              let code = response?.statusCode,
-              serverErrorCodes.contains(code) else { return false }
+              let code = response?.statusCode else {
+            Log.capture(message: "GraphQl Error - unknown response/status code.", filename: filename, line: line, column: column, funcName: funcName)
```
cyndichin, line 85: should we capture general info we get from `error` here too while we have it? 

GraphQl Error - unknown response/status code \(error).

cyndichin (approved):
one non blocking comment

PocketKit/Sources/Sync/ApolloClientProtocol+Extensions.swift
Gio2018: Good point @cyndichin ! Actually the existing interceptor carries a bunch of info that we might want to carry over. thanks for spotting this!

PR #748, opened 2023-05-17
feat(sync): update Item URI attributes to String

dskuza:
## Summary

Updates the Item entity's `givenURL` and `resolvedURL` attributes from `URI` to `String`. This required adding a new version of the data model, as well as generating a mapping from the origin model to the new model.

This pull request additionally begins work for percent-encoding URLs as necessary (e.g when being shared, presented in a browser) by creating a function that _can_ percent-encode URLs; however, this function is currently unused (but tested).

## References 

IN-1455

## Implementation Details

- [x] Create a new data model (PocketModel 2)
- [x] Create a mapping from PocketModel -> PocketModel 2 (PocketModel-to-PocketModel-2)
  - [x] Maps to the new value by using the `absoluteString` value of the previous URI
- [x] Set the persistent container to automatically migrate persistent stores
- [x] Update Item-related URL usage (of `givenURL` and `resolvedURL` to use `String`, replacing`URL`
- [x] Update tests accordingly to build / pass

## Test Steps

- [ ] Install the latest `develop` branch, and install this branch over the existing app. The app should not crash.
- [ ] Install the latest `develop` branch, log in, sync your list, and install this branch over the existing app. The app should not crash.
- [ ] Considering the first two pass, a general regression test

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA


PocketKit/Sources/PocketKit/Home/SlateDetailViewModel.swift
```diff
@@ -111,7 +112,8 @@ extension SlateDetailViewModel {
         let item = recommendation.item
         var destination: ContentOpen.Destination = .internal
         if item.shouldOpenInWebView {
-            let url = pocketPremiumURL(item.bestURL, user: user)
+            guard let bestURL = URL(string: item.bestURL) else { return }
```
cyndichin, line 115: should we capture the url string if this does fail and send to Sentry?

PocketKit/Sources/PocketKit/Recommendation/Recommendation+Extenstions.swift
```diff
@@ -11,7 +11,11 @@ extension Recommendation {
     }
 
     var bestDomain: String? {
-        item.syndicatedArticle?.publisherName ?? item.domainMetadata?.name ?? item.domain ?? item.bestURL.host
+        item.syndicatedArticle?.publisherName
+        ?? item.domainMetadata?.name
+        ?? item.domain
+        ?? URL(string: item.bestURL)?.host
+        ?? "" // TODO: What should be the final fallback string?
```
cyndichin, line 18: what about item.bestURL itself?

PocketKit/Tests/SharedPocketKitTests/URLExtensionsTests.swift
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    func test_withPercentEncoding_isValidURL_returnsURL() {
+        let source = "https://example.com/foo%20bar"
+        let url = URL(percentEncoding: source)
+        XCTAssertNotNil(url)
+        XCTAssertEqual(url!.absoluteString, "https://example.com/foo%20bar")
+    }
+
+    func test_withPercentEncoding_isInvalidURL_returnsURL() {
```
cyndichin, line 30:
```suggestion
    func test_withPercentEncoding_isInvalidURL_returnsNil() {
```

cyndichin (approved):
some comments, but none blocking - code looks good!

PocketKit/Sources/PocketKit/Home/SlateDetailViewModel.swift
dskuza: This'll change a bit once we're percent encoding every URL for its usage when used under certain circumstances (e.g presenting a web view), so logging can be added once that logic's in place. Good callout, though!

PocketKit/Sources/PocketKit/Recommendation/Recommendation+Extenstions.swift
dskuza: Sounds reasonable to me! I'll make the change in the SavedItem-related PR

PocketKit/Tests/SharedPocketKitTests/URLExtensionsTests.swift
dskuza: Good catch, I'll make the fix in the SavedItem-related PR

PR #751, opened 2023-05-22
feat(list): disallow selection of archived or pending items

dskuza:
Disallows the selection of archived items that contain no offline data, or are pending. Additionally, automatically reloads the current Saves list when the network status changes to show the latest state of the items.

## References 

IN-1308

## Implementation Details

Adds a helper function, `isItemDisabled`, that returns true if:
- the user is offline, and an archived item has no article data
- the user is offline, and an item is pending
- the user is online, and an item is pending

This is then used to initialize item presenters appropriately for the required state by adding a `isDisabled` argument. This is because determining whether or not an item is disabled required use of network monitoring, which is outside the scope of a `SavedItem`.

Additionally, a network monitor is used to determine when to reload the Saves list as to show the most up-to-date status of items. On monitor change, we send an updated snapshot, reloading the item identifiers representing the visible cells (as to not reload the whole list). This will update the item(s) to their correct coloring.

## Test Steps

- [ ] Save an item that will have no article components (e.g a google search). Archive the item.
- [ ] Enter airplane mode, and view your archive. The saved item should be greyed out, and tapping it should do nothing.
- [ ] Leave airplane mode, and view your archive. The saved item should not be greyed out, and tapping it should open in web view.

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/PocketKit/MyList/ItemsList/ItemsListItemPresenter.swift
+++ b/PocketKit/Sources/PocketKit/MyList/ItemsList/ItemsListItemPresenter.swift
@@ -40,7 +40,7 @@ class ItemsListItemPresenter {
     private let item: ItemsListItem
     private let isDisabled: Bool
 
-    init(item: ItemsListItem, isDisabled: Bool) {
+    init(item: ItemsListItem, isDisabled: Bool = false) {
         self.item = item
         self.isDisabled = isDisabled
     }
--- a/PocketKit/Sources/PocketKit/MyList/ItemsList/ItemsListItemPresenter.swift
+++ b/PocketKit/Sources/PocketKit/MyList/ItemsList/ItemsListItemPresenter.swift
@@ -38,17 +38,19 @@ private extension Style {
 
 class ItemsListItemPresenter {
     private let item: ItemsListItem
+    private let isDisabled: Bool
 
-    init(item: ItemsListItem) {
+    init(item: ItemsListItem, isDisabled: Bool) {
         self.item = item
+        self.isDisabled = isDisabled
     }
 
     var attributedTitle: NSAttributedString {
-        NSAttributedString(string: title, style: item.isPending ? .pendingTitle : .title)
+        NSAttributedString(string: title, style: isDisabled ? .pendingTitle : .title)
     }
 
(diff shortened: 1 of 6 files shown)


PocketKit/Sources/PocketKit/MyList/ItemsList/SavedItemsList/SavedItemsListViewModel.swift
```diff
@@ -826,3 +836,24 @@ extension SavedItemsListViewModel {
...
+        if networkStatus == .unsatisfied {
+            if item.isArchived {
+                return !(item.item?.hasArticleComponents ?? false)
+            } else {
+                return item.isPending
+            }
+        }
+
+        return item.isPending
+    }
```
Gio2018, line 851: maybe we can simplify the conditions ere with something like this, wdyt?

```
guard networkStatus == .unsatisfied, item.isArchived else {
   return item.isPending
}
return !(item.item?.hasArticleComponents ?? false)

```

PocketKit/Sources/PocketKit/MyList/ItemsList/ItemsListItemPresenter.swift
```diff
@@ -38,17 +38,19 @@ private extension Style {
 
 class ItemsListItemPresenter {
     private let item: ItemsListItem
+    private let isDisabled: Bool
 
-    init(item: ItemsListItem) {
+    init(item: ItemsListItem, isDisabled: Bool) {
```
Gio2018, line 43: looks like in the existing calls, `isDisabled` is predominantly set to `false`: maybe we can add a default value?

Gio2018 (approved):
Works as described @dskuza ! Just a couple of optional comments, other than that
:shipit: !

PocketKit/Sources/PocketKit/MyList/ItemsList/ItemsListItemPresenter.swift
dskuza: I think that'd clean up the unit tests a lot, specifically (which is mainly where the additions are).

Gio2018 (approved):

PR #759, opened 2023-05-26
refactor(saveto): improve link extraction when sharing to pocket

dskuza:
## Summary

Improves link extraction when sharing from certain apps that may share a deep-link into their app (like Ivory).

## References 

IN-1469

## Implementation Details

- If the shared item is a URL
  - If the URL is http(s), use the URL
  - If the URL is not, attempt to find the first included link that _is_ http(s)
- If the shared item is a String
  - If the string _is_ a URL, use the string as a URL
  - If the string contains a URL, find the first included link that is http(s)
  - If the string does not contain a URL, do nothing (error) 

Any time a URL is used, percent encoding is first removed. This simplifies what's stored in the local db, as well as what's sent to the backend. We want to ensure as much consistency between platforms as we can, here.

## Test Steps

- [ ] Test from an app like Ivory (or other apps? Twitter?) and share a link to Pocket. The original URL to the post should be saved, and not a percent-encoded deep-link
- [ ] General Share extension regression test

## PR Checklist:

- [x] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/SaveToPocketKit/SavedItem/SavedItemViewModel.swift
+++ b/PocketKit/Sources/SaveToPocketKit/SavedItem/SavedItemViewModel.swift
@@ -150,45 +150,11 @@ extension SavedItemViewModel {
         }
 
         for provider in providers {
-            let plainTextUTI = "public.plain-text"
-            let urlUTI = "public.url"
-
-            if provider.hasItemConformingToTypeIdentifier(plainTextUTI) {
-                guard let string = try? await provider.loadItem(forTypeIdentifier: plainTextUTI, options: nil) as? String,
-                      let url = retrieveURLFromString(with: string) else {
-                    continue
-                }
-
-                return url
-            } else if provider.hasItemConformingToTypeIdentifier(urlUTI) {
-                guard let url = try? await provider.loadItem(forTypeIdentifier: urlUTI, options: nil) as? URL else {
-                    continue
-                }
-
-                return url
-            } else {
-                continue
+            if let url = await URLExtractor.url(from: provider) {
+                return URL(string: url)
             }
         }
 
-        return nil
-    }
-
-    /// Modified from https://www.hackingwithswift.com/example-code/strings/how-to-detect-a-url-in-a-string-using-nsdatadetector
-    /// - Parameter inputString: string input used to search for a URL
-    /// - Returns: URL found within the string
-    private func retrieveURLFromString(with inputString: String) -> URL? {
-        guard let detector = try? NSDataDetector(types: NSTextCheckingResult.CheckingType.link.rawValue) else {
-            Log.capture(message: "Unable to initialize detector")
-            return nil
-        }
-        let matches = detector.matches(in: inputString, options: [], range: NSRange(location: 0, length: inputString.utf16.count))
-
-        for match in matches {
-            guard let range = Range(match.range, in: inputString) else { continue }
-            let string = String(inputString[range])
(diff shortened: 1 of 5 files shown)


PocketKit/Sources/SaveToPocketKit/SavedItem/SavedItemViewModel.swift
```diff
@@ -150,24 +150,8 @@ extension SavedItemViewModel {
         }
 
         for provider in providers {
-            let plainTextUTI = "public.plain-text"
-            let urlUTI = "public.url"
-
-            if provider.hasItemConformingToTypeIdentifier(plainTextUTI) {
-                guard let string = try? await provider.loadItem(forTypeIdentifier: plainTextUTI, options: nil) as? String,
-                      let url = retrieveURLFromString(with: string) else {
```
cyndichin, line 158: should we remove `retrieveURLFromString` since we're not using that implementation?

PocketKit/Sources/SaveToPocketKit/URLExtractor.swift
```diff
@@ -0,0 +1,80 @@
...
+
+        return nil
+    }
+
+    private static func isValidScheme(_ scheme: String) -> Bool {
+        let validSchemes = ["http", "https", "file", "ftp"]
+        return validSchemes.contains(scheme)
+    }
+
+    private static func firstURL(in string: String) -> String? {
```
cyndichin, line 39: do we think it'll be helpful to also add comments / doc?

PocketKit/Sources/SaveToPocketKit/URLExtractor.swift
```diff
@@ -0,0 +1,80 @@
...
+            return firstURL(in: url)
+        } else if itemProvider.hasItemConformingToTypeIdentifier("public.plain-text") { // We're handed a URL String
+            guard let string = try? await itemProvider.loadItem(forTypeIdentifier: "public.plain-text", options: nil) as? String else {
+                Log.capture(message: "Unable to load String from itemProvider")
+                return nil
+            }
+
+            if let firstURL = firstURL(in: string), let url = URL(string: firstURL) {
+                return self.firstURL(in: url)
+            }
```
cyndichin, line 25: a bit confusing to have both the functions name `firstURL`, maybe make more descriptive?

cyndichin (approved):
lgtm! Some small comments for considerations.

PocketKit/Sources/SaveToPocketKit/URLExtractor.swift
dskuza: I personally think its expressive enough if the variable names make sense. For example:

```
let url = URL(string: "https://getpocket.com")!
var firstURL = URLExtractor.firstURL(in: url)

let urlString = "https://getpocket.com"
firstURL = URLExtractor.firstURL(in: urlString)
```

which _to me_ is more syntactically appealing than:

```
let url = URL(string: "https://getpocket.com")!
var firstURL = URLExtractor.firstURLFromURL(url)

let urlString = "https://getpocket.com"
firstURL = URLExtractor.firstURLFromString(urlString)
```

BUT…I did address your above comment and leave some comments.

cyndichin: agreed, I think it was just the `firstURL` for both throwing me off, but the comments definitely helps! thanks for addressing my comments.

PR #766, opened 2023-05-31
refactor(saveto): update URLExtractor logging

dskuza:
## Summary

Updates `Log` usage in `URLExtractor` to switch from `message` to `breadcrumb`. One `message` is newly captured if and only if no URL could be saved at the _end_ of parsing all item providers.

## Implementation Details

Most things that may happen in `URLExtractor` are not "actionable", but the final state of "couldn't save anything" _might_ be. So, if we breadcrumb up until the "couldn't save anything" scenario, we'll only have one general message that may prove fruitful, and all the things that might have led us there.

## PR Checklist:

- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/SaveToPocketKit/SavedItem/SavedItemViewModel.swift
+++ b/PocketKit/Sources/SaveToPocketKit/SavedItem/SavedItemViewModel.swift
@@ -69,8 +69,7 @@ class SavedItemViewModel {
 
         for item in extensionItems {
             guard let url = try? await url(from: item) else {
-                // Breadcrumbs all the way down, but if we could genuinely not parse out a URL, capture that.
-                Log.capture(message: "Could not find URL in extensions items to save")
+                tracker.track(event: Events.SaveTo.unableToSave())
                 infoViewModel = .error
                 break
             }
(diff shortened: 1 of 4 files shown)


PocketKit/Sources/SaveToPocketKit/SavedItem/SavedItemViewModel.swift
```diff
@@ -65,6 +65,7 @@ class SavedItemViewModel {
 
         for item in extensionItems {
             guard let url = try? await url(from: item) else {
+                tracker.track(event: Events.SaveTo.unableToSave())
```
bassrock, line 68: @dskuza should we pass in the data we have?

bassrock (approved):

PocketKit/Sources/SaveToPocketKit/SavedItem/SavedItemViewModel.swift
dskuza: I guess we could throw with some of the data that couldn't be processed, but that seems a little overkill for now. Maybe something to revisit.

PR #768, opened 2023-05-31
In 1482 - App Badge Not Displaying

CMasterson:
## Summary
Changes how we save the AppBadge flag to absolute value rather than just flipping the existing value. For some reason the value coming out of UserDefaults was always true, which would then be toggled to false.
Also add some analytics to appBadge events (toggling the setting and one for rejected permissions.)

## References 
https://getpocket.atlassian.net/browse/IN-1482

## Implementation Details
Passing the actual value of the toggle to UserDefaults rather than inverting the existing value.

## Test Steps
You should now be able to enable and disable the app badge from Pocket Settings.

## PR Checklist:
- [ ] Basic Self QA

## Screenshots
appBadge


PocketKit/Sources/Analytics/Events/System.swift
```diff
@@ -11,17 +11,24 @@ public struct System: Event, CustomStringConvertible {
     public static let schema = "iglu:com.pocket/system_log/jsonschema/1-0-0"
 
     let type: System.SystemLogType
-    let source: MigrationSource
+    let source: MigrationSource!
```
dskuza, line 14: Why is this implicitly unwrapped? If it's only used under _some_ conditions, perhaps it can be marked as optional, or refactored such that it's extracted from here, and maybe used as an associated value in enum cases where it's needed.

TL;DR: I'm scared of implicitly-unwrapped optionals, especially if we're saying `nil` is a valid case with one of the initializers.

PocketKit/Sources/PocketKit/Settings/AccountViewModel.swift
```diff
@@ -98,19 +98,30 @@ class AccountViewModel: ObservableObject {
...
                     self?.appBadgeToggle = false
                 }
+
                 return
             }
 
-            let currentValue = self.userDefaults.bool(forKey: AccountViewModel.ToggleAppBadgeKey)
-            self.userDefaults.setValue(!currentValue, forKey: AccountViewModel.ToggleAppBadgeKey)
+            self.userDefaults.setValue(isEnabled, forKey: AccountViewModel.ToggleAppBadgeKey)
+            self.userDefaults.synchronize()
```
dskuza, line 124: We shouldn't have to call `synchronize` anymore. Per Apple docs: 

> this method is unnecessary and shouldn't be used.

PocketKit/Sources/Analytics/Events/System.swift
CMasterson: This does need a refactor, I figured I'd get away with it here.
I made it implicit for convenience. If we're doing migration stuff we'll init it, if not we don't use it.

PocketKit/Sources/PocketKit/Settings/AccountViewModel.swift
CMasterson: Oh good!
I hadn't seen that and the ticket seemed to be caused by some kind of sync issue in UserDefaults.

PocketKit/Sources/PocketKit/Settings/AccountViewModel.swift
```diff
@@ -98,19 +98,29 @@ class AccountViewModel: ObservableObject {
...
+                if let error {
+                    Log.capture(error: error)
+                }
+
+                tracker.track(event: Events.Settings.appBadgePermissionDenied())
+
                 self.userDefaults.set(false, forKey: AccountViewModel.ToggleAppBadgeKey)
                 DispatchQueue.main.async { [weak self] in
                     self?.appBadgeToggle = false
                 }
```
cyndichin, line 118: curious to understand why we need to set this as opposed to using the userDefaults as the source of truth to update the view. just a thought, not a blocker

cyndichin (approved):
looks good, one comment for curiosity!

PocketKit/Sources/PocketKit/Settings/AccountViewModel.swift
CMasterson: appBadgeToggle is observed by the Toggle UI element.
I can't think of how we might bind a UserDefault value to a ui component.

PR #771, opened 2023-06-02
feat(reportIssue): add UI refinements

cyndichin:
## Summary
Add refinements to our Report an Issue UI and added a feature flag to show / hide the report an issue button for the empty state view and banner view

## References 
IN-1524

## Implementation Details
* Created new `ReportHeader` view and modify `ReportIssueView` to prevent users from editing their email
* Created feature flag `perm.ios.report_issue` that determines whether to show or hide the report button for the error empty state and error banner view

## Test Steps
- [ ] Given that the user reports an issue, when they see their email populate, then they should not be able to edit it.
- [ ] Given that the user reports an issue, when they first encounter the view, then they should see optional next to both name and comment.
- [ ] Given that the feature flag is off, then the user should not be able to see report an issue button for any of the views.

## PR Checklist:
- N / A Added Unit / UI tests
- [x] Basic Self QA

## Screenshots
| iPhone | 
| ------  |
| image |

### DIFF
--- a/PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchViewModel.swift
+++ b/PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchViewModel.swift
@@ -89,14 +89,22 @@ class SearchViewModel: ObservableObject {
             image: .warning,
             title: Localization.Search.limitedResults,
             detail: Localization.Search.Banner.errorMessage,
-            action: BannerModifier.BannerData.BannerAction(
+            action: reportButton(with: featureFlags)
+        )
+        return isOffline ? offlineView : errorView
+    }
+
+    /// Handles whether to show a report button for a banner
+    private func reportButton(with featureFlagsService: FeatureFlagServiceProtocol) -> BannerAction? {
+        if featureFlagsService.isAssigned(flag: .reportIssue) {
+            return BannerAction(
                 text: Localization.General.Error.sendReport,
                 style: PocketButtonStyle(.primary, .small)
             ) {
                 self.isPresentingReportIssue.toggle()
             }
-        )
-        return isOffline ? offlineView : errorView
+        }
+        return nil
     }
 
     var defaultState: SearchViewState {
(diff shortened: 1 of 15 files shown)


Tests iOS/MyList/SearchTests.swift
```diff
@@ -329,8 +333,8 @@ class SearchTests: XCTestCase {
     }
 
     // MARK: - Search Error State
-    func test_search_showsErrorView_andReportView() {
-        server.routes.post("/graphql") { request, eventLoop -> FutureResponse in
+    func test_search_showsErrorView() {
```
cyndichin, line 336: As per slack conversation, reverted test to not check for the report an issue feature. An investigation ticket will be created to follow up.
https://pocket.slack.com/archives/C01UH57EJKD/p1686592242193829?thread_ts=1686588403.349329&cid=C01UH57EJKD

CMasterson (approved):

Tests iOS/MyList/SearchTests.swift
Gio2018: 👍

Gio2018 (approved):
Great work @cyndichin ! Left a few comments here and there, other than that :shipit: !!

PocketKit/Sources/PocketKit/MyList/SearchItemsList/SearchViewModel.swift
```diff
@@ -88,14 +89,22 @@ class SearchViewModel: ObservableObject {
...
             title: Localization.Search.limitedResults,
             detail: Localization.Search.Banner.errorMessage,
-            action: BannerModifier.BannerData.BannerAction(
+            action: reportButton(with: featureFlags)
+        )
+        return isOffline ? offlineView : errorView
+    }
+
+    /// Handles whether to show a report button for a banner
+    private func reportButton(with featureFlagsService: FeatureFlagServiceProtocol) -> BannerAction? {
```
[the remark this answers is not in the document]
cyndichin: The alternative was to make it so that the `reportButton` function is decoupled from the feature flags implementation. But now that I am looking at it, it might be better to remove it from the signature. @CMasterson we discussed this, any opposition to remove the feature flag service from the method signature since we discussed this previously?

PR #779, opened 2023-06-22
feat(widgets): add initial analytics

cyndichin:
## Summary
Add analytics for widgets
* Track when widget is tapped and opens the application
* Track the current count and details of widget that a user has

## References 
IN-1464

## Implementation Details
Add two analytic events `widgetOpenEngagement` and `widgetDetailsEngagement`

## Test Steps
1. Add a widget and tap on it. Verify that the `widgetOpenEngagement` is triggered.
2. Add multiple widgets and launch the app. Verify that the `widgetDetailsEngagement` is triggered.

## PR Checklist:
- N/A Added Unit / UI tests
- [x] Basic Self QA


PocketKit/Sources/Analytics/AppEvents/Widget.swift
```diff
@@ -0,0 +1,40 @@
...
+        return Engagement(
+            .general,
+            uiEntity: UiEntity(
+                .screen,
+                identifier: "widget-extension.details",
+                componentDetail: String(details.joined(separator: ", ")),
+                value: String(details.count)
+            )
+        )
+    }
```
cyndichin, line 39: not sure if this is the best representation of tracking this type of detail @Gio2018. This also only occurs at launch and not necessarily when a user installs / uninstalls an application. We can use this as a starting point and track the difference if the user's userDefaults value has changed based on whether they have more or less widgets on the next launch.

PocketApp.swift
```diff
@@ -12,6 +12,10 @@ struct PocketApp: App {
     var body: some Scene {
         WindowGroup {
             RootView(model: RootViewModel())
+                .onOpenURL { url in
+                    guard url.scheme == "widget-deeplink" else { return }
```
Gio2018, line 16: this works, but we probably want to leave `PocketApp` as agnostic as possible. An alternative could be to just call a more generic method in the appDelegate (for example `handle(url: URL)` and then let this one handle the logic? This way, we can handle different types of urls and the main does not need to know anything about it

PocketKit/Sources/Sync/Widgets/WidgetsSessionService.swift
```diff
@@ -9,6 +9,7 @@ import WidgetKit
 public protocol WidgetsSessionService {
     var isLoggedIn: Bool { get }
     func setLoggedIn(_ isLoggedIn: Bool)
+    var widgetsDetails: [String] { get }
```
Gio2018, line 12: do we need to store `widgetDetails` into `UserDefaults`, or can we just call `getWidgetConfigurations()` when we need it? This way we ensure to always obtain up to date information

Gio2018 (commented):
Great groundwork @cyndichin ! Left a couple of comments

PocketKit/Sources/Sync/Widgets/WidgetsSessionService.swift
cyndichin: we can call it directly, but storing it in the userDefaults will allow us to keep track if a user deletes / adds a widget as a workaround

See comment: https://github.com/Pocket/pocket-ios/pull/779#discussion_r1238745988

Wanted to get your input, but can remove the userDefaults and revisit in the future.

PocketKit/Sources/PocketKit/PocketAppDelegate.swift
```diff
@@ -216,4 +216,8 @@ public class PocketAppDelegate: UIResponder, UIApplicationDelegate {
         )
         Adjust.appDidLaunch(adjustConfig)
     }
+
+    public func trackWidgetOpen() {
```
Gio2018, line 220: this name might not be totally appropriate, since what we are really tracking here is the app opened from the widget. Once we have deep links in place, it will track an article opened in the app from the widget

PocketKit/Sources/PocketKit/Services.swift
```diff
@@ -216,6 +216,8 @@ struct Services {
         recentSavesWidgetUpdateService = RecentSavesWidgetUpdateService(store: UserDefaultsItemWidgetsStore(userDefaults: userDefaults, key: .recentSavesWidget))
         editorsPicksWidgetUpdateService = RecommendationsWidgetUpdateService(store: UserDefaultsItemWidgetsStore(userDefaults: userDefaults, key: .recommendationsWidget))
         widgetsSessionService = UserDefaultsWidgetSessionService(defaults: userDefaults)
+
+        tracker.track(event: Events.Widget.widgetCountEngagement(details: widgetsSessionService.widgetsDetails))
```
Gio2018, line 220: as you already mention, this might not be the most appropriate place to track this event, because it only happens at app launch and has no context on widgets being installed and removed. I wonder if we can add a similar call in `getTimeline`, with the caveat that we need to filter those events. For example, we could store that dictionary somewhere, and then compare it with the dictionary we get by calling `getCurrentConfigurations` and, if there is a difference in widgets installed, we could fire an event of widgets installed/removed?

PocketKit/Sources/PocketKit/PocketAppDelegate.swift
Gio2018: So yes, if we change the call site (see this [comment](https://github.com/Pocket/pocket-ios/pull/779#discussion_r1238938299)) it might be worth storing into `UserDefaults`. Not sure that `sessionService` is the best place to do it. We could add a dedicated `WidgetTracker` object (injected in the `TimelineProvider` possibly`) that takes care of that

PocketKit/Sources/Sync/Widgets/WidgetsSessionService.swift
Gio2018: So yes, if we change the call site (see this [comment](https://github.com/Pocket/pocket-ios/pull/779#discussion_r1238938299)) it might be worth storing into `UserDefaults`. Not sure that `sessionService` is the best place to do it. We could add a dedicated `WidgetTracker` object (injected in the `TimelineProvider` possibly`) that takes care of that

PocketKit/Sources/PocketKit/PocketAppDelegate.swift
cyndichin: sounds good, will rename to `trackWidgetTapped` and we can refactor when we have deep links. Let me know if you prefer a different name! naming is hard 🤣

PocketApp.swift
cyndichin: will update!

PocketKit/Sources/Sync/Widgets/WidgetsSessionService.swift
cyndichin: updated and created a `WidgetTracker`

PocketKit/Sources/PocketKit/Services.swift
cyndichin: I added the call in `getTimeline`, while it seems to work fairly well for installing a widget, it doesn't seem to trigger for uninstalling a widget. However, there will be an engagement call eventually once the code gets triggered again through some other actions.

cyndichin:
Discussed with @Gio2018 that we can hold off on this PR for now. May revisit when priorities are set.

PR #790, opened 2023-07-07
feat(settings): add original view toggle

cyndichin:
## Summary
* Add a settings toggle “Always open original view,” the effect of which is that all articles will always open directly to original view. 

## References 
IN-1575

## Implementation Details
* Add new setting toggle to Settings screen and modify `SettingsRowToggle` to be more flexible
* Added `shouldDisableReader` computed property to the various view models where we check if we should open in reader or web view

## Test Steps

## PR Checklist:
- [ ] Added Unit / UI tests
- [ ] Basic Self QA


PocketKit/Sources/PocketKit/Home/HomeViewModel.swift
```diff
@@ -120,6 +120,10 @@ class HomeViewModel: NSObject {
     private let recentSavesController: NSFetchedResultsController<SavedItem>
     private let recomendationsController: RichFetchedResultsController<Recommendation>
 
+    private var shouldDisableReader: Bool {
+        featureFlags.isAssigned(flag: .disableReader) || userDefaults.bool(forKey: UserDefaults.Key.toggleOriginalView)
+    }
```
Gio2018, line 125: Because this property is repeated a few times and it is in fact a combined feature flag itself, what do you think about moving it to `FeatureFlagService`? One way to do it without disrupting the current pattern could be to add an `override` or similarly named property to `CurrentFeatureFlags` which could allow us to add a local override only for the flags that we want (like `disableReader` in this case).

Gio2018 (commented):
Awesome work @cyndichin ! I left a few comments for you!

PocketKit/Sources/PocketKit/Home/HomeViewModel.swift
cyndichin: I moved this property to `FeatureFlagService` which can be accessed via `featureFlags.shouldDisableReader`. I'm not sure if I follow the local override, but happy to further dive and discussed that.

PocketKit/Sources/PocketKit/Settings/Components/SettingsRowToggle.swift
```diff
@@ -7,21 +7,19 @@ import Textile
...
 
     var body: some View {
         VStack {
-            Toggle(title, isOn: model.$appBadgeToggle)
-                .onChange(of: model.appBadgeToggle) { newValue in
+            Toggle(title, isOn: isOn)
+                .onChange(of: isOn.wrappedValue) { newValue in
                     self.action(newValue)
                 }
         }
```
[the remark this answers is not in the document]
cyndichin: this is great! I prefer this approach as well. I didn't want to refactor too much of the existing app badge work, but think it's a good mentality to make small improvement changes along the way such as this as opposed to following the pattern. appreciate the thoroughness in review! thanks @Gio2018

Gio2018 (approved):
Awesome work @cyndichin !! :shipit:

PR #809, opened 2023-07-25
Native Collections: add Collection Slug to CorpusItem in Unified Home

Gio2018:
## Summary
*This PR adds the `Collection.slug` to a `CorpusItem` fetched from Unified Home

## References 
* Issue number in branch name

## Implementation Details
* Update `unifiedHome.graphql`, add `CollectionSummary` (at this moment, it only contains `slug`)

## Test Steps
* No UI changes at the moment. If you want to test that we are actually pulling the slug, you can set a conditional breakpoint (or a printout) in `Item+remoteMapping.swift`, inside the method `func update(from corpusItem: CorpusSlateParts.Recommendation.CorpusItem, in space: Space)` and check (for collections only) that `corpusItem.target?.asCollection?.slug` is not nil and actually contains the slug that you are expecting.

## PR Checklist:
- [ ] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/PocketGraph/user-defined-operations/unifiedHome.graphql
+++ b/PocketKit/Sources/PocketGraph/user-defined-operations/unifiedHome.graphql
@@ -10,6 +10,10 @@ fragment CorpusItemParts on CorpusItem {
             __typename
             ...SyndicatedArticleParts
         }
+        ... on Collection {
+            __typename
+            ...CollectionSummary
+        }
     }
 }
 
@@ -39,4 +43,8 @@ query HomeSlateLineup($locale: String!) {
             ...CorpusSlateParts
         }
     }
+}
+
+fragment CollectionSummary on Collection {
+    slug
 }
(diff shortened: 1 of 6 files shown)


PocketKit/Sources/PocketGraph/user-defined-operations/unifiedHome.graphql
```diff
@@ -40,3 +44,7 @@ query HomeSlateLineup($locale: String!) {
         }
     }
 }
+
+fragment CollectionSummary on Collection {
+    slug
+}
```
cyndichin, line 50: I like that you broke it into a fragment, shall we apply this `...CollectionSummary` fragment on `ItemParts` as well?

cyndichin (approved):
looks good! one small comment to use this for `ItemParts` as well

PocketKit/Sources/PocketGraph/user-defined-operations/unifiedHome.graphql
Gio2018: unless we need more fields on `ItemParts`? If so, we can also create another fragment with a different name.

PR #811, opened 2023-07-25
Native Collections: open collections in native view from Unified Home

Gio2018:
## Summary
* This PR allows opening collections in the native view from Home

## References 
* Issue number in branch name

## Implementation Details
* Add `collection` readable type
* Add `collectionSlug` field to `Recommendation

## Test Steps
* go to Home
* tap on a collection
* make sure it opens in the native view

## PR Checklist:
- [ ] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/PocketKit/Home/HomeViewController.swift
+++ b/PocketKit/Sources/PocketKit/Home/HomeViewController.swift
@@ -353,6 +353,9 @@ extension HomeViewController {
             present(url: viewModel.premiumURL)
         case .none:
             readerSubscriptions = []
+        case .collection(let viewModel):
+            let controller = CollectionViewController(model: viewModel)
+            navigationController?.pushViewController(controller, animated: true)
         }
     }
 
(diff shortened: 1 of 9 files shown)


cyndichin (commented):
great work, code looks good to me!

Looking at the AC for the ticket, will we be handling the engineering note in another PR?

```
Engineering notes: 
When the Collection is loaded from the Home tab, if the collection is not saved, then tapping into the native Collection view should load the Collection data remotely. 
When the collection is loaded from the Home tab, if the collection is saved, then tapping into the native Collection view should load the Collection data locally.
```

Also, notice that there's an extra gap in native collection for Home, but not for Saves - we address this in a different PR.

<p align=center>



</p>

Gio2018:
Thanks for reviewing @cyndichin !!

> Looking at the AC for the ticket, will we be handling the engineering note in another PR?
> 
> ```
> Engineering notes: 
> When the Collection is loaded from the Home tab, if the collection is not saved, then tapping into the native Collection view should load the Collection data remotely. 
> When the collection is loaded from the Home tab, if the collection is saved, then tapping into the native Collection view should load the Collection data locally.
> ```
> 

correct! Because we are not currently saving collection stories yet, the second point will need to be handled from where the load actually happens.

> Also, notice that there's an extra gap in native collection for Home, but not for Saves - we address this in a different PR.
> 

Good catch: we can open an issue to fix that gap in that view, if it's not there already

cyndichin (approved):

PR #816, opened 2023-07-27
feat(collections): add error / offline screen

cyndichin:
## Summary
Add offline / error screen for native collections

## References 
IN-1596

## Implementation Details
* Added `error` section to collections 
* Use network monitor to observe changes to network via `observeNetworkChanges` and `handleNetworkChange`
* Refactor `ReportIssueButton` to be more flexible and not tied to search
* Created error view by using `ErrorEmptyState(featureFlags: featureFlags, user: user)` for the empty state view

## Test Steps
- [x] Given I have tapped to open a Collection,
and the collection is being loaded remotely,
when the remote download errors,
then I should see an error view replacing the spinner.
- [x] Given I have tapped to open a Collection,
and then collection is being loaded remotely,
and there is no network connection available (i.e offline),
when the screen is presented,
then I should see an error view replacing the spinner.
- [x] Given I have tapped to open a Collection,
and there was no network connection available (i.e offline),
and the error view is visible,
when the network becomes available again,
then the collection should automatically attempt to be downloaded again, and the spinner should appear, replacing the error view.

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

## Screenshots
Simulator Screenshot - iPhone 14 Pro - 2023-07-27 at 13 00 46

### DIFF
--- a/PocketKit/Sources/PocketKit/Collections/CollectionViewController.swift
+++ b/PocketKit/Sources/PocketKit/Collections/CollectionViewController.swift
@@ -92,6 +92,7 @@ class CollectionViewController: UIViewController {
         collectionView.register(cellClass: LoadingCell.self)
         collectionView.register(cellClass: CollectionMetadataCell.self)
         collectionView.register(cellClass: RecommendationCell.self)
+        collectionView.register(cellClass: EmptyStateCollectionViewCell.self)
 
         model.$snapshot.receive(on: DispatchQueue.main).sink { [weak self] snapshot in
             self?.dataSource.apply(snapshot)
@@ -171,6 +172,11 @@ class CollectionViewController: UIViewController {
 }
 
 private extension CollectionViewController {
+    enum Constants {
+        /// Height that centers the error section so that it appears approximately in the middle
+        static let errorSectionHeight: CGFloat = 0.65
+        static let errorSectionInsets = NSDirectionalEdgeInsets(top: 0, leading: 20, bottom: 0, trailing: 20)
+    }
     func section(for index: Int, environment: NSCollectionLayoutEnvironment) -> NSCollectionLayoutSection? {
         let section = self.dataSource.sectionIdentifier(for: index)
         switch section {
@@ -262,6 +268,25 @@ private extension CollectionViewController {
             )
 
             return section
+        case .error:
+            let section = NSCollectionLayoutSection(
+                group: .vertical(
+                    layoutSize: .init(
+                        widthDimension: .fractionalWidth(1),
+                        heightDimension: .fractionalHeight(Constants.errorSectionHeight)
+                    ),
+                    subitems: [
+                        .init(
+                            layoutSize: .init(
+                                widthDimension: .fractionalWidth(1),
+                                heightDimension: .fractionalHeight(1)
+                            )
+                        )
+                    ]
+                )
+            )
+            section.contentInsets = Constants.errorSectionInsets
+            return section
         default:
             return .empty()
         }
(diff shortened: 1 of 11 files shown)


cyndichin:
Wrote unit tests to cover the cases and seems to pass. Need assistance in running on device to verify that this PR meets AC.

cyndichin:
> Wrote unit tests to cover the cases and seems to pass. Need assistance in running on device to verify that this PR meets AC.

@dskuza was able to help me build on my device to verify 🥳

PocketKit/Sources/PocketKit/Collections/CollectionViewModel.swift
```diff
@@ -207,15 +219,46 @@ class CollectionViewModel {
...
+
+    /// Observes network changes and fetch collection when user returns online
+    private func handleNetworkChange(_ status: NWPath.Status) {
+        guard status != currentNetworkStatus else {
+            return
+        }
+
+        if currentNetworkStatus == .unsatisfied, status == .satisfied {
+            self.snapshot = Self.loadingSnapshot()
+            fetch()
```
Gio2018, line 248: Hey @cyndichin , since we are storing the list locally, I am not sure that we should re-fetch the collection simply because users could go temporarily offline, which could cause an overhead that, most likely, will provide the same data that we have locally, what do you think?

PocketKit/Sources/PocketKit/Collections/CollectionViewModel.swift
```diff
@@ -89,14 +98,17 @@ class CollectionViewModel {
...
         Task {
             do {
                 self.collection = try await source.fetchCollection(by: slug)
-                let snapshot = buildSnapshot()
-                guard snapshot.numberOfItems != 0 else { return }
-                self.snapshot = snapshot
+                updateSnapshot()
             } catch {
+                self.snapshot = errorSnapshot()
                 Log.capture(message: "Failed to fetch details for CollectionViewModel: \(error)")
```
Gio2018, line 112: This code will change with the use of Core Data and an `NSFetchedResultController` containing the stories

PocketKit/Sources/PocketKit/Collections/CollectionViewController.swift
```diff
@@ -170,6 +171,10 @@ class CollectionViewController: UIViewController {
 }
 
 private extension CollectionViewController {
+    enum Constants {
+        static let offlineSectionHeight: CGFloat = 0.65
```
Gio2018, line 175: can we add a comment to why we are using this specific value?

Gio2018 (commented):
Hey @cyndichin , code looks generally good!
I left a couple of comments; also, this one has some overlap with the Core Data one that I'm working on: how do you feel about syncing tomorrow at some point before merging both PRs?

PocketKit/Sources/PocketKit/Collections/CollectionViewModel.swift
cyndichin: Oh, I thought we were storing the list locally only when their user opens the collection. In the case here, when the user is offline and never fetches the collection and then resumes online, I was thinking we should try to fetch the collection again. I do expect us to do a check for the local store first and then fetch remotely if not available.

PocketKit/Sources/PocketKit/Collections/CollectionViewController.swift
cyndichin: yes we can!

PR #822, opened 2023-08-02
refactor(feature-flags): check assignment from in-memory cache

dskuza:
## Summary

Fixes an app hang (commonly seen on initial login) by utilizing in-memory caching of feature flags when checking assignments.

## Implementation Details

An app hang was commonly visible when initially signing in (with a large account) and immediately attempting to switch tabs. After pausing execution, it appeared as if Sentry, when requesting sampling rates, was forcing its closure(s) on the main thread, including downstream tasks, such as fetching an object from Core Data. I _think_ what was happening is that Sentry would call a closure on the main thread, with the side effect of fetching an object from a (background) context _from_ the main thread, where the context was being used heavily, and already on another thread.

This pull request fixes the fetching of feature flag (assignments / payloads) by utilizing an in-memory representation of the feature flags stored in Core Data, as feature flags are inserted / updated / deleted. This frees up any callees in search of feature flag details to utilize an in-memory cache of in-memory representations of those feature flags, which can then more safely be retrieved.

## Test Steps

- [ ] Log out from an account
- [ ] Log into an account with lots of items (the test account I used had ~1.5k)
- [ ] Upon login, immediately toggle between tabs, scrolling through content as desired.
- [ ] The app should not hang

## PR Checklist:

- [ ] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/PocketKit/FeatureFlagService.swift
+++ b/PocketKit/Sources/PocketKit/FeatureFlagService.swift
@@ -5,6 +5,7 @@
 import Foundation
 import Analytics
 import Sync
+import CoreData
 
 protocol FeatureFlagServiceProtocol {
     /// Determine if a user is assigned to a test and a variant.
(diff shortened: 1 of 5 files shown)


PocketKit/Sources/PocketKit/FeatureFlagService.swift
```diff
@@ -74,6 +82,32 @@ class FeatureFlagService: FeatureFlagServiceProtocol {
...
+        // Reset and update our feature flags with the latest in Core Data.
+        // Here, "latest" should be after the last refresh of feature flags from the server.
+        // In-memory representations of the fetched feature flags will be cached for O(1) retrieval
+        // as needed. In-memory caching is used since our use of Sentry seems to force-fetching on the main thread
+        // when requesting tracing / profiling rates (since we check by feature flag whether to be included
+        // in sampling. By moving to in-memory-only, we remove the use of Core Data,
+        // where we do not want to be fetching from the main thread, especially while
+        // doing something like the initial sync, where large amounts of data are being added / updated.
+        featureFlags = [:]
+        for f in flags {
```
Gio2018, line 101: `forEach` maybe? You'd need to replace the `guard` with an `if`

Gio2018 (approved):
Looks good to me @dskuza !!! I used a large list account and jumped around home and saves quite a bit and did not observe hangs. Left only one optional comment.

PR #828, opened 2023-08-07
Native Collections - Fix navigation for sub collections

Gio2018:
## Summary
* This PR fixes a bug that caused navigation to stories not working when navigating to a sub collection and going back to the parent collection.
* It also adds purging orphaned collections and stories when refreshing Home

## References 
* branch name
## Implementation Details
* Update `Space`, add purging logic for `Collection` and `CollectionStory`
* Add `SubscriptionsStack`, a convenience type to store sets of Combine subscriptions on a stack, with `push` and `pop` operations, used to pop unused sets to avoid retaining unnecessary references to `CollectionViewModel`
* Update logic to push collection in `HomeViewController` and `SavesContainerViewController` to use `SubscriptionStack`
* Update logic to push collection in `HomeViewController` to fix the navigation issue mentioned in the description (it was caused by removing all collection subscriptions when pushing a new collection, which caused the parent view controller to not receive navigation status updates anymore)

## Test Steps
1. Build/run
2. Tap "Home"
3. Scroll to find a collection with a sub collection (e.g. "5 Tech Habits to Quit for a Happier Life")
4. Open the collection, then found a sub collection
5. Open the sub collection
6. Go back and make sure that you can navigate to other stories (collection or not)
7. Go back
8. Select any item that is not a collection, then go back 
> [!NOTE]
> this is because an instance of `readableType` is assigned in `HomeViewModel`, and will retain a reference to the latest readable type. If you don't select another readable type, you will still see an instance of `CollectionViewModel`. We will need to change it as well, but since it's a `HomeViewModel` refactor that impacts more than just collections, I consider this outside the scope of the current change, and we will do it in a separate PR.
11. From Xcode, open the memory graph and make sure there are no `CollectionViewModel` instances
12. Restart the debugger
13. Save a collection that has sub collections
14. go to saves, select that collection, then a sub collection then go back to the saves list
15. repeat step 8
16. Go to Home
17. Refresh
18. Open Core Data lab (or other tool to check locally saved entities), and make sure that there are no orphaned collections or stories

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA


cyndichin:
ooh looks interesting, would be interested in how you discovered the memory leaks

PocketKit/Sources/SharedPocketKit/SubscriptionsStack.swift
```diff
@@ -0,0 +1,29 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import Combine
+
+/// A stack used to store sets of Combine subscriptions that can be pushed or popped
+/// Useful for recursive calls containing sets of subscriptions.
```
cyndichin, line 8: nice! 🔥

PocketKit/Tests/SharedPocketKitTests/SubscriptionsStackTests.swift
```diff
@@ -0,0 +1,109 @@
...
+        // callCount should be equal to 3
+        XCTAssertEqual(callCount, 3)
+        stack.pop()
+
+        firstTester.updateMessage("whateverMessage")
+        secondTester.updateMessage("whateverMessage")
+
+        // no more updates should be received, callCount should be still equal to 3
+        XCTAssertEqual(callCount, 3)
+    }
```
cyndichin, line 108: do we want to check for the tester expectations in this test similar to what we do in line 56?

cyndichin (approved):
thanks for this fix! appreciate the details in the PR descriptions and was able to validate through the test steps! I did spot an error where archive / saves button seems to be broken which may be due to us checking `isArchived` a bit later, but we can create a separate ticket for that. 


https://github.com/Pocket/pocket-ios/assets/6743397/366847e8-2e00-4fa7-8a36-97064c7c10ba

PocketKit/Tests/SharedPocketKitTests/SubscriptionsStackTests.swift
Gio2018: we could, but it would be redundant since we already check those in the previous test. Unless we have a way to validate that an expectation was not fulfilled, to enforce that the popped subscription did not trigger an update? I am not aware if we can do that

Gio2018:
> I did spot an error where archive / saves button seems to be broken which may be due to us checking `isArchived` a bit later, but we can create a separate ticket for that.

we can look into this for sure!
If I understand the problem correctly, believe the main point is that, if we set that button early, depending on the actual state of the item, the icon could change and users would see that instead of an initial absence of button.
Another option would be to use a ghost/skeleton view until everything has been loaded, and we can certainly do a PR for that.

Gio2018:
Thank you so much for reviewing @cyndichin !!

PocketKit/Tests/SharedPocketKitTests/SubscriptionsStackTests.swift
cyndichin: what about using `expectation.isInverted = true` to validate that it wasn't fulfilled? but I'm okay with this, just a passing thought!

PR #832, opened 2023-08-08
In 1597 Sharing PDF links to Pocket

CMasterson:
## Summary
These changes allow users to save PDFs such as https://arxiv.org/pdf/2306.00739.pdf through the Share extension.

## References 
https://getpocket.atlassian.net/browse/IN-1597

## Implementation Details
Main Change: When parsing strings from the array of ExtensionItems, previous code would return early if we encountered an unparseable entry, but now we continue through the array until it is complete.

Other Changes: I broke the code up into smaller functions with discrete tasks and added a unit test to ensure the code continues to work correctly.

## Test Steps
In Safari/another browser; navigate to https://arxiv.org/pdf/2306.00739.pdf

Share the page to Pocket, using the build-in sharesheet

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA


PocketKit/Sources/SaveToPocketKit/SavedItem/SavedItemViewModel.swift
```diff
@@ -63,18 +63,48 @@ class SavedItemViewModel {
...
-                infoViewModel = .error
-                break
+                continue
             }
 
+            urlsToSave.append(url)
+        }
+
+        return urlsToSave
+    }
```
dskuza, line 104: If I'm interpreting the code correctly, we don't want to save _all_ URLs that are found; at least, that's not what the legacy app did, and not what the original objective of this share extension was / is. We want to find the _first_ URL that we can save, and save that.

CMasterson: We only seem to pull one URL out, if we do pick one would you prefer we .first or .last?

dskuza: What the original code was _supposed_ to do, which the `break`s broke, was for each extension item, grab its url, if it could, and attempt to save the first URL that was validly pulled from an extension item. 

So, for the PDF-situation specifically, we had two extension items -- one whose UTI was `com.adobe.pdf`, and one whose UTI was `public.url`.  The `for` loop would attempt to extract a URL from the first extension item, and fail since its UTI wasn't correct, then move onto the second extension item, from which the URL _would_ be extracted, and we would then save that.

In a hypothetical world where there was a _third_ extension item, the fact that we pulled a URL to save from the _second_ extension item would stop the `for` loop, and the third extension item, no matter the URL, would not attempt to be saved.

TL;DR - Stop and save the _first_ extracted URL from an extension item, and only that URL.

PR #833, opened 2023-08-09
Add new color for Darkmode URLs in Reader

CMasterson:
## Summary
The standard blue highlight was difficult to read against the dark background of darkmode. Added a more legible color for URLs

## References 
https://getpocket.atlassian.net/browse/IN-1608

## Implementation Details
Added the new color to the standard Asset catalog. Adding to the Textile catalog added the need to import Textile for just one color on one line and seemed excessive. 

## Test Steps
Open and item that is  displayed in Reader.
Check for URLs, all should be the new teal color in darkmode and the standard blue in light mode.

## PR Checklist:
- [x] Basic Self QA

## Screenshots
Screenshot 2023-08-09 at 11 58 12 AM
Screenshot 2023-08-09 at 11 58 07 AM

### DIFF
--- a/PocketKit/Sources/PocketKit/Article/Cells/ArticleComponentTextCell.swift
+++ b/PocketKit/Sources/PocketKit/Article/Cells/ArticleComponentTextCell.swift
@@ -3,6 +3,7 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import UIKit
+import Textile
 
 // An object that conforms to this protocol is commonly capable of responding to
 // (overridden) events that occur within a PocketTextView nested within a PocketTextCell.
@@ -60,7 +61,7 @@ class ArticleComponentTextView: UITextView {
         isScrollEnabled = false
         delegate = self
 
-        linkTextAttributes = [.foregroundColor: UIColor(named: "textfieldURL")]
+        linkTextAttributes = [.foregroundColor: ColorAsset.ui.textfieldURL as Any]
 
         interactions
             .filter { $0 is UIContextMenuInteraction }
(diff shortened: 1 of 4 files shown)


PocketKit/Sources/PocketKit/Article/Cells/ArticleComponentTextCell.swift
```diff
@@ -60,6 +60,8 @@ class ArticleComponentTextView: UITextView {
         isScrollEnabled = false
         delegate = self
 
+        linkTextAttributes = [.foregroundColor: UIColor(named: "textfieldURL")]
```
cyndichin, line 63: nice! the color looks so much better in dark mode! should we consolidate this hard coded string in the `Palettes` file to consolidate the asset strings?

CMasterson: Should be better now!

PocketKit/Sources/Textile/Style/Colors/Palettes.swift
```diff
@@ -23,6 +23,7 @@ public struct UIPalette {
 
     public let homeCellBackground = ColorAsset.ui("homeCellBackground")
     public let saveButtonText = ColorAsset.ui("saveButtonText")
+    public let textfieldURL = UIColor(named: "textfieldURL")
```
cyndichin, line 26: If we add the color to `Colors.xcassets` in `Textile` and then use `ColorAsset.ui("textfieldURL")`, do we still see the crash? Or only when we call it in line `64` above?

CMasterson: That function is currently `internal`, changing it to `public` for this test.

That still crashes the same way.

cyndichin: hmm I tried to reproduce the crash using those steps in the previous comment, but wasn't able to.

cyndichin (approved):
looks so good 🥳 , thanks for troubleshooting this with me! 
Simulator Screenshot - iPhone 14 Pro - 2023-08-09 at 15 47 45

CMasterson:
>thanks for troubleshooting this with me!

Wait, You were troubleshooting this with ME!

PR #834, opened 2023-08-09
fix(tags): update tags logic in shared extension

cyndichin:
## Summary
Update logic for tags in shared extension:
* Fix recent tags that are not being updated and fix order to be most recently used
* Sort tags alphabetically unless a recent tag

## References 
IN-1569, IN-1598

## Implementation Details
* `itemTagNames` was being updated and not providing to correct value so created a clearer variable `originalTagNames` that gets set on the object's init method instead of being a computed property; therefore provided the wrong input values for the recent tag provider
* Notice that recent tags were being sorted alphabetically instead of being sorted by most recently used
* Update `other tags` to be sorted in alphabetical order in `SaveToAddTagsViewModel.swift`

## Test Steps

**Fix recent tags that are not being updated**
- [ ] Given that I add a tag to a saved item in the app,
when I save another item via the share extension, 
and navigate to the Add Tags screen,
then I should see that the recent tags updated with that tag. 

**Sort tags alphabetically unless a recent tag**
- [ ] Given that I save an item via the shared extension,
and I navigate to the Add Tags screen, 
then I should see that the tags are sorted alphabetically in the `My tags` section.

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

## Screenshots
Simulator Screenshot - iPhone 14 Pro - 2023-08-10 at 09 37 58

### DIFF
--- a/PocketKit/Sources/PocketKit/Tags/PocketAddTagsViewModel.swift
+++ b/PocketKit/Sources/PocketKit/Tags/PocketAddTagsViewModel.swift
@@ -35,7 +35,7 @@ class PocketAddTagsViewModel: AddTagsViewModel {
     /// Fetches recent tags to display to the user only if premium and user has more than 3 tags
     var recentTags: [TagType] {
         guard user.status == .premium && fetchAllTags.count > 3 else { return [] }
-        return recentTagsFactory.recentTags.sorted().compactMap { TagType.recent($0) }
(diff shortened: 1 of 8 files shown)


Gio2018 (approved):
Looks good to me @cyndichin !

dskuza (approved):

PR #839, opened 2023-08-11
Widget Deeplinks Setup

CMasterson:
## Summary
Adds links to Widget items and allows Pocket to receive deep links

## References 
https://getpocket.atlassian.net/browse/IN-1617

## Implementation Details
Registered `pocket` as a URL scheme
Added .widgetURL property to ItemWidgetRow

## Test Steps
No applicable test steps yet

## PR Checklist:
- [ ] Added Unit / UI tests
- [ ] Basic Self QA

### DIFF
--- a/Info.plist
+++ b/Info.plist
@@ -50,18 +50,6 @@
 				<string>pocket-next</string>
 			</array>
 		</dict>
-		<dict>
-			<key>CFBundleTypeRole</key>
-			<string>Editor</string>
-			<key>CFBundleURLIconFile</key>
-			<string>Icon</string>
-			<key>CFBundleURLName</key>
-			<string></string>
-			<key>CFBundleURLSchemes</key>
-			<array>
-				<string>pocket</string>
-			</array>
-		</dict>
 	</array>
 	<key>CFBundleVersion</key>
 	<string>$(CURRENT_PROJECT_VERSION)</string>
--- a/Info.plist
+++ b/Info.plist
@@ -50,6 +50,18 @@
 				<string>pocket-next</string>
 			</array>
 		</dict>
+		<dict>
+			<key>CFBundleTypeRole</key>
+			<string>Editor</string>
+			<key>CFBundleURLIconFile</key>
+			<string>Icon</string>
+			<key>CFBundleURLName</key>
+			<string></string>
+			<key>CFBundleURLSchemes</key>
+			<array>
+				<string>pocket</string>
+			</array>
+		</dict>
 	</array>
 	<key>CFBundleVersion</key>
 	<string>$(CURRENT_PROJECT_VERSION)</string>
--- a/PocketApp.swift
+++ b/PocketApp.swift
@@ -12,6 +12,21 @@ struct PocketApp: App {
     var body: some Scene {
         WindowGroup {
             RootView(model: RootViewModel())
+                .onOpenURL { url in
+                    if url.scheme == "pocket" {
+                        let queryItems = URLComponents(url: url, resolvingAgainstBaseURL: false)?.queryItems
+                        let urlItem = queryItems?.filter {
+                            $0.name == "url"
+                        }
+
+                        if url.host() == "home" {
+                            // TODO: Open Item
+                        } else {
+                            // TODO: Log invalid deeplink
+                            return
+                        }
+                    }
+                }
         }
     }
 }
--- a/PocketApp.swift
+++ b/PocketApp.swift
@@ -12,21 +12,6 @@ struct PocketApp: App {
     var body: some Scene {
         WindowGroup {
             RootView(model: RootViewModel())
-                .onOpenURL { url in
-                    if url.scheme == "pocket" {
-                        let queryItems = URLComponents(url: url, resolvingAgainstBaseURL: false)?.queryItems
-                        let urlItem = queryItems?.filter {
-                            $0.name == "url"
-                        }
-
-                        if url.host() == "home" {
-                            // TODO: Open Item
-                        } else {
-                            // TODO: Log invalid deeplink
-                            return
-                        }
-                    }
-                }
         }
     }
 }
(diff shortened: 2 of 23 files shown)


PocketKit/Sources/ItemWidgetsKit/Views/ItemWidgetsRow.swift
```diff
@@ -37,6 +38,19 @@ struct ItemWidgetsRow: View {
                 ItemThumbnail(image: image)
             }
         }
+        .widgetURL(deeplinkURL(url))
+    }
+
+    /// <#Description#>
+    /// - Parameter url: The URL to add as a parameter of the getpocket URL
+    /// - Returns: Returns a deeplink url e.g. com.getpocket://getpocket.com/app/save?url=[GIVEN URL]
+    func deeplinkURL(_ url: String) -> URL? {
```
CMasterson: Moving this code to a model.

Info.plist
```diff
@@ -50,6 +50,18 @@
 				<string>pocket-next</string>
 			</array>
 		</dict>
+		<dict>
+			<key>CFBundleTypeRole</key>
+			<string>Editor</string>
+			<key>CFBundleURLIconFile</key>
+			<string>Icon</string>
```
Gio2018: do we need an icon here?

PocketApp.swift
```diff
@@ -12,6 +12,21 @@ struct PocketApp: App {
...
+                        }
+
+                        if url.host() == "home" {
+                            // TODO: Open Item
+                        } else {
+                            // TODO: Log invalid deeplink
+                            return
+                        }
+                    }
+                }
```
Gio2018: Hey @CMasterson , we should move out this logic out of the main app; `RootView` should not be concerned at all about how we handle urls. It'd be better to create a dedicated type for this, which gets injected in `RootViewModel` and the view would just pass the value on to the viewModel and, in turn, to the dedicated type, what do you think?

PocketKit/Sources/ItemWidgetsKit/Views/ItemWidgetsRow.swift
Gio2018: This method looks redundant with `pocketURL` in `ItemContent`. Because the view should not be concerned on how we handle urls, I believe it would be preferable to refactor `ItemContent.pocketUrl` and just remove this one. This way, we can just infer `pocketUrl` from the content and this view will be happy as it is.

Gio2018 (changes requested):
Hey @CMasterson , had a couple of architecture changes requests and one comment. Thank you!

PocketApp.swift
CMasterson: I think those are good suggestions, but maybe outside the scope of this ticket? I just put that code in there as an POC of catching and parsing the deeplink URL.

PocketKit/Sources/ItemWidgetsKit/Views/ItemWidgetsRow.swift
CMasterson: I have already moved that code to the ItemRowContent model, I shied away from the pocketURL in ItemContent as it specified 
`/// The URL used to open the saved item within the Pocket reader.`
So I figured there may be a chance of functionality collisions since that specified Reader.
If that's not the case then we could combine functionality, that would take a minor effort to update the example parsing code above.

Let me know!

Info.plist
CMasterson: I'm not sure what the icon is used for.

PocketKit/Sources/ItemWidgetsKit/Views/ItemWidgetsRow.swift
Gio2018: You can totally use that one

PocketApp.swift
Gio2018: Well, if we are setting the foundation to open an URL and we are merging into `develop`, then I'd say this is part of the scope of this ticket, otherwise we'd need to refactor our codebase afterwards. Happy to chat live on how to architect this.

CMasterson: The ticket is about adding the URLs to the widget items, so it kinda makes sense to just remove the handling code altogether and do that in the followup tickets.

PocketKit/Sources/ItemWidgetsKit/Views/ItemWidgetsRow.swift
CMasterson: I gave it a little tweaking and switched over to using it.

Info.plist
cyndichin: just curious - do you have any resource guide to point me to learn more about setting this up? Why are we adding this to the `.plist`. Is this to support URL Scheme? 

I was researching since I'm not too familiar with URL scheme, but it seems it was used for deep linking in previous iOS versions. I read this [blog post](https://byby.dev/ios-deep-linking) - it seems that universal links has a couple of benefits over custom URL schemes. 

Does this also support universal links?

CMasterson: Adding a URL Type tells the app what URL Schemes it can handle.

In this case we are registering `pocket://` as a scheme that we want to handle. 
My understanding is that when this app is installed iOS will be told that if it recieves a `pocket://something/something` URL iOS knows to power up Pocket and hand off that URL to Pocket for handling. We parse out the intent and navigate to the related content in code.

It's been a while, but iirc Universal Links don't use this. Universal Links are standard `https://` links. You set up an endpoint on your server to tell Apple that this URL can also be opened by your app.

https://facebook.com/apple-app-site-association
Is Facebook's Universal Link association file, for example. That means that you can try to open `https://www.facebook.com/someContent` on either the Facebook app, if installed, or just in a browser if not.

dskuza (approved):
I think this, along with #843, work together, so approved with the understanding that both will be merged in at some point, since the work is related

Info.plist
Gio2018: It seems that we don't need this at all, the entitlement is already in there

CMasterson: Removed

Gio2018 (approved):

PR #840, opened 2023-08-11
feat(saveto): log out from share extension on 401

dskuza:
## Summary

Logs out of the share extension when receiving a 401 from a GraphQL request. Logging out from the share extension will also log a user out from the app.

There is a tiny quirk where, since the save is async, we first will show that a URL could (not) be saved, followed by updating the UI _after_ the request is made, however long that takes.

## References 

IN-1615

## PR Checklist:
- [ ] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/SaveToPocketKit/SaveToUserManagementService.swift
+++ b/PocketKit/Sources/SaveToPocketKit/SaveToUserManagementService.swift
@@ -1,0 +1,50 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import Foundation
+import SharedPocketKit
+import Sync
+import Combine
+
+protocol SaveToUserManagementServiceProtocol {
+    func logout()
+}
+
+final class SaveToUserManagementService: ObservableObject, SaveToUserManagementServiceProtocol {
+    let appSession: AppSession
+    let user: User
+    let notificationCenter: NotificationCenter
+
+    // Deletion state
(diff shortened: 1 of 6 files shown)


PocketKit/Sources/SaveToPocketKit/SaveToUserManagementService.swift
```diff
@@ -0,0 +1,50 @@
...
+                guard let self = self else { return }
+                Task {
+                    await self.logout()
+                }
+            }
+            .store(in: &subscriptions)
+    }
+
+    /// Logs out the user
+    /// We do this on the main thread because the Notification Center will post to areas that will change the UI
```
Gio2018, line 42: Question: if we don't strictly need this to run on the main thread, would `receive(on: DispatchQueue.main)` at the receiver be enough?

PocketKit/Sources/SaveToPocketKit/Services.swift
```diff
@@ -19,6 +19,11 @@ struct Services {
     let notificationCenter: NotificationCenter
     let braze: SaveToBraze
 
+    /// The user management service that will log a user out when the correct notification is posted.
+    /// This is marked as private since we do not need to access it, but we need it to be around for the lifetime
+    /// of a Services instance. When using `.shared`, this lifetime will exist for the duration of the share extension.
+    private let userManagementService: SaveToUserManagementServiceProtocol
```
Gio2018, line 25: nit: should we rename to `logoutService` or similar? Seems that's all it does

Gio2018 (approved):
Looks great @dskuza ! Just an optional comment and a question, other than that
:shipit: !

PR #841, opened 2023-08-11
feat: cards display in iPad

cyndichin:
## Summary
Update card display for iPad view for both our slate detail (see all) sections and collection stories section. 

Currently the pages have a single column layout for cards, which is okay for smaller screens, but we want to improve the experience when viewing these pages on big surfaces.

## References 
IN-1577

## Implementation Details
Modified the section layout code for `CollectionViewController` and `SlateDetailViewController`. The code between the two are fairly similar besides the fact that they take in two different core data models, `CollectionStory` and `Recommendation` respectively. 

Steps to determine what layout to show:
1. Check whether device is using wide layout (based on device being iPad and has horizontal size class of regular)
2. Display either `sectionForWideLayout` or `sectionForCompact` depending on requirement in 1. Note: `sectionForCompact` was mainly a copy and paste of existing code into its own function.
3. If we want a wide layout that has a 3 column grid layout, there are 2 requirements: the view's width length must be large enough and the device must be in landscape view; otherwise, the wide layout uses a 2 column grid layout.
4. We create the section based on list of items and height of the cells. For the height, we need to calculate the row heights by retrieving the max height of the items in a row using `getMaxHeightForRow`.

Update: After some conversations, refactored the code to live in a class called `GridSectionLayout`. Logic remains the same, but organizing code and renaming differed.

## Test Steps
Navigate to Home and tap on "See All" for the different sections and view the recommendations layout. Verify the cards display experience in various devices and modes (i.e. iphone mini, iPhone plus, iPad Pro, iPad mini, Split View, landscape, portrait). 

Navigate to a Collection whether in Home, Saves or Search and view the stories layout. Verify the cards display experience in various devices and modes (i.e. iphone mini, iPhone plus, iPad Pro, iPad mini, Split View, landscape, portrait). 

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

## Screenshots
| iPad Landscape | iPad Portrait |
| --- | --- |
| image | image |

### DIFF
--- a/PocketKit/Sources/PocketKit/Collections/CollectionViewController.swift
+++ b/PocketKit/Sources/PocketKit/Collections/CollectionViewController.swift
@@ -65,6 +65,11 @@ class CollectionViewController: UIViewController {
 
     private var metadata: CollectionMetadata?
     private var subscriptions: [AnyCancellable] = []
+
+    private var showThreeColumnLayout: Bool {
+        return self.view.bounds.width >= Constants.minWidthBoundaryForFullColumnLayout && !UIDevice.current.orientation.isPortrait
+    }
+
     private let model: CollectionViewModel
 
     init(
@@ -290,30 +295,30 @@ private extension CollectionViewController {
     }
 
     /// Determines the stories layout in the native collection view on iPad mode with regular horizontal size class (including split view). Number of columns for the grid layout depends on device orientation and the view's width length. If view's width is less than 800, show 2 col, otherwise show 2 col if it is in portrait and 3 col if it is in landscape mode.
-    private func collectionSectionForWideLayout(with stories: [CollectionStory], width: CGFloat, margin: CGFloat) -> NSCollectionLayoutSection {
-        if self.view.bounds.width >= Constants.minWidthBoundaryForFullColumnLayout && !UIDevice.current.orientation.isPortrait {
-            return collectionSectionForiPadLayout(with: stories, width: width, margin: margin, numberOfColumns: 3)
-        } else {
-            return collectionSectionForiPadLayout(with: stories, width: width, margin: margin, numberOfColumns: 2)
-        }
-    }
-
-    /// Determines the stories layout for the native collection view on iPad mode with regular horizontal size class
     /// - Parameters:
     ///   - stories: list of stories associated with collection to display
     ///   - width: width that the section occupies
     ///   - margin: padding adding to the side of the section
     /// - Returns: section layout for stories on iPad view and regular horizontal size class
-    private func collectionSectionForiPadLayout(with stories: [CollectionStory], width: CGFloat, margin: CGFloat, numberOfColumns: CGFloat) -> NSCollectionLayoutSection {
-        let numberOfRows = (CGFloat(stories.count) / numberOfColumns).rounded(.up)
-
-        let recommendationsHeight: [CGFloat] = stories.map { story in
-            return RecommendationCell.fullHeight(viewModel: model.storyViewModel(for: story), availableWidth: width / numberOfColumns - (margin * 2)) + margin
-        }
+    private func collectionSectionForWideLayout(with stories: [CollectionStory], width: CGFloat, margin: CGFloat) -> NSCollectionLayoutSection {
+        let numberOfColumns: CGFloat = showThreeColumnLayout ? 3 : 2
+        let recommendationsHeight: [CGFloat] = getRecommendationHeights(with: stories, width: width, margin: margin, numberOfColumns: numberOfColumns)
 
         /// Retrieves max height for each row and returns an array of row heights
         let totalGroupHeights = recommendationsHeight.getMaxHeightForRow(of: Int(numberOfColumns))
 
+        let components = createComponentsForWideLayout(with: stories, numberOfColumns: numberOfColumns, totalGroupHeights: totalGroupHeights)
+        return createSectionFromGroup(with: components, and: margin)
+    }
+
+    /// Create components needed for wide section layout
+    /// - Parameters:
+    ///   - recommendations: list of stories to display to the user
+    ///   - width: width that the section occupies
+    ///   - sideMargin: margin for the section layout
+    /// - Returns: tuple of total height and list of items used to determine section layout
+    private func createComponentsForWideLayout(with stories: [CollectionStory], numberOfColumns: CGFloat, totalGroupHeights: [CGFloat]) -> (CGFloat, [NSCollectionLayoutGroup]) {
+        let numberOfRows = (CGFloat(stories.count) / numberOfColumns).rounded(.up)
         let item = NSCollectionLayoutItem(
             layoutSize: NSCollectionLayoutSize(
                 widthDimension: .fractionalWidth(1 / numberOfColumns),
@@ -322,7 +327,7 @@ private extension CollectionViewController {
         )
         item.contentInsets = Constants.itemPadding
 
-        let components = (0..<Int(numberOfRows)).reduce((CGFloat(0), [NSCollectionLayoutGroup]())) { result, rowIndex in
+        return (0..<Int(numberOfRows)).reduce((CGFloat(0), [NSCollectionLayoutGroup]())) { result, rowIndex in
             let currentHeight = result.0
             guard let height = totalGroupHeights[safe: rowIndex] else { return result }
             var groups = result.1
@@ -338,24 +343,19 @@ private extension CollectionViewController {
             groups.append(group)
             return (currentHeight + height, groups)
         }
+    }
 
-        let group = NSCollectionLayoutGroup.vertical(
-            layoutSize: .init(
-                widthDimension: .fractionalWidth(1),
-                heightDimension: .absolute(components.0)
-            ),
-            subitems: components.1
-        )
-
-        let section = NSCollectionLayoutSection(group: group)
-        section.contentInsets = NSDirectionalEdgeInsets(
-            top: 0,
-            leading: margin,
-            bottom: 0,
-            trailing: margin
-        )
-
-        return section
+    /// Get a list of heights for each story
+    /// - Parameters:
+    ///   - recommendations: list of stories
+    ///   - width: width that the section occupies
+    ///   - sideMargin: padding adding to the side of the section
+    ///   - numberOfColumns: number of columns layout should have
+    /// - Returns: return a list of heights for all the stories
+    private func getRecommendationHeights(with stories: [CollectionStory], width: CGFloat, margin: CGFloat, numberOfColumns: CGFloat) -> [CGFloat] {
+        return stories.map { story in
+            return RecommendationCell.fullHeight(viewModel: model.storyViewModel(for: story), availableWidth: width / numberOfColumns - (margin * 2)) + margin
+        }
     }
 
     /// Determines the stories layout for the native collection view on iPhone mode (single column layout)
@@ -365,7 +365,18 @@ private extension CollectionViewController {
     ///   - margin: padding adding to the side of the section
     /// - Returns: section layout for compact (i.e. iPhone mode)
     private func collectionSectionForCompactLayout(with stories: [CollectionStory], width: CGFloat, margin: CGFloat) -> NSCollectionLayoutSection {
-        let components = stories.reduce((CGFloat(0), [NSCollectionLayoutItem]())) { result, story in
+        let components = createComponentsForCompact(with: stories, width: width, margin: margin)
+        return createSectionFromGroup(with: components, and: margin)
+    }
+
+    /// Create components needed for compact section layout
+    /// - Parameters:
+    ///   - recommendations: list of stories to display to the user
+    ///   - width: width that the section occupies
+    ///   - sideMargin: margin for the section layout
+    /// - Returns: tuple of total height and list of items used to determine section layout
+    private func createComponentsForCompact(with stories: [CollectionStory], width: CGFloat, margin: CGFloat) -> (CGFloat, [NSCollectionLayoutItem]) {
+        return stories.reduce((CGFloat(0), [NSCollectionLayoutItem]())) { result, story in
             let currentHeight = result.0
 
             let height = RecommendationCell.fullHeight(
@@ -386,15 +397,21 @@ private extension CollectionViewController {
 
             return (currentHeight + height, items)
         }
+    }
 
+    /// Creates section from group layout
+    /// - Parameters:
+    ///   - components: components that consist of height and list of items for the group
+    ///   - margin: margin for the section layout
+    /// - Returns: section layout
+    private func createSectionFromGroup(with components: (CGFloat, [NSCollectionLayoutItem]), and margin: CGFloat) -> NSCollectionLayoutSection {
         let group = NSCollectionLayoutGroup.vertical(
-            layoutSize: NSCollectionLayoutSize(
+            layoutSize: .init(
                 widthDimension: .fractionalWidth(1),
                 heightDimension: .absolute(components.0)
             ),
             subitems: components.1
         )
-
         let section = NSCollectionLayoutSection(group: group)
         section.contentInsets = NSDirectionalEdgeInsets(
             top: 0,
@@ -402,7 +419,6 @@ private extension CollectionViewController {
             bottom: 0,
             trailing: margin
         )
-
         return section
     }
 
--- a/PocketKit/Sources/PocketKit/Collections/CollectionViewController.swift
+++ b/PocketKit/Sources/PocketKit/Collections/CollectionViewController.swift
@@ -63,12 +63,9 @@ class CollectionViewController: UIViewController {
         return collectionView
     }()
 
+    private let sectionProvider: GridSectionLayoutProvider
     private var metadata: CollectionMetadata?
     private var subscriptions: [AnyCancellable] = []
-
-    private var showThreeColumnLayout: Bool {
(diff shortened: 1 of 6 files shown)


PocketKit/Sources/SharedPocketKit/Array+Extension.swift
```diff
@@ -13,3 +13,24 @@ extension Array {
         return self[index]
     }
 }
+
+extension Array where Element == CGFloat {
```
CMasterson: I don't think this should be an extension of Array. Can we swap to be a helper function?

PocketKit/Sources/PocketKit/Collections/CollectionViewController.swift
```diff
@@ -318,6 +289,123 @@ private extension CollectionViewController {
...
     }
 
+    /// Determines the stories layout in the native collection view on iPad mode with regular horizontal size class (including split view). Number of columns for the grid layout depends on device orientation and the view's width length. If view's width is less than 800, show 2 col, otherwise show 2 col if it is in portrait and 3 col if it is in landscape mode.
+    private func collectionSectionForWideLayout(with stories: [CollectionStory], width: CGFloat, margin: CGFloat) -> NSCollectionLayoutSection {
+        if self.view.bounds.width >= Constants.minWidthBoundaryForFullColumnLayout && !UIDevice.current.orientation.isPortrait {
+            return collectionSectionForiPadLayout(with: stories, width: width, margin: margin, numberOfColumns: 3)
+        } else {
+            return collectionSectionForiPadLayout(with: stories, width: width, margin: margin, numberOfColumns: 2)
+        }
+    }
```
CMasterson:
```suggestion
    private func collectionSectionForWideLayout(with stories: [CollectionStory], width: CGFloat, margin: CGFloat) -> NSCollectionLayoutSection {
        collectionSectionForiPadLayout(with: stories, width: width, margin: margin, numberOfColumns: numberOfColumns())
    }
    
    private func numberOfColumns() -> Int {
        if self.view.bounds.width >= Constants.minWidthBoundaryForFullColumnLayout && !UIDevice.current.orientation.isPortrait {
            return 2
         } else {
            return 3
         }
    }
```
Break up the logic a little.

PocketKit/Sources/PocketKit/Collections/CollectionViewController.swift
```diff
@@ -318,6 +289,123 @@ private extension CollectionViewController {
...
+    /// - Parameters:
+    ///   - stories: list of stories associated with collection to display
+    ///   - width: width that the section occupies
+    ///   - margin: padding adding to the side of the section
+    /// - Returns: section layout for stories on iPad view and regular horizontal size class
+    private func collectionSectionForiPadLayout(with stories: [CollectionStory], width: CGFloat, margin: CGFloat, numberOfColumns: CGFloat) -> NSCollectionLayoutSection {
+        let numberOfRows = (CGFloat(stories.count) / numberOfColumns).rounded(.up)
+
+        let recommendationsHeight: [CGFloat] = stories.map { story in
+            return RecommendationCell.fullHeight(viewModel: model.storyViewModel(for: story), availableWidth: width / numberOfColumns - (margin * 2)) + margin
```
CMasterson: Maybe pull `availableWidth` into it's own function?

PocketKit/Sources/PocketKit/Collections/CollectionViewController.swift
```diff
@@ -318,6 +289,123 @@ private extension CollectionViewController {
...
+        }
+    }
+
+    /// Determines the stories layout for the native collection view on iPad mode with regular horizontal size class
+    /// - Parameters:
+    ///   - stories: list of stories associated with collection to display
+    ///   - width: width that the section occupies
+    ///   - margin: padding adding to the side of the section
+    /// - Returns: section layout for stories on iPad view and regular horizontal size class
+    private func collectionSectionForiPadLayout(with stories: [CollectionStory], width: CGFloat, margin: CGFloat, numberOfColumns: CGFloat) -> NSCollectionLayoutSection {
```
CMasterson: I'd like to break this function up a bit.

PocketKit/Sources/PocketKit/Collections/CollectionViewController.swift
```diff
@@ -318,6 +289,123 @@ private extension CollectionViewController {
...
+        let item = NSCollectionLayoutItem(
+            layoutSize: NSCollectionLayoutSize(
+                widthDimension: .fractionalWidth(1 / numberOfColumns),
+                heightDimension: .fractionalHeight(1)
+            )
+        )
+        item.contentInsets = Constants.itemPadding
+
+        let components = (0..<Int(numberOfRows)).reduce((CGFloat(0), [NSCollectionLayoutGroup]())) { result, rowIndex in
+            let currentHeight = result.0
```
CMasterson, line 326: Pull into a function?

PocketKit/Sources/PocketKit/Collections/CollectionViewController.swift
```diff
@@ -318,6 +289,123 @@ private extension CollectionViewController {
...
+                layoutSize: .init(
+                    widthDimension: .fractionalWidth(1),
+                    heightDimension: .absolute(height)
+                ),
+                repeatingSubitem: item,
+                count: Int(numberOfColumns)
+            )
+            group.interItemSpacing = .fixed(Constants.interItemSpacing)
+            groups.append(group)
+            return (currentHeight + height, groups)
```
CMasterson: Pull into a function?

PocketKit/Sources/PocketKit/Home/SlateDetailViewController.swift
```diff
@@ -186,52 +193,131 @@ private extension SlateDetailViewController {
...
+            }
+        default:
+            return .empty()
+        }
+    }
 
-                items.append(item)
+    /// Determines the section layout for the slate detail view on iPad mode with regular horizontal size class (including split view). Number of columns for the grid layout depends on device orientation and the view's width length. If view's width is less than 800, show 2 col, otherwise show 2 col if it is in portrait and 3 col if it is in landscape mode.
+    private func slateSectionForWideLayout(with recommendations: [Recommendation], width: CGFloat, sideMargin: CGFloat) -> NSCollectionLayoutSection {
+        if self.view.bounds.width >= Constants.minWidthBoundaryForFullColumnLayout && !UIDevice.current.orientation.isPortrait {
```
CMasterson: The condition is hard to parse, would it be possible to pull into something like `func viewWidthExceedsPortraitMinimum() -> Bool`

PocketKit/Sources/PocketKit/Collections/CollectionViewController.swift
cyndichin: I kept the code consistent with previous logic. In this case availableWidth seems small enough to be included, but I can pull calculating heights into its own function.

PocketKit/Sources/PocketKit/Collections/CollectionViewController.swift
cyndichin: I think the function `collectionSectionForWideLayout` is not useful above as it just calls another function, so instead I will combine the two.

PocketKit/Sources/SharedPocketKit/Array+Extension.swift
cyndichin: curious to hear your perspective on why. The function is heavily tied to the array and acts as a method to the array. Not sure what the preference on the team to do standalone helper functions vs. extensions.

PocketKit/Sources/PocketKit/Home/SlateDetailViewController.swift
cyndichin: I updated it to a private computed variable 
```
    private var showThreeColumnLayout: Bool {
        return self.view.bounds.width >= Constants.minWidthBoundaryForFullColumnLayout && !UIDevice.current.orientation.isPortrait
    }
```

cyndichin:
thank you @CMasterson for the review! I updated the PR by moving some of the logic into separate functions and responded to some of your comments. Let me know if there's anything else I missed!

PocketKit/Sources/SharedPocketKit/Array+Extension.swift
Gio2018: I see both points here. On one hand, `getMaxHeightForRow` feels like a method that is specific for a `UITableView` (or `UICollectionView`). On the other, the operation is on an array.

My opinion is that we either add an helper or move this to where it's used and make it a `private` extension of `Array`.
Also, I'd ditch the `get` in the method name, but it's just a personal preference.

PocketKit/Sources/PocketKit/Collections/CollectionViewController.swift
```diff
@@ -318,6 +294,134 @@ private extension CollectionViewController {
...
     }
 
+    /// Determines the stories layout in the native collection view on iPad mode with regular horizontal size class (including split view). Number of columns for the grid layout depends on device orientation and the view's width length. If view's width is less than 800, show 2 col, otherwise show 2 col if it is in portrait and 3 col if it is in landscape mode.
+    /// - Parameters:
+    ///   - stories: list of stories associated with collection to display
+    ///   - width: width that the section occupies
+    ///   - margin: padding adding to the side of the section
+    /// - Returns: section layout for stories on iPad view and regular horizontal size class
+    private func collectionSectionForWideLayout(with stories: [CollectionStory], width: CGFloat, margin: CGFloat) -> NSCollectionLayoutSection {
+        let numberOfColumns: CGFloat = showThreeColumnLayout ? 3 : 2
```
Gio2018: instead of having a `showThreeColumnLayout`, we could maybe have a `numberOfColumns` property that does the exact same calculations. Seems more explicit and avoids one boolean property (which tend to be confusing as they might only represent part of the states that a view can have)

PocketKit/Sources/PocketKit/Collections/CollectionViewController.swift
```diff
@@ -254,46 +265,11 @@ private extension CollectionViewController {
...
-                )
-                item.contentInsets = NSDirectionalEdgeInsets(top: 8, leading: 0, bottom: 8, trailing: 0)
-
-                items.append(item)
-
-                return (currentHeight + height, items)
+            if environment.traitCollection.shouldUseWideLayout() {
+                return collectionSectionForWideLayout(with: stories, width: width, margin: margin)
+            } else {
+                return collectionSectionForCompactLayout(with: stories, width: width, margin: margin)
```
Gio2018: you could extract this logic in one single method, perhaps called `collectionSection(_:)`

PocketKit/Sources/PocketKit/Collections/CollectionViewController.swift
```diff
@@ -318,6 +294,134 @@ private extension CollectionViewController {
...
 
+    /// Determines the stories layout in the native collection view on iPad mode with regular horizontal size class (including split view). Number of columns for the grid layout depends on device orientation and the view's width length. If view's width is less than 800, show 2 col, otherwise show 2 col if it is in portrait and 3 col if it is in landscape mode.
+    /// - Parameters:
+    ///   - stories: list of stories associated with collection to display
+    ///   - width: width that the section occupies
+    ///   - margin: padding adding to the side of the section
+    /// - Returns: section layout for stories on iPad view and regular horizontal size class
+    private func collectionSectionForWideLayout(with stories: [CollectionStory], width: CGFloat, margin: CGFloat) -> NSCollectionLayoutSection {
+        let numberOfColumns: CGFloat = showThreeColumnLayout ? 3 : 2
+        let recommendationsHeight: [CGFloat] = getRecommendationHeights(with: stories, width: width, margin: margin, numberOfColumns: numberOfColumns)
```
Gio2018: nit: why not just rely on type inference?

Gio2018 (commented):
Thanks for working on this @cyndichin ! Looks like a big improvement for iPad UI.
I left some code comments, but I have one broader consideration:
as you mention, there's quite a bit of repeated code between `CollectionViewController` and `SlateDetailViewController`. What do you think of creating a specialized type that takes care of calculating the layout, and inject this in both view controllers instead?
Also (and I know this predates your PR), I can see that we are passing Core Data entities in the view controllers, that we use to instantiate view models, which contain references to core data entities, source etc. But what we actually need are only a few properties, defined in `RecommendationViewModel` (`attributedTitle`, `attributedCollection`, etc).
I wonder if we could just map the array of entities (e.g. `[CollectionStory]`) onto an array of types that only contains those attributed strings. You could even get away by adding an implementation of `RecommendationCellViewModel` that only contains the necessary attributes, or extract an additional protocol from it, and then extend it to `RecommendationCellViewModel`. The main purpose of doing all that, other than modularize and isolate responsibilities, is to avoid unnecessary references to Core Data entities, which may lead to unintended behavior. What do you think?

PocketKit/Sources/PocketKit/Collections/CollectionViewController.swift
cyndichin: update to 
```
    /// Determines the number of columns for the layout
    /// - Parameter view: view that we want to provide a layout for
    /// - Returns: 3 if the view meets the requirements for a three column layout, otherwise 2
    private func numberOfColumns(with view: UIView) -> Int {
        let hasRequirementForThreeColumnLayout = view.bounds.width >= Constants.minWidthBoundaryForFullColumnLayout && !UIDevice.current.orientation.isPortrait
        return hasRequirementForThreeColumnLayout ? 3 : 2
    }
```

PocketKit/Sources/PocketKit/Collections/CollectionViewController.swift
cyndichin: removed `[CGFloat]` to rely on type inference

CMasterson: Type inference is nearly 2x slower when compiling.
No difference when running, though.

https://forums.swift.org/t/regarding-swift-type-inference-compile-time-performance/49748

PocketKit/Sources/SharedPocketKit/Array+Extension.swift
CMasterson: If there is some core functionality to Array that we need in +50% of the app, then I think you have a good case for a public extension.
In this case, we only want this functionality in one or two spots, so does it make sense to make that available to the entire codebase?
In this case we are operating on an Array, yes. But it is an Array of heights, so you could say the functionality is more closely tied to Heights than to Array.

You could go so far as to make a happly little 
```
struct RowHeights {
    let heights: [CGFloat]

    func maxHeight() -> [CGFloat] {
        //Do the thing
    }
}
```
Which ties all the related functionality together in one place.

I do still have reservations around using Extensions as freely as we do. They obfuscate where code is and who wrote it.

PocketKit/Sources/PocketKit/Collections/CollectionViewController.swift
cyndichin: thanks @CMasterson~! I think it was helpful for me at the time of coding to have things explicit, but I'm open to what the team suggests if type inference is more readable. I removed for now

cyndichin:
Thank you @CMasterson and @Gio2018 for the suggestions ! The PR is looking a lot better and is ready for another look 👀 . In terms of tests, I didn't want to move functions from private to public just to test them and it seems more worth to have a UI test in this case. However, we don't have UI tests for iPad set up. Thoughts?

PocketKit/Sources/PocketKit/Collections/CollectionViewController.swift
cyndichin: I moved this to a separate class under `gridSection`

PocketKit/Sources/PocketKit/Home/SlateDetailViewController.swift
cyndichin: ^ outdated, I made some refactoring changes based on a combination of yours and Giorgio's comments.

PocketKit/Sources/SharedPocketKit/Array+Extension.swift
cyndichin: If I make it a private extension, then we won't be able rely on tests. Do we think a struct is over kill? I can convert it to a private function that takes in an array. But the operation being on an array seems more cleaner than injecting it. I have no strong opinions, so open to what's best for the team.

PocketKit/Sources/PocketKit/Home/GridSectionProvider.swift
```diff
@@ -0,0 +1,184 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import Foundation
+import CoreData
+import Sync
+import UIKit
+
+/// Configures the section layout for items in slate detail view and native collections
+class GridSectionLayoutProvider {
```
CMasterson, line 11: :chefkiss:

dskuza (approved):
Seems like a lot of conversation was made over the last day! Awesome. I think the outcome looks nice. Seems like a lot was addressed.

PocketKit/Sources/SharedPocketKit/Array+Extension.swift
Gio2018: you can make it a public extension, and maybe rename it in a more generic way?

Gio2018 (approved):
great work @cyndichin ! Code looks a lot more compact and readable now!! We are still passing references to Core Data entities in UI objects, but I am giving an approval since this was pre-existing, and any further refactor can safely be addressed in a separate PR.

PR #843, opened 2023-08-14
Universal links - Add links to Widget Rows

Gio2018:
## Summary
* This PR adds logic to send item links to the app in order to open them in the dedicated view

## References 
* Related widget Jira tickets

## Implementation Details
* Add `Link` modifier to `ItemWidgetRow`
* Add `Router` type to handle links

> [!IMPORTANT]
> the current implementation does not actually show the linked articles in the app yet. This will need to be added in a separate PR
> Unit tests will also need to be added here.

## Test Steps
* Put a breakpoint inside `Router` after the `guard` statement succeeds
* Install one instance of Saved Item or Recommendation widget, if you don't already have one
* tap on an article, and make sure the breakpoint gets hit

## PR Checklist:
- [ ] Added Unit / UI tests
- [x] Basic Self QA


PocketApp.swift
```diff
@@ -9,9 +9,11 @@ import SwiftUI
 struct PocketApp: App {
     @UIApplicationDelegateAdaptor var delegate: PocketAppDelegate
 
+    private let dependencyContainer = PocketAppDependencyContainer()
```
Gio2018: the use of `PocketAppDependencyContainer` is optional at this stage, since we don't do much injection from here. We can keep it if we feel that we might get rid of the singleton at some point

PocketKit/Sources/PocketKit/Main/PocketAppDependencyContainer.swift
```diff
@@ -0,0 +1,22 @@
...
+
+@MainActor
+public struct PocketAppDependencyContainer {
+    public init() {}
+    public func makeRootView() -> some View {
+        let model = makeRootViewModel()
+
+        return RootView(model: model)
+            .onOpenURL { url in
+                model.handleUrl(url)
```
Gio2018: we could even move the `onOpenUrl` code into `RootView` and attach it to `mainView`, since when we are logged out, we don't need to handle links

dskuza (approved):
Makes sense to me. 

Quick question, though: when it comes to _where_ the deeplink URL is handled, that is currently in the main app, and then what is stored for the widget to present later. There were two options - 1. the app creates the deeplink URL, 2. the widget creates the deeplink URL. Is the thought of going with option 1 that it makes more sense for the app to create URLs that it knows it can handle?

PocketKit/Sources/ItemWidgetsKit/Views/ItemWidgetsRow.swift
```diff
@@ -18,23 +18,26 @@ struct ItemWidgetsRow: View {
     let domain: String
     let readingTime: String?
     let image: Image?
+    let url: URL
```
CMasterson: Naming this `url` can be misleading.
All(?) pocket items have a url, this url in particular is a deeplink with a custom scheme which I would suggest requires more clarity.

PocketKit/Sources/SharedPocketKit/ItemContent.swift
```diff
@@ -54,10 +54,10 @@ public struct ItemContent: Codable, Equatable {
     }
 
     /// The URL used to open the saved item within the Pocket reader.
```
CMasterson: Updating this function makes this comment outdated.

PR #844, opened 2023-08-15
fix(sync): reset persistent container on load error

dskuza:
## Summary

Fixes an issue where the app would crash if a migration in the model version would error.

## References 

IN-1349

## Implementation Details

This pull request includes loading a persistent container outside of initialization, but very early in the app and share extension lifecycle.

The basic premise is: when loading the persistent container, attempt to load a store, and if that load fails, destroy and re-add the store, thus rebuilding an empty store from scratch. See comments for more details.

Upon reset, the app has to force-refresh all refresh coordinators, to act similar to login; sync ALL the on-disk data since it was removed in the reset. This will also present a toast in the main app. We have to force-refresh since there's a possibility a sync won't be called due to the refresh interval criteria not being met, which would cause apps to display no data. By force-refreshing, we reload all views across the app. 

## Test Steps

**Simulator Only**
- [ ] Checkout v8.0.5
- [ ] Build and run, login, and stop the build mid-sync
- [ ] Open the underlying Core Data SQLite database in an app like DB Browser
- [ ] Find the `PersistentSyncTask' table, and in any available rows, NULL-out `syncTaskContainer`, saving the changes
- [ ] Checkout this branch
- [ ] Launch the app
- [ ] The app should not crash, and a toast should appear that items are being redownloaded.

## PR Checklist:
- [ ] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketApp.swift
+++ b/PocketApp.swift
@@ -7,11 +7,20 @@ import SwiftUI
 
 @main
 struct PocketApp: App {
+    @Environment(\.scenePhase)
+    var scenePhase
+
     @UIApplicationDelegateAdaptor var delegate: PocketAppDelegate
 
     var body: some Scene {
         WindowGroup {
             RootView(model: RootViewModel())
+        }.onChange(of: scenePhase) { newValue in
+            switch newValue {
+            case .active:
+                delegate.scenePhaseDidChange(scenePhase)
+            default: return
+            }
         }
     }
 }
--- a/PocketKit/Sources/PocketKit/Services.swift
+++ b/PocketKit/Sources/PocketKit/Services.swift
@@ -12,7 +12,7 @@ import Kingfisher
 import Network
 
 struct Services {
-    static let shared: Services = { Services() }()
+    static let shared: Services = Services()
 
     let userDefaults: UserDefaults
     let appSession: AppSession
@@ -216,6 +216,13 @@ struct Services {
         recentSavesWidgetUpdateService = RecentSavesWidgetUpdateService(store: UserDefaultsItemWidgetsStore(userDefaults: userDefaults, key: .recentSavesWidget))
         recommendationsWidgetUpdateService = RecommendationsWidgetUpdateService(store: UserDefaultsItemWidgetsStore(userDefaults: userDefaults, key: .recommendationsWidget))
         widgetsSessionService = UserDefaultsWidgetSessionService(defaults: userDefaults)
+    }
+
+    /// Starts up all services as required.
+    /// - Parameter onReset: The function to call if a service has been reset.
+    /// - Note: `onReset` can be called when a migration within the persistent container fails
+    func start(onReset: @escaping () -> Void) {
+        persistentContainer.load(onReset: onReset)
     }
 }
 
--- a/PocketKit/Sources/Sync/PersistentContainer.swift
+++ b/PocketKit/Sources/Sync/PersistentContainer.swift
@@ -27,8 +27,11 @@ public class PersistentContainer: NSPersistentContainer {
         case inMemory
         case shared
     }
+    private let storage: Storage
 
     public init(storage: Storage = .shared, groupID: String) {
+        self.storage = storage
+
         ValueTransformer.setValueTransformer(ArticleTransformer(), forName: .articleTransfomer)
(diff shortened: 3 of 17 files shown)


PocketKit/Sources/PocketKit/Services.swift
```diff
@@ -217,6 +223,18 @@ struct Services {
         recommendationsWidgetUpdateService = RecommendationsWidgetUpdateService(store: UserDefaultsItemWidgetsStore(userDefaults: userDefaults, key: .recommendationsWidget))
         widgetsSessionService = UserDefaultsWidgetSessionService(defaults: userDefaults)
     }
+
+    func start() {
+        persistentContainer.load {
+            self.refreshCoordinators.forEach { $0.refresh(isForced: true) { } }
+            let data = BannerModifier.BannerData(
```
Gio2018: if we could move this logic into a separate method it'd be more readable maybe?

PocketKit/Sources/Sync/PersistentContainer.swift
```diff
@@ -61,13 +64,37 @@ public class PersistentContainer: NSPersistentContainer {
...
-        loadPersistentStores { storeDescription, error in
-            if let error = error as NSError? {
-                fatalError("Unresolved error \(error), \(error.userInfo)")
+        spotlightIndexer = CoreDataSpotlightDelegate(forStoreWith: storeDescription, coordinator: self.persistentStoreCoordinator)
+        spotlightIndexer?.startSpotlightIndexing()
+    }
+
+    public func load(onReset: @escaping () -> Void) {
+        loadPersistentStores { [weak self] storeDescription, error in
+            guard let self = self else { return }
```
Gio2018: `= self` is redundant and can be omitted

Gio2018 (commented):
just left some preliminary comments

PocketKit/Sources/PocketKit/Services.swift
dskuza: I'm planning on moving this out of the `shared` initializer!

PocketApp.swift
```diff
@@ -7,11 +7,20 @@ import SwiftUI
...
+    @Environment(\.scenePhase)
+    var scenePhase
+
     @UIApplicationDelegateAdaptor var delegate: PocketAppDelegate
 
     var body: some Scene {
         WindowGroup {
             RootView(model: RootViewModel())
+        }.onChange(of: scenePhase) { newValue in
+            switch newValue {
```
Gio2018, line 19: if `scenePhase` only serves `delegate`, could we pass it to `sceneDidBecomeActive(_:)` (maybe renamed to `sceneDidChange(_:)` and it'll know what to do?

PocketKit/Sources/Sync/PersistentContainer.swift
```diff
@@ -61,13 +64,43 @@ public class PersistentContainer: NSPersistentContainer {
...
+    }
+
+    /// Attempts to load the persistent container. If the load errors, the persistent container
+    /// is destroyed and rebuilt, performing a closure if a reset was necessary.
+    /// - Parameter onReset: Called when the persistent container required to be reset
+    /// - Note: If a reset was necessary, all previous on-disk data will be removed.
+    public func load(onReset: @escaping () -> Void) {
+        loadPersistentStores { [weak self] storeDescription, error in
+            guard let self else { return }
+            if error != nil {
```
Gio2018: do we want to log to Sentry that we found an error and we are attempting the reset?

Gio2018 (approved):
Couple more comments, but other than that ✅

PocketApp.swift
dskuza: Unless I'm missing something, I don't think we have access to that, since that's a part of `UISceneDelegate`. But, since we're using the SwiftUI view lifecycle, we don't have a scene delegate.

PocketKit/Sources/Sync/PersistentContainer.swift
dskuza: Probably!

PocketApp.swift
dskuza: Misunderstood! Addressed.

PR #846, opened 2023-08-17
Universal Links - Add navigation to an article selected from a widget

Gio2018:
## Summary
* This PR lets you tap on an article in a widget and show it in the app.

## References 
* Branch name

## Implementation Details
* Add `UrlValidator`, a type that validates urls against the supported universal links format(s)
* Update `MainViewModel`, add `handle(_ url: URL)`, which takes urls intercepted by the app, validates them and push the associated article to `HomeViewController`

## Test Steps
* Build/run this branch
* Login if necessary
* Install one or more widgets
* Tap on a widget item
* Verify that it shows correctly, regardless where you were in the app
* Go back, and verify you are on Home

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/Analytics/AppEvents/Home.swift
+++ b/PocketKit/Sources/Analytics/AppEvents/Home.swift
@@ -77,7 +77,7 @@ public extension Events.Home {
     /**
      Fired when a user clicks a card in the `Recent Saves` section
      */
-    static func RecentSavesCardContentOpen(url: String, positionInList: Int) -> ContentOpen {
+    static func RecentSavesCardContentOpen(url: String, positionInList: Int?) -> ContentOpen {
         return ContentOpen(
             contentEntity:
                 ContentEntity(url: url),
@@ -92,7 +92,7 @@ public extension Events.Home {
     /**
      Fired when a user clicks a card on Home using the /discover API
      */
-    static func SlateArticleContentOpen(url: String, positionInList: Int, slateId: String, slateRequestId: String, slateExperimentId: String, slateIndex: Int, slateLineupId: String, slateLineupRequestId: String, slateLineupExperimentId: String, recommendationId: String, destination: ContentOpen.Destination) -> ContentOpen {
+    static func SlateArticleContentOpen(url: String, positionInList: Int?, slateId: String, slateRequestId: String, slateExperimentId: String, slateIndex: Int?, slateLineupId: String, slateLineupRequestId: String, slateLineupExperimentId: String, recommendationId: String, destination: ContentOpen.Destination) -> ContentOpen {
         return ContentOpen(
             destination: destination,
             contentEntity:
--- a/PocketKit/Sources/PocketKit/Home/HomeViewController.swift
+++ b/PocketKit/Sources/PocketKit/Home/HomeViewController.swift
@@ -136,7 +136,6 @@ class HomeViewController: UIViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
         view.accessibilityIdentifier = "home"
-        navigationController?.delegate = self
 
         observeModelChanges()
 
@@ -677,36 +676,7 @@ extension HomeViewController {
     }
 }
 
-extension HomeViewController: UINavigationControllerDelegate {
-    func navigationController(_ navigationController: UINavigationController, willShow viewController: UIViewController, animated: Bool) {
-        // By default, when pushing the reader, switching to landscape, and popping,
-        // the list will remain in landscape despite only supporting portrait.
-        // We have to programatically force the device orientation back to portrait,
-        // if the view controller we want to show _only_ supports portrait
-        // (e.g when popping from the reader).
-        if viewController.supportedInterfaceOrientations == .portrait, UIDevice.current.orientation.isLandscape {
-            UIDevice.current.setValue(UIDeviceOrientation.portrait.rawValue, forKey: "orientation")
-        }
-    }
-
-    func navigationController(_ navigationController: UINavigationController, didShow viewController: UIViewController, animated: Bool) {
-        if viewController === self {
-            slateDetailSubscriptions = []
-            model.clearTappedSeeAll()
-            model.clearSelectedItem()
-        }
-
-        if viewController is SlateDetailViewController {
-            model.clearRecommendationToReport()
-            model.tappedSeeAll?.clearSelectedItem()
-        }
-    }
-
-    func navigationControllerSupportedInterfaceOrientations(_ navigationController: UINavigationController) -> UIInterfaceOrientationMask {
-        guard navigationController.traitCollection.userInterfaceIdiom == .phone else { return .all }
-        return navigationController.visibleViewController?.supportedInterfaceOrientations ?? .portrait
-    }
-
+extension HomeViewController {
     private func popToPreviousScreen() {
         if let presentedVC = navigationController?.presentedViewController {
             presentedVC.dismiss(animated: true) { [weak self] in
--- a/PocketKit/Sources/PocketKit/Home/HomeViewController.swift
+++ b/PocketKit/Sources/PocketKit/Home/HomeViewController.swift
@@ -136,6 +136,8 @@ class HomeViewController: UIViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
         view.accessibilityIdentifier = "home"
+        navigationController?.delegate = self
+
         observeModelChanges()
 
         collectionView.translatesAutoresizingMaskIntoConstraints = false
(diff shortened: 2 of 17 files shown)


PocketKit/Sources/PocketKit/Home/HomeViewController.swift
```diff
@@ -136,6 +136,8 @@ class HomeViewController: UIViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
         view.accessibilityIdentifier = "home"
+        navigationController?.delegate = self
```
Gio2018: > [!NOTE]
> this was missing in the existing implementation, thus the delegate methods, which have an implementation, were not called.

cyndichin: ooh nice catch! I'm surprised we didn't notice any issue prior to this PR in terms of the delegate methods not being called.

PocketKit/Sources/PocketKit/Main/MainViewModel.swift
```diff
@@ -217,12 +217,28 @@ class MainViewModel: ObservableObject {
...
-        // TODO: pass itemUrl to HomeViewModel to select the correct readable type
+        guard let item = source.fetchViewContextItem(itemUrl) else {
+            return
+        }
+        // show the item associated to the given URL
+        if let savedItem = item.savedItem {
+            home.select(savedItem: savedItem, readableSource: .widget)
+        } else if let recommendation = item.recommendation {
+            home.select(recommendation: recommendation, readableSource: .widget)
+        }
```
cyndichin, line 242: do we want to handle the case if an `item` has no savedItem of recommendation associated? send a log to sentry? (tho the case shouldn't be trigger if its coming from the widgets)

cyndichin (commented):
great work @Gio2018 ! I ran through the test steps and I was brought to Home from Saves, Settings. But when I navigated to an item, collection or slate detail, I remained on that page instead of Home tab. Is that expected?

dskuza (commented):
Cyndi leaves a good (optional, imo) comment.

A few things: 

1. A "good" refactor (that probably shouldn't be addressed by this PR) would be to update some functions to not pass around `NSManagedObject`s as arguments to functions, but rather their object IDs.
2. Can we verify with Analytics that it's okay to not add an index to the analytics? I understand that it's optional per the data structure (and thus would be good in Snowplow Micro, I think?), but I'm wondering if Analytics, for some metrics, are allowing UI entities without an index. Maybe it's worth trying to grab the index of the item we're opening within the list, rather than sending nil. Not sure of the complexity of that, though.

PocketKit/Tests/PocketKitTests/UrlValidatorTests.swift
```diff
@@ -0,0 +1,36 @@
...
+        let correctUrl = URL(string: correctUrlString)
+        XCTAssertNotNil(correctUrl)
+        let itemUrl = validator.getItemUrl(from: correctUrl!)
+        XCTAssertNotNil(itemUrl)
+        XCTAssertEqual(itemUrl, "https://example.com/this_is_an_article/")
+    }
+
+    func testWrongUrlReturnsNil() {
+        let validator = UrlValidator()
+        let index = Int.random(in: 0..<Self.validUrlNotFromWidgets.count)
```
CMasterson: Does this pull a random URL each time?

PocketKit/Tests/PocketKitTests/UrlValidatorTests.swift
```diff
@@ -0,0 +1,36 @@
...
+        let wrongtUrlString = Self.validUrlNotFromWidgets[index]
```
CMasterson: What does 'wrong' mean in this instance?

Also: wrong((t))

PocketKit/Sources/PocketKit/Main/MainViewModel.swift
Gio2018: > do we want to handle the case if an item has no savedItem of recommendation associated?

They would not come from a widget, so this will never happen here

Gio2018:
> 2. Can we verify with Analytics that it's okay to not add an index to the analytics? I understand that it's optional per the data structure (and thus would be good in Snowplow Micro, I think?), but I'm wondering if Analytics, for some metrics, are allowing UI entities without an index. Maybe it's worth trying to grab the index of the item we're opening within the list, rather than sending nil. Not sure of the complexity of that, though.

I looked at the analytics spreadsheet, and there were no specs about the UI element index (which, indeed, is optional). I can sure ask for clarification but won't hold this PR until after the all hands, in case the analytics team is not available (as it might be at this point Friday), would you agree @dskuza ?

PocketKit/Sources/Analytics/AppEvents/Home.swift
```diff
@@ -92,7 +92,7 @@ public extension Events.Home {
     /**
      Fired when a user clicks a card on Home using the /discover API
      */
-    static func SlateArticleContentOpen(url: String, positionInList: Int, slateId: String, slateRequestId: String, slateExperimentId: String, slateIndex: Int, slateLineupId: String, slateLineupRequestId: String, slateLineupExperimentId: String, recommendationId: String, destination: ContentOpen.Destination) -> ContentOpen {
+    static func SlateArticleContentOpen(url: String, positionInList: Int?, slateId: String, slateRequestId: String, slateExperimentId: String, slateIndex: Int?, slateLineupId: String, slateLineupRequestId: String, slateLineupExperimentId: String, recommendationId: String, destination: ContentOpen.Destination) -> ContentOpen {
```
CMasterson, line 95: Maybe a bit late to the party, can we make a model to pass in here as one item instead of the many we have now?

PocketKit/Tests/PocketKitTests/UrlValidatorTests.swift
Gio2018: good catch: probably best to rename as `invalid`

PocketKit/Tests/PocketKitTests/UrlValidatorTests.swift
Gio2018: yes

PocketKit/Sources/Analytics/AppEvents/Home.swift
Gio2018: this is legacy code, we can refactor but in a separate PR.

Gio2018:
> great work @Gio2018 ! I ran through the test steps and I was brought to Home from Saves, Settings. But when I navigated to an item, collection or slate detail, I remained on that page instead of Home tab. Is that expected?

@cyndichin , would you mind showing me the repro steps in a live chat? I can't repro..

PocketKit/Tests/PocketKitTests/UrlValidatorTests.swift
CMasterson: I really feel that tests should be the exact same every time they are run.
In this test case we could run through all the urls and make sure they all work as expected.

CMasterson (changes requested):

PocketKit/Tests/PocketKitTests/UrlValidatorTests.swift
Gio2018: Fine: I'll just use one url then

CMasterson: More is more better!

PocketKit/Tests/PocketKitTests/UrlValidatorTests.swift
```diff
@@ -0,0 +1,27 @@
...
+        XCTAssertEqual(itemUrl, "https://example.com/this_is_an_article/")
+    }
+
+    func testInvalidUrlReturnsNil() {
+        let validator = UrlValidator()
+        let invalidUrl = URL(string: "https://getpocket.com/hooray/not-an-article-from-widgets")
+        XCTAssertNotNil(invalidUrl)
+        let itemUrl = validator.getItemUrl(from: invalidUrl!)
+        XCTAssertNil(itemUrl)
+    }
```
CMasterson, line 26: Can you add one more test that throws some random garbage URL into the validator?

Gio2018: That would be a test for `URL.init(_:)`, so I don't think it fits here.

Gio2018:
@dskuza , as discussed on Slack, I removed the `UINavigationControllerDelegate` methods from `HomeViewController`, since they were actually not needed.

cyndichin (approved):
tested again and was able to see the items now. thank you!

PocketKit/Tests/PocketKitTests/UrlValidatorTests.swift
CMasterson: More what I was getting at was adding a valid URL that looks nothing like what the validator would be expecting

```let weirdoURL = URL(string: "testing:some.url.that.is.still.a/~valid/url?but=also.has&aBunch=of,stuff&array=that&array=we&array=dont&usually=have1234567890$-_.+!*'()")```

PR #850, opened 2023-08-21
fix(leaks):

Gio2018:
## Summary
* This PR prevents memory leaks and multiple instantiations of objects due to `WindowGroup` being called multiple times after we capture `scenePhase`, by injecting reference types generating from `RootViewModel` instead of instantiating it inside `WindowGroup`

## References 
* NA

## Implementation Details
* Update PocketApp, inject `RootViewModel` instead of instantiating it inside` WindowGroup`

## Test Steps
* Look into the memory graph and make sure there are no multiple instances of `HomeViewModel` or `MainViewModel`

## PR Checklist:
- [ ] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketApp.swift
+++ b/PocketApp.swift
@@ -7,14 +7,16 @@ import SwiftUI
 
 @main
 struct PocketApp: App {
+    @UIApplicationDelegateAdaptor var delegate: PocketAppDelegate
+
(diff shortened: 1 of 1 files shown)


dskuza (approved):
Interesting. Likely something we can look into for sure.

PR #861, opened 2023-09-06
fix(saveto): fall back to attempting public.url items as Data

dskuza:
## Summary

Fixes an issue where the "Save to Pocket" extension would not work when running the app on Apple Silicon.

## Implementation Details

On iOS, a `public.url` item would be provided to us as a `URL` type, but on Apple Silicon, it appears to be `Data`. This pull request introduces some basic logic where we first attempt to use the provided item as a `URL`, falling back to attempting as `Data`.

## Test Steps

- [ ] Install the app on an Apple Silicon machine
- [ ] Enable the "Save to Pocket" extension
- [ ] Open a URL in Safari
- [ ] Share to the "Save to Pocket" extension
- [ ] The item should be saved to your list, and the share extension should note that the item was saved

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA


CMasterson (approved):
Neat!

cyndichin (approved):
woo 🥳 and yay for adding tests!

PR #872, opened 2023-09-27
Add localized description to unknown sentry errors

CMasterson:
## Summary
Unknown errors are our highest cause of Sentry issues.
We check for one style of error and bundle the rest into 'unknown' which gives us no clarity into what is happening.

## References 
https://pocket.sentry.io/issues/4384022246/events/6b995ba2fd1e4736adb57f4a90a8a337/

### DIFF
--- a/PocketKit/Sources/Sync/ApolloClientProtocol+Extensions.swift
+++ b/PocketKit/Sources/Sync/ApolloClientProtocol+Extensions.swift
@@ -80,8 +80,7 @@ public extension ApolloClientProtocol {
         let skippableErrors = [403]
 
         guard let responseError = error as? ResponseCodeInterceptor.ResponseCodeError else {
-
-            Log.capture(message: "GraphQl Error - unknown error: \(error.localizedDescription)", filename: filename, line: line, column: column, funcName: funcName)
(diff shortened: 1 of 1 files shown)


PocketKit/Sources/Sync/ApolloClientProtocol+Extensions.swift
```diff
@@ -80,7 +80,8 @@ public extension ApolloClientProtocol {
         let skippableErrors = [403]
 
         guard let responseError = error as? ResponseCodeInterceptor.ResponseCodeError else {
-            Log.capture(message: "GraphQl Error - unknown error.", filename: filename, line: line, column: column, funcName: funcName)
+
+            Log.capture(message: "GraphQl Error - unknown error: \(error.localizedDescription)", filename: filename, line: line, column: column, funcName: funcName)
```
Gio2018: adding error info is absolutely valid. One thing I'd change, based on my experience is using `\(error)` instead of `\(error.localizedDescription)` because the former usually carries a lot more info, while the latter will often contain messages like "unknown error" or such.

Gio2018 (approved):
I think this makes sense @CMasterson ! I'd only apply the suggested small change, other than that :shipit: !

cyndichin (approved):
thanks! :)

PR #880, opened 2023-10-31
fix(tests): update corpus item for ReportARecommendationTests

cyndichin:
## Summary
Fixes the following UI tests:
* ReportARecommendationTests

## References 
MOSOIOS-70

## Implementation Details
* Modify url strings to match those in `corpusSlates.json`
* Update syndicated item to use item-2 instead of item 4 as we had in our old slates.json file

## Test Steps
Run the `ReportARecommendationTests` tests and ensure that it passes locally.

## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

## Screenshots
| ReportARecommendation | 
| --- |
| image |

### DIFF
--- a/Tests iOS/Fixtures/recommendation-detail-2.json
+++ b/Tests iOS/Fixtures/recommendation-detail-2.json
@@ -1,0 +1,47 @@
+{
+  "data": {
+    "__typename": "[type-name-here]",
+    "itemByUrl": {
+      "remoteID": "recommended-item-2",
+      "__typename": "[some-type-name]",
+      "givenUrl": "https://getpocket.com/explore/item/article-2",
+      "resolvedUrl": "https://getpocket.com/explore/item/article-2",
+      "title": "Slate 1, Recommendation 2",
+      "language": "en",
+      "topImageUrl": "http://example.com/slate-1-rec-2/top-image.png",
+      "timeToRead": 3,
+      "wordCount": 200,
+      "authors": [
+
+      ],
+      "collection": null,
+      "datePublished": "2021-01-01 12:01:01",
+      "images": [
+
+      ],
+      "excerpt": "Cursus Aenean Elit",
+      "domain": "slate-1-rec-2.example.com",
+      "domainMetadata": {
+        "__typename": "DomainMetadata",
+        "name": "Pocket",
+        "logo": "https://getpocket.com/logo.png"
+      },
+      "isArticle": true,
+      "hasImage": "HAS_IMAGES",
+      "hasVideo": "HAS_VIDEOS",
+      "syndicatedArticle": {
+        "__typename": "SyndicatedArticle",
+        "itemId": "recommended-item-2",
+        "title": "Slate 1, Rec 2",
+        "excerpt": "A better syndicated excerpt",
+        "mainImage": "http://example.com/slate-1-rec-2/top-image.png",
+        "publisher": {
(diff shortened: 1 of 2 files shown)


Tests iOS/Fixtures/corpusSlates.json
```diff
@@ -89,6 +89,30 @@
...
+                                    "mainImage": "image-2-2-4",
+                                    "title": "syndicatedTitle-2-2-4",
+                                    "excerpt": "syndicatedExcerpt-2-2-4",
+                                    "publisher": {
+                                      "__typename": "Publisher",
+                                      "name": "syndicatedPublisher-2-2-4"
+                                    }
+                                }
+                            }
+                        },
```
Gio2018: Hey @cyndichin , instead of adding another item in this fixture, would it be possible to use `item-2` (or `item-3`), which are already `SyndicatedArticle` for this purpose? Reason I am asking is because I am adding another item to this same fixture, which needs not to be a `SyndicatedArticle`, and would try to keep the fixture as compact as possible. What do you think?

Gio2018 (commented):
Thank you @cyndichin , had a comment in regard to this PR

Tests iOS/Fixtures/corpusSlates.json
cyndichin: I wanted to mimic our old `slates.json` file to avoid breaking more tests. A couple of tests rely on `Syndicated Article Slate 2, Rec 2` which is missing from the `corpusSlates.json`. If we want to test an item that is not a `SyndicatedArticle`, shall we use `Slate 1, Rec 1` for that purpose?

Overall, would it be beneficial to modify the fixtures so that they are similar as opposed to having to modify test code that may span across different classes?

Gio2018: > shall we use Slate 1, Rec 1 for that purpose?

that one is a collection, and not sure if it's used that way in other tests

Gio2018: > A couple of tests rely on Syndicated Article Slate 2, Rec 2

if we know what they are, can we just update the reference in those tests too? Still with the purpose of not making the fixture too big

cyndichin: `Slate 1, Rec 2 and Slate 2, Rec 1` are also not syndicated in the old `slates.json`, so those are valid as well

The other related syndicated tests are in `HomeTests`, so we can update them when working on that ticket. However, if I modify the other recommendation items to be syndicated, it may impact the previous tests that rely on the recommendations not being syndicated articles.

cyndichin: happy to go either direction, but just wanted to express those concerns

Gio2018: What I am thinking is:
- to fix `ReportARecommendationTests`, update the references in it, so it uses `item-2` or `item-3`. This would be the least amount of changes, with no changes in fixtures: this should ensure that we do not introduce other failures
- I am working on `HomeTests` and will fix anything that need fixing in there
Does that work?

cyndichin: Thanks! I updated the tests and decided to use `item-2`. I had to add a new fixture to replace `recommendation-detail-4` so that we don't introduce any new failures.

Gio2018 (approved):

PR #886, opened 2023-11-01
feat(listen): add listen support to an individual item [INXT-1362]

dskuza:
## Summary

Adds a Listen button when viewing an individual item in the reader view.

## Implementation Details

- Adds a `UIBarButtonItem` to `ReadableHostViewController` that uses its view model to determine listen support for the presented item
- Item support / eligibility has been extracted to a convenience getter, which is reused when building the Listen view model from the main list, as well as determining single-item support

## PR Checklist:
- [ ] Added Unit / UI tests
- [x] Basic Self QA

## Screenshots
image

### DIFF
--- a/PocketKit/Sources/PocketKit/MyList/SavesContainerViewController.swift
+++ b/PocketKit/Sources/PocketKit/MyList/SavesContainerViewController.swift
@@ -257,6 +257,14 @@ extension SavesContainerViewController {
             self?.showListen(listenViewModel: listenViewModel)
         }.store(in: &subscriptions)
 
+        model.savedItemsList.$presentedListenViewModel
+            .filter { $0 != nil }
+            .sink { [weak self] listenViewModel in
+                guard let listenViewModel else { return }
+                self?.showListen(listenViewModel: listenViewModel)
+                self?.model.savedItemsList.presentedListenViewModel = nil
+            }.store(in: &subscriptions)
+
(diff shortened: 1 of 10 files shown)


PocketKit/Sources/PocketKit/MyList/SavesContainerViewController.swift
```diff
@@ -369,6 +378,14 @@ extension SavesContainerViewController {
             self?.present(viewModel: addTagsViewModel)
         }.store(in: &readableSubscriptions)
 
+        readable.$presentedListenViewModel
+            .filter { $0 != nil }
+            .sink { [weak self] listenViewModel in
+                guard let listenViewModel else { return }
+                self?.showListen(listenViewModel: listenViewModel)
+                readable.presentedListenViewModel = nil
+            }.store(in: &subscriptions)
```
dskuza: This is a quirky thing to do. Basically, this is a publisher-as-a-delegate. This is done so that we can _use_ the requested view model to kickstart Listen, but the "host" view model (in this highlighted case, the reader view) doesn't really need to maintain a reference to it, only supply it. To not retain the view model here, we immediately nil-out the supplied view model.

What's a more appropriate way?

dskuza: Cleaned up in a followup commit.

Gio2018 (approved):
:shipit: !

PR #901, opened 2023-11-15
feat: lift article load limit to 10,000 items

nzeltzer:
## Summary
* Raises limit of downloaded articles after sign-in to 10,000 for both saves and archived articles.


bassrock (approved):

Gio2018 (approved):

dskuza:
@bassrock this, for some reason, didn't kick off a proper bitrise build, so the status check is borked. I tried starting a build from the branch, but that doesn't update the checks of this pull request, since the build isn't tied to it. Can we properly start a build, or is the change small enough where we can comfortably merge? I feel it's the latter.

PR #915, opened 2023-12-13
Highlights: display existing highlights

Gio2018:
## Summary
* This PR lets you display existing highlights in the Reader

## Implementation Details
* Add `ArticleComponentHighlights`, a collection of extensions that enable highlights:
    - extend `[ArticleComponent]` to add the `highlighted(_ patches: [String]) -> [ArticleComponent]` method: this method uses `DiffMatchPatch` to apply the given patches to the components. Components will be transformed in a single `String blob`, then, after the patches are applied, will be deconstructed back into an array of `ArticleComponent`
    - extend `NSAttributedString` to find the tags applied by the above step and convert them into highlighted text

> [!NOTE]
> This PR does not cover adding new highlights. These will be added in a separate PR.

## Test Steps
* Build/run this branch and login with an account that has articles with existing highlights
* Make sure the highlights appear in the reader

## PR Checklist:
- [ ] Added Unit / UI tests
- [x] Basic Self QA


PocketKit/Frameworks/DiffMatchPatch.xcframework/Info.plist
```diff
@@ -0,0 +1,61 @@
...
+			<key>SupportedPlatform</key>
+			<string>ios</string>
+			<key>SupportedPlatformVariant</key>
+			<string>simulator</string>
+		</dict>
+		<dict>
+			<key>BinaryPath</key>
+			<string>DiffMatchPatch.framework/Versions/A/DiffMatchPatch</string>
+			<key>LibraryIdentifier</key>
+			<string>ios-arm64_x86_64-maccatalyst</string>
```
bassrock, line 42: ➕

PocketKit/Package.swift
```diff
@@ -42,6 +43,10 @@ let package = Package(
             url: "https://github.com/Pocket/pocket-ios/releases/download/release%2Fv8.2.0-offline-tts/PKTListen.xcframework.zip",
             checksum: "47a318af51ff3196d142ede33fe1da33f93b6d6c459b5834e0968bb4b02406e9"
         ),
+        .binaryTarget(
```
bassrock, line 46: Should we host DiffMatchPatch  on a Github Release like we do Listen? (or find a better hosted place for these 2)

PocketKit/Sources/PocketKit/Article/Highlights/ArticleComponentHighlights.swift
```diff
@@ -0,0 +1,249 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import DiffMatchPatch
+import SharedPocketKit
+import Sync
+import UIKit
+
+extension Array where Element == ArticleComponent {
```
bassrock, line 10: this is cool

PocketKit/Sources/PocketKit/Article/Highlights/ArticleComponentHighlights.swift
```diff
@@ -0,0 +1,249 @@
...
+                scanner.currentIndex = highlightableString.index(after: scanner.currentIndex)
+            }
+        }
+        let mutable = NSMutableAttributedString(attributedString: self)
+        highlightableRanges.forEach {
+            let nsRange = NSRange($0, in: highlightableString)
+            mutable.addAttribute(.backgroundColor, value: HighlightConstants.highlightColor, range: nsRange)
+            mutable.addAttribute(.foregroundColor, value: HighlightConstants.highlightedTextColor, range: nsRange)
+        }
+        mutable.mutableString.replaceOccurrences(
```
bassrock, line 206: Will this affect the highlighted range above or does NSAttributedString know that?

PocketKit/Sources/PocketKit/Article/Highlights/ArticleComponentHighlights.swift
```diff
@@ -0,0 +1,249 @@
...
+import DiffMatchPatch
+import SharedPocketKit
+import Sync
+import UIKit
+
+extension Array where Element == ArticleComponent {
+    /// Apply an array of patches to an array of `ArticleComponent`
+    /// - Parameter patches: the patches to apply
+    /// - Returns: the patched components
+    func highlighted(_ patches: [String]) -> [ArticleComponent] {
```
bassrock, line 14: One note here, is that this whole file given that its the meat, has a lot of conditionals, and has, lets call it, "fun" could prob do with some more comments about the purpose of every few lines.

PocketKit/Sources/Sync/RemoteMapping/SavedItem+RemoteMapping.swift
```diff
@@ -24,17 +24,31 @@ extension SavedItem {
...
         createdAt = Date(timeIntervalSince1970: TimeInterval(remote._createdAt))
         deletedAt = remote._deletedAt.flatMap(TimeInterval.init).flatMap(Date.init(timeIntervalSince1970:))
         archivedAt = remote.archivedAt.flatMap(TimeInterval.init).flatMap(Date.init(timeIntervalSince1970:))
         isArchived = remote.isArchived
         isFavorite = remote.isFavorite
-
-        guard let context = managedObjectContext else {
-            Log.capture(message: "Managed context was nil")
-            return
+        if let annotations = remote.annotations, let remoteHighlights = annotations.highlights, !remoteHighlights.isEmpty {
```
bassrock, line 39: What if a user had highlights, but then deletes them from the web? I think this skips that case? Also if thats followup and the main thing here was highlight blob creation ignore me!

bassrock (approved):
The main highlight logic looks sound to me, but I think should wait for @nzeltzer to review too, as theres a lot going on there, this might be our most complicated feature!

I have some questions about user scenarios, but I am gonna leave that out for now since the main point of this PR is only about the highlight algorithm logic!

Really nice work @Gio2018 I know this feature and logic is something thats more intense then it should be given some of the architecture. Way to go!

nzeltzer (approved):
lgtm

PocketKit/Sources/PocketKit/Article/Highlights/ArticleComponentHighlights.swift
nzeltzer: NSAttributedString should auto-update its attribute dictionaries for the string replacement.

PocketKit/Package.swift
Gio2018: We can but the current hosting we are using for listen is less than ideal, so until we find it, I'd rather keep it here

PocketKit/Sources/PocketKit/Article/Highlights/ArticleComponentHighlights.swift
Gio2018: yup, thanks @nzeltzer ! Any changes to the string is reflected to the attributed one. If that wasn't the case, we'd see the tags in the article text

PocketKit/Sources/Sync/RemoteMapping/SavedItem+RemoteMapping.swift
Gio2018: good catch @bassrock ! Indeed if no highlights are found, we should clear them

PocketKit/Sources/PocketKit/Article/Highlights/ArticleComponentHighlights.swift
Gio2018: will do!

PR #930, opened 2024-01-22
fix(analytics): adding in a way to get mac analytics

bassrock:
## Summary

Currently we have no way to segment our analytics by Mac users. This at least will let us attribute actions to mac users, even if the internal snowplow data still says its iPad.

Hopefully https://github.com/snowplow/snowplow-ios-tracker/issues/863 gets traction to give us better tracking.


Gio2018 (approved):

kirill-demtchouk:
@bassrock will Mac events appear with a separate API ID as well? I can't recall what the existing Mac app ID is, but perhaps we can re-use that one.

bassrock:
No it won't. Would that make this easier?

PR #937, opened 2024-01-27
Add Highlights

Gio2018:
## Summary
* This PR adds the ability to create highlights from iOS

## References 
*  branch name
## Implementation Details
* Update `SavedItemViewModel`, add logic to:
    - Create a patch from a quote in a component
    - Create and save a local `Highlight`
    - Sync the highlight to the backend
- Update `ArticleComponentTextCell`, add logic to capture an highlightable range and send it to the view model via callback

## Test Steps
Premium users:
* Open an article in Reder, select some text:
    - if the selection includes an highlight, make sure you don't see an Highlight button in the context menu
    - if the selection does not include an highlight, make sure you see the Highlight button
    - Tap "Highlight" and make sure the highlight gets added
    - Optional: check on another platform that the highlight exists
Free users:
- Same as above but if you have already saved 3 highlights, the "Highlight" button should do nothing

> [!NOTE] 
> the alert/premium upsell will be added in another PR
 

## PR Checklist:
- [ ] Added Unit / UI tests
- [x] Basic Self QA


bassrock (approved):

PocketKit/Sources/Sync/PocketSource.swift
```diff
@@ -504,6 +504,56 @@ extension PocketSource {
...
+                version: 2
+            )
+            var highlights = (savedItem.highlights?.array ?? [])
+            highlights.append(highlight)
+            savedItem.highlights = NSOrderedSet(array: highlights)
+
+            do {
+                try space.save()
+            } catch {
+                Log.capture(error: error)
```
bassrock, line 532: Do we need to do anything else here if this fails?

bassrock (approved):
LGTM

PocketKit/Sources/Sync/PocketSource.swift
Gio2018: Probably not at this stage, we use this pattern on almost all `Source` methods. We can consider revisiting (and perhaps consider if we still want to trigger backend updates or not), but I'd leave that to a broader `Source` revisit/refactor

PR #963, opened 2024-03-11
Share: update shortUrl on demand, if needed, when sharing an url

Gio2018:
## Summary
* This PR fetches `shortUrl` if it's nil before sharing an item, in order to ensure we share `shortUrl` every time we can
## References 
* Branch name

## Implementation Details
* Add graphQl query to fetch an `Item`'s `shortUrl`
* Use it to update `shortUrl` if nil, when sharing an item: this can happen when users migrate from previous versions to `8.5.0`
## Test Steps
* Install the production (App Store) version and perform a refresh
* Within 5 minutes, install this branch as an upgrade (do not delete the prod version)
* Put a breakpoint where the share logic happens, for example [here](https://github.com/Pocket/pocket-ios/blob/5bcff04f2fac59e8197aaa6e893701d98e30542a/PocketKit/Sources/PocketKit/Home/HomeViewModel.swift#L799)
* verify that `shortUrl` is nil
* put another breakpoint [here](https://github.com/Pocket/pocket-ios/blob/5bcff04f2fac59e8197aaa6e893701d98e30542a/PocketKit/Sources/PocketKit/Home/HomeViewModel.swift#L806) and verify that the shortUrl has been fetched and assigned to the corresponding item
* verify that share works properly, and that you are actually sharing a `pocket.co` url
* 
## PR Checklist:
- [x] Added Unit / UI tests
- [x] Basic Self QA

### DIFF
--- a/PocketKit/Sources/PocketKit/Article/ReadableViewModel/RecommendableItemViewModel.swift
+++ b/PocketKit/Sources/PocketKit/Article/ReadableViewModel/RecommendableItemViewModel.swift
@@ -61,6 +61,14 @@ class RecommendableItemViewModel: ReadableViewModel {
         self.savedItemCancellable = item.publisher(for: \.savedItem).sink { [weak self] savedItem in
             self?.update(for: savedItem)
         }
+
+        if let url = item.shortURL {
+            self.shortUrl = url
+        } else {
+            Task {
+                self.shortUrl = try? await source.getItemShortUrl(item.givenURL)
+            }
+        }
     }
 
     var components: [ArticleComponent]? {
@@ -96,9 +104,7 @@ class RecommendableItemViewModel: ReadableViewModel {
         item.bestURL
     }
 
-    var shortUrl: String? {
-        item.shortURL
-    }
+    var shortUrl: String?
 
     var itemSaveStatus: ItemSaveStatus {
         guard let savedItem = item.savedItem else {
--- a/PocketKit/Sources/PocketKit/Home/HomeViewModel.swift
+++ b/PocketKit/Sources/PocketKit/Home/HomeViewModel.swift
@@ -646,7 +646,9 @@ extension HomeViewModel {
             favoriteAction: favoriteAction,
             overflowActions: [
                 .share { [weak self] sender in
-                    self?.share(savedItem, at: indexPath, with: sender)
+                    Task {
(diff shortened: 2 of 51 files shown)


PocketKit/Sources/PocketKit/Article/ReadableViewModel/RecommendableItemViewModel.swift
```diff
@@ -61,6 +61,14 @@ class RecommendableItemViewModel: ReadableViewModel {
         self.savedItemCancellable = item.publisher(for: \.savedItem).sink { [weak self] savedItem in
             self?.update(for: savedItem)
         }
+
+        if let url = item.shortURL {
+            self.shortUrl = url
+        } else {
+            Task {
+                self.shortUrl = try? await source.getItemShortUrl(item.givenURL)
```
bassrock, line 69: Is there a risk of no short url by the time the user saves?

PocketKit/Sources/PocketKit/Home/HomeViewModel.swift
```diff
@@ -731,8 +733,10 @@ extension HomeViewModel {
         let viewModel = HomeItemCellViewModel(
             item: sharedWithYouItem.item,
             overflowActions: [ .share { [weak self] sender in
-                    self?.share(sharedWithYouItem, at: indexPath, with: sender)
-                }],
+                Task {
+                    await self?.share(sharedWithYouItem, at: indexPath, with: sender)
```
bassrock, line 737: Does this work since its ui and needs main thread?

bassrock (approved):
approving, but have a few questions around the async behavior

PocketKit/Sources/PocketKit/Home/HomeViewModel.swift
Gio2018: it does because `HomeViewModel` is marked as `@MainActor`

<p align=center>


</p>

PocketKit/Sources/PocketKit/Article/ReadableViewModel/RecommendableItemViewModel.swift
Gio2018: see slack discussion, short answer: probably not a lot of real world cases, except offline users sharing via airdrop, where we would share the original url anyway.

PR #1018, opened 2024-05-30
chore(deps): may 2024 deps

bassrock:
## Summary

Update all the dependencies to the latest

### DIFF
--- a/PocketKit/Package.swift
+++ b/PocketKit/Package.swift
@@ -25,16 +25,16 @@ let package = Package(
         .library(name: "DiffMatchPatch", targets: ["DiffMatchPatch"])
     ],
     dependencies: [
-        .package(url: "https://github.com/apollographql/apollo-ios.git", from: "1.9.0"),
+        .package(url: "https://github.com/apollographql/apollo-ios.git", from: "1.12.2"),
         .package(url: "https://github.com/onevcat/Kingfisher.git", from: "7.11.0"),
         .package(url: "https://github.com/getsentry/sentry-cocoa.git", from: "8.22.4"),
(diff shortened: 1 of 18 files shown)


PocketKit/Package.swift
```diff
@@ -25,16 +25,16 @@ let package = Package(
...
     dependencies: [
-        .package(url: "https://github.com/apollographql/apollo-ios.git", from: "1.9.0"),
+        .package(url: "https://github.com/apollographql/apollo-ios.git", from: "1.12.2"),
         .package(url: "https://github.com/onevcat/Kingfisher.git", from: "7.11.0"),
         .package(url: "https://github.com/getsentry/sentry-cocoa.git", from: "8.22.4"),
-        .package(url: "https://github.com/snowplow/snowplow-objc-tracker", from: "6.0.1"),
-        .package(url: "https://github.com/ccgus/fmdb", from: "2.7.6"),
-        .package(url: "https://github.com/airbnb/lottie-ios.git", from: "4.4.0"),
+        .package(url: "https://github.com/snowplow/snowplow-objc-tracker", from: "6.0.3"),
+        .package(url: "https://github.com/ccgus/fmdb", from: "2.7.11"),
```
Gio2018, line 32: it shouldn't be necessary to update the minors since we have the `from version` parameter. On the other hand it won't hurt

Gio2018 (approved):

PocketKit/Package.swift
bassrock: @Gio2018 yep, but the lock file was out of date, and this will help remove them from the dep issue.

PR #1024, opened 2024-06-17
fix(home): switching all analytics to corpus recommendation id

bassrock:
## Goal

All analytics from Home should carry with them the Corpus Recommendation Entity since we switched to unified home.

Separately all reports from Home should contain the recommendation id so we can tell where it came from.

## Test Steps
* Ensure that all actions that are from a recommendation send a corpus recommendation entity


PocketKit/Sources/Analytics/AppEvents/ExpandedSlate.swift
```diff
@@ -30,7 +30,7 @@ public extension Events.ExpandedSlate {
     /**
      Fired when a user clicks a card on Home using the /discover API
      */
-    static func SlateArticleContentOpen(url: String, positionInList: Int, slateId: String, slateRequestId: String, slateExperimentId: String, slateIndex: Int, slateLineupId: String, slateLineupRequestId: String, slateLineupExperimentId: String, recommendationId: String, destination: ContentOpen.Destination) -> ContentOpen {
+    static func SlateArticleContentOpen(url: String, positionInList: Int, recommendationId: String, destination: ContentOpen.Destination) -> ContentOpen {
```
Gio2018, line 33: since we are at this, can we change the name and use the correct `lowerCamelCase` convention? Same for all the others, thank you!

Gio2018 (changes requested):
Thanks for tackling this @bassrock ! I have a small change request, other than that looks good!

PocketKit/Sources/Analytics/AppEvents/ExpandedSlate.swift
bassrock: @Gio2018  Can we do that in a seperate pr for all the analytics functions? Seems out of place in this one.

Gio2018: My point of view is that, if we refactor some code and spot improvements in that same lines of code, then it should be done at the moment. 
The situation in analytics methods is already mixed (meaning we already have methods that use the correct convention, and some others don't), so this won't introduce any disruption.
I can take care of this, but in this case let's leave this PR open and I'll merge it after I operate the changes. Thanks!

bassrock: I agree. My issue with doing it in this pr is if we are going to enforce a standard we should do it via a linter or some pre-established rules. This was just a caught off guard surprise to me.

Gio2018: Sounds good, we can merge and then I can fix the naming. One thing: seems like a bunch of tests are failing, can you double check it's not related to the changes?

Gio2018 (approved):

PR #1029, opened 2024-07-02
Add Tip to explain Shared With You Actions

Gio2018:
## Goal
* This PR adds an explanatory tip for the Shared With You Actions:
    - Tap to reply
    - Long press to delete an item or reply

## Test Steps
* Build run on an actual device where you have shared Pocket content from another user on iMessage
* Verify that the tip shows (see screenshot)

## Screenshots
<p align=center>

</p>

### DIFF
--- a/PocketKit/Sources/Localization/Strings.swift
+++ b/PocketKit/Sources/Localization/Strings.swift
@@ -728,8 +728,8 @@ public enum Localization {
     public enum SharedWithYouActions {
       /// Tap the attribution view to reply, long press it to remove an item or reply.
       public static let message = Localization.tr("Localizable", "tips.sharedWithYouActions.message", fallback: "Tap the attribution view to reply, long press it to remove an item or reply.")
-      /// Shared With You actions
-      public static let title = Localization.tr("Localizable", "tips.sharedWithYouActions.title", fallback: "Shared With You actions")
+      /// Tap here to reply to this message or long press to remove this item.
+      public static let title = Localization.tr("Localizable", "tips.sharedWithYouActions.title", fallback: "Tap here to reply to this message or long press to remove this item.")
     }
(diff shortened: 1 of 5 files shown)


bassrock (approved):

PocketKit/Sources/Localization/Strings.swift
```diff
@@ -725,6 +725,12 @@ public enum Localization {
         public static let title = Localization.tr("Localizable", "tips.recommendationsWidget.selectTopic.title", fallback: "New Recommendations Widget!")
       }
     }
+    public enum SharedWithYouActions {
+      /// Tap the attribution view to reply, long press it to remove an item or reply.
+      public static let message = Localization.tr("Localizable", "tips.sharedWithYouActions.message", fallback: "Tap the attribution view to reply, long press it to remove an item or reply.")
```
dnarcese: Can we change this to "Tap here to reply to this message or long press to remove this item."
Can we remove the title?

Gio2018: @dnarcese yes we can change it. Keep in mind that long press won't only show the option to remove the item, but also to reply (so basically there's two ways to reply). Apple requires the title to be present, while the message is optional: we could set the title to the message to achieve what you say: the only thing is that it will be bolded.
Does that work?

dnarcese: @Gio2018 Ok, keep the title just change the description, thanks!

Gio2018: Here's what I mean:

<p align=center>

</p>

Gio2018: ok changes made! Here's the final version

<p align=center>


<p>

PR #1037, opened 2024-07-12
Fixes: crash and localized collection

Gio2018:
## Goal
* This PR attempts to fix a background crash reported in Sentry, and adds support for localized collections

## Test Steps
* make sure localized collections open in the native view
* Background the app, enable/disable the internet connection and make sure there are no crashes

## Issues
POCKET-10334
POCKET-10335

### DIFF
--- a/PocketKit/Sources/PocketKit/Home/HomeViewModel.swift
+++ b/PocketKit/Sources/PocketKit/Home/HomeViewModel.swift
@@ -197,12 +197,17 @@ class HomeViewModel: NSObject {
 
     /// Fetch the latest data from core data and get the NSFetechedResults Controllers subscribing to updates
     func fetch() {
-        do {
-            try recentSavesController.performFetch()
-            try recomendationsController.performFetch()
-            try sharedWithYouController.performFetch()
-        } catch {
-            Log.capture(error: error)
+        // NOTE: despite HomeViewModel runs on MainActor, this call ends up on a different thread
+        // when the app is backgrounded, thus we force it back to the main queue to avoid crashes
+        // since these fetched result controller are created on viewContext
+        DispatchQueue.main.async { [unowned self] in
+            do {
+                try recentSavesController.performFetch()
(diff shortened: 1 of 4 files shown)


PocketKit/Sources/Sync/Collection/CollectionUrlFormatter.swift
```diff
@@ -5,25 +5,38 @@
 import Foundation
 
 /// A formatter for collection URLs
+/// valid formats are
```
bassrock, line 8: Do you need to do this to Syndicated Items too? And Read url items?

PocketKit/Sources/PocketKit/Home/HomeViewModel.swift
```diff
@@ -197,12 +197,17 @@ class HomeViewModel: NSObject {
 
     /// Fetch the latest data from core data and get the NSFetechedResults Controllers subscribing to updates
     func fetch() {
-        do {
-            try recentSavesController.performFetch()
-            try recomendationsController.performFetch()
-            try sharedWithYouController.performFetch()
-        } catch {
-            Log.capture(error: error)
+        // NOTE: despite HomeViewModel runs on MainActor, this call ends up on a different thread
```
bassrock, line 200: 🤣  MainActor != safe.

@nzeltzer you will enjoy this one...

bassrock (approved):

PocketKit/Sources/PocketKit/Home/HomeViewModel.swift
Gio2018: 🙃

PocketKit/Sources/Sync/Collection/CollectionUrlFormatter.swift
Gio2018: we only need this on collections because in certain situations we need to extract the slug from the path. The other two do not have a formatter like this one.

PR #1040, opened 2024-07-15
CI - Fix Danger

Gio2018:
## Goal
* Update Bitrise config to fix broken danger

### DIFF
--- a/UITests-1.xctestplan
+++ b/UITests-1.xctestplan
@@ -70,6 +70,7 @@
       "skippedTests" : [
         "AddSavedItemTests",
         "AddTagsItemTests",
+        "ArchiveAnItemTests",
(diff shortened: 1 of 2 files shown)


UITests-1.xctestplan
```diff
@@ -70,6 +70,7 @@
       "skippedTests" : [
         "AddSavedItemTests",
         "AddTagsItemTests",
+        "ArchiveAnItemTests",
```
bassrock, line 73: supposed to add here?

bassrock (approved):

UITests-1.xctestplan
Gio2018: Those tests were already executed in suite 2 so removed from here

PR #1046, opened 2024-07-18
Fix NSInternalInconsistency Exception errors

Gio2018:
## Goal
* This PR attempts to tackle a few `NSInternalInconsistencyException` errors seen in the app: these are all related to invalid section creation or indexing in a `UICollectionView` that uses compositional layout.

## Test Steps
* Build/run
* Smoke test the app
* Install it on a real device and make sure no background crashes occur

### DIFF
--- a/PocketKit/Sources/PocketKit/Home/HomeViewController.swift
+++ b/PocketKit/Sources/PocketKit/Home/HomeViewController.swift
@@ -48,15 +48,15 @@ class HomeViewController: UIViewController {
     private lazy var layout = UICollectionViewCompositionalLayout { [weak self] sectionIndex, env in
         guard let self else {
             Log.breadcrumb(category: "home", level: .debug, message: "Returning a nil section. Reason: HomeViewController is nil.")
-            return nil
+            return .empty()
         }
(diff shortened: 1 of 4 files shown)


PocketKit/Sources/PocketKit/Home/HomeViewController.swift
```diff
@@ -48,15 +48,15 @@ class HomeViewController: UIViewController {
...
         }
         guard let dataSource = self.dataSource else {
             Log.breadcrumb(category: "home", level: .debug, message: "Returning a nil section. Reason: datasource is nil.")
-            return nil
+            return .empty()
         }
         guard let section = dataSource.sectionIdentifier(for: sectionIndex) else {
             Log.breadcrumb(category: "home", level: .debug, message: "Returning a nil section. Reason: sectionIdentifier for \(sectionIndex) is nil.")
-            return nil
+            return .empty()
```
Gio2018, line 59: We can leave these as is for now in case we still see crashes, eventually we should remove those breadcrumbs and collapse the three `guard` statements into one

nzeltzer (approved):
lgtm

PR #1051, opened 2024-07-26
Fix listen not working from home

Gio2018:
## Goal
* Fix a bug where listen was not working when opening a recent saves item from home. This was because the listen presenting logic was baked into `SavesContainerViewController`. We could have added the same logic to `HomeViewController`, but a better solution was to move it to `ReadableHostViewController`, so that Listen would work regardless where we decide to present the saved article from.

## Test Steps
* Build/run
* Open an article from Home/Recent Saves
* Make sure listen works
* Also test listen from Saves and from Saves list, and make sure it still works as expected.


nzeltzer (approved):
Nice solution :)

PR #1053, opened 2024-07-27
Fix crash upon `SwiftUICollectionViewCell` deinit and add some more breadcrumbs and optimization in `HomeViewModel` snapshot calculation and rendering

Gio2018:
## Goal
- `SwiftUICollectionViewCell` cell retained unnecessary hosting controller to embed a SwiftUI view inside a UIKit UIView
- moreover, it attempted to manually destroy the hosting vc upon `denit` (also not necessary), causing crashes sometimes. Removing that reference and the code inside `deinit` entirely, should address the crash
- Add some more breadcrumbs to track more precisely the `NSInternalInconsistencyException` crash that occurs in Home
- Fix a broken logic, still in Home, that caused repeated snapshot updates when anything in `homeViewController.traitCollection` changed...


nzeltzer (approved):
lgtm

PR #1054, opened 2024-07-29
Improve tip presentation on saves

Gio2018:
## Goal
* On saves list, do not show the tip if the list is empty
* In the reader, only show the tip when reading a saved item

## Issues
POCKET-10399
POCKET-10400


nzeltzer (approved):
lgtm

PR #1067, opened 2024-08-28
Fix syndicated articles not opening in Reader from collections

Gio2018:
## Goal
* The navigation logic from native collections (which will need refactoring pretty soon) was wrong in case of syndicated articles

## Test Steps
* Open a collection with syndicated articles and make sure they open in the Reader


nzeltzer (approved):
lgtm

PR #1097, opened 2024-10-18
Listen: extend default language list to all supported languages

Gio2018:
## Goal
* Extend the default languages list in listen, since there are cases where it might not be fetched from the server

## Test Steps
* Listen to something that you couldn't before


nzeltzer (approved):
lgtm

PR #1123, opened 2024-11-26
Move code out of appdelegate

Gio2018:
## Goal
* Include short description of what this PR accomplishes

## Test Steps


Gio2018:
Closing this one as of now because we need app delegate to handle notifications, for the time being

PR #1126, opened 2024-12-04
Fix articles not opening when offline, and non parseable items sometimes being presented as articles

Gio2018:
## Goal
Fix the following issues on article:
- articles not opening when offline
- sometimes, in SwiftUI Home, the app attempted to display urls as articles when they were not parseable

## Test Steps
* Go off-line and make sure you can view articles
* Save a non-parseable item then access it from recent saves on SwftUI home and make sure it goes to a web view


nzeltzer (approved):
lgtm

rachtsingh:
Hey, I think this is still an issue. I just downloaded the update from the app store from 4 days ago (it says 8.23.1, which I think has the changes in this PR) and articles are still all grey bars when opened offline. This is after letting it sit there with wifi on for a little while (to allow it to download articles), then turning off wifi, then tapping on a saved article. 

I'm ok posting this somewhere else if it'll get the issue fixed, but as far as my intended purposes go (i.e. reading articles I saved earlier on the subway) the app is not usable in the current form.

PR #1127, opened 2024-12-04
Enable SwiftUI Home for all users

Gio2018:
## Goal
* This PR enables SwiftUI Home to all users
* This PR also removes the UIKit Home code. Only exception is the Collections code, that is currently still being used in Saves.
* This PR also moves the SwiftData `ModelContainer` initialization up to the app start, and out of `Services`
* Other changes:
    * move the remaining UIKit code under `Collections`
    * Update accessibility identifiers in SwiftUI Home to make ui tests work
    * Temporarily disable a few ui tests that will require further refactoring 
    * Add the `SlateInfo` type to send slate-related analytics
    * Remove `HomeViewModelTests` and `SlateDetailViewModelTests`, as those view models do not exist anymore

## Test Steps
* Even more now, make sure home feels like home


nzeltzer (approved):
lgtm

PocketApp.swift
```diff
@@ -4,24 +4,34 @@
...
+    private var scenePhase
 
-    @StateObject private var rootViewModel = RootViewModel()
+    @StateObject private var rootViewModel: RootViewModel
+
+    private let modelContainer: ModelContainer
+
+    init() {
+        _rootViewModel = StateObject(wrappedValue: RootViewModel())
+        modelContainer = DataController().makeModelContainer()
```
Gio2018, line 22: this was moved here from `Services` so that the container is now available to all the app.

PocketKit/Sources/PocketKit/Article/ReadableViewModel/RecommendableItemViewModel.swift
```diff
@@ -406,7 +406,7 @@ extension RecommendableItemViewModel {
 
 // MARK: instantiation from url
 extension RecommendableItemViewModel {
-    static func fromURL(_ urlString: String) -> RecommendableItemViewModel? {
+    static func fromURL(_ urlString: String, readableSource: ReadableSource) -> RecommendableItemViewModel? {
```
Gio2018, line 409: injecting `readableSource` here as this method is now used by callers with different `ReadableSource` values

PocketKit/Sources/PocketKit/Collections/CollectionViewModel.swift
```diff
@@ -11,6 +11,31 @@ import Localization
 import Analytics
 import Network
 
+enum ReadableType {
```
Gio2018, line 14: This code used to reside in `HomeViewModel` and moved here since it's now the only place where it's used.

PocketKit/Sources/PocketKit/DataController.swift
```diff
@@ -6,16 +6,17 @@
 import Foundation
 @preconcurrency import SwiftData
 import SharedPocketKit
+import Sync
 
 public struct DataController: Sendable {
-    public func makeModelContainer(groupID: String) -> ModelContainer {
+    public func makeModelContainer() -> ModelContainer {
```
Gio2018, line 12: the construction of `ModelContainer` was moved to `PocketKit` so there's no need to inject a `groupID` externally

PocketKit/Sources/PocketKit/Home/Models/ItemInfo.swift
```diff
@@ -18,3 +18,13 @@ struct AnalyticsInfo {
         self.externalDestination = externalDestination
     }
 }
+
+struct SlateInfo: Codable, Equatable, Hashable {
```
Gio2018, line 22: added for slate detail tracking purposes. Also renamed `AnalyticsInfo` to `ItemInfo` for consistency

PocketKit/Sources/PocketKit/Home/Models/ReadableSource.swift
```diff
@@ -0,0 +1,12 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import Foundation
+
+enum ReadableSource: String, Codable {
```
Gio2018, line 7: this was extracted from the deleted `HomeViewModel`

PocketKit/Sources/PocketKit/Home/Views/Cards/CardView.swift
```diff
@@ -149,6 +151,20 @@ private extension CardView {
         }
     }
 
+    /// We are legally allowed to open the item in reader view except one of the following conditions is met:
+    /// a) the item is not an article (i.e. it was not parseable)
+    /// b) the item is an image
+    /// c) the item is a video
+    /// - Returns: true if we can open this item in the reader
+    func isReaderSupported() -> Bool {
```
Gio2018, line 159: added this method that checks if the underlying `Item` supports reader. This is done because while we observe `SavedItem` via `@Query` (which allows to dynamically update the view when an item is saved, archived or favorited), swift observation does not necessarily guarantees that changes to related items will be immediately reflected (thus for example the `isArticle` property might not be immediately be updated if a user saves a recommendation and open it right after: this could cause the item to be still opened in a web view instead of the reader).
We could also do a `@Query` to observe both the `SavedItem` and the underlying `Item` but that would create unnecessary overhead, since we only need this when a user navigates to a saved item from the related recommendation.

PocketKit/Sources/PocketKit/Main/MainView.swift
```diff
@@ -92,29 +85,16 @@ public struct MainView: View {
         .sheet(isPresented: $model.isPresentingHooray) {
             PremiumUpgradeSuccessView()
         }
-    }
-
-    func makeUIKitHome() -> some View {
```
Gio2018, line 97: all UIKit home code was removed, so...

PocketKit/Sources/PocketKit/Main/MainView.swift
```diff
@@ -92,29 +85,16 @@ public struct MainView: View {
...
-                } else {
-                    Image(asset: .tabHomeDeselected)
-                }
-                Text(Localization.home)
-            }
-            .accessibilityIdentifier("home-tab-bar-button")
-            .tag(MainViewModel.AppSection.home)
-            .task {
-                Log.breadcrumb(category: "SWiftUIHome", level: .debug, message: "UIKit Home initialized.")
+        .fullScreenCover(isPresented: $model.isPresentingWebView) {
```
Gio2018, line 88: by removing UIKit home, some things needed to he handled differently, and in particular the presentation of an item from a deep link that needs a web view goes here.

PocketKit/Sources/PocketKit/Main/MainViewModel.swift
```diff
@@ -270,27 +232,42 @@ extension MainViewModel {
         linkRouter.matchRoute(from: components.url!)
     }
 
+    private func handleHomeNavigation(givenURL: String, collectionSlug: String?, isReadableSavedItem: Bool, isSyndicated: Bool, source: ReadableSource) {
```
Gio2018, line 235: deep links are now handled via the SwiftUI navigation system.

PR #1139, opened 2025-01-16
Sync logic for Notes

Gio2018:
## Goal
* Add sync logic for notes
> [!NOTE]
> the sync logic will be updated when we add notes to SavedItems etc

## Test Steps
* Nothing testable with the UI yet, but if you have tools like Core Data Lab and access to Apollo studio, you can:
    - create notes from Apollo studio to your account (ideally more than 30 so you have more than one page, notes can also have the same title and content)
    - Launch the app, login with that same account, and observe the notes being populated in Core Data Lab (or any other SQLIte viewer of choice)
    - Make sure the fields you receive are consistent with what you see in Apollo studio


nzeltzer (approved):
lgtm

PR #1141, opened 2025-01-29
Fix some crashes

Gio2018:
## Goal
* This PR improves accessing some data from SwiftData/Core Data to prevent potential crashes

## Test Steps
* Smoke test home and make sure it does not crash


nzeltzer (approved):
lgtm

